PR #3, opened 2020-05-06
Add examples of industry choices


evrardjp:
We should add the "state of the art" in our documentation.

This patch should tackle it by initiating a new list
of examples on how companies might have solved the same problem.


evrardjp:
This is just a list to get started, and we should probably expand with our research.

pjbgf:
@evrardjp yet another example: https://github.com/UKHomeOffice/seccomp-config

Do you mind adding it in? Otherwise I can create a PR once this one is merged.

evrardjp:
Adding it!

PR #4, opened 2020-05-25
Add initial scaffolding for seccomp profile controller

hasheddan:
This sets up some initial scaffolding for reconciling config maps that specify themselves as seccomp profiles. The implementation simply filters config maps and logs the resource version of config maps with the seccomp profile annotation.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>


controllers/profile/profile.go
```diff
@@ -0,0 +1,100 @@
...
+	reconcileTimeout = 1 * time.Minute
+
+	longWait = 1 * time.Minute
+
+	errGetProfile = "cannot get profile"
+)
+
+// SeccompProfileAnnotation is the annotation on a ConfigMap that specifies its
+// intention to be treated as a seccomp profile.
+const SeccompProfileAnnotation = "seccomp.security.alpha.kubernetes.io/profile"
```
saschagrunert, line 46: Do we want to remove the alpha because it will be GA in 1.19?
```suggestion
const SeccompProfileAnnotation = "seccomp.security.kubernetes.io/profile"
```

controllers/profile/profile_test.go
```diff
@@ -0,0 +1,157 @@
...
+			want: true,
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			got := isProfile(tc.obj)
+			if got != tc.want {
+				t.Errorf("isProfile(...): got %v, want %v", got, tc.want)
+			}
```
saschagrunert, line 61: Can we use `testify/require` for those? I know there was no decision around testing but this would reduce the noise:
```suggestion
			require.Equal(t, got, want)
```

main.go
```diff
@@ -0,0 +1,63 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/klog/klogr"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/saschagrunert/seccomp-operator/controllers/profile"
+)
+
+var (
+	sync     = time.Second * 30
+	scheme   = runtime.NewScheme()
```
saschagrunert, line 32: This var seems unused to me.

main.go
```diff
@@ -0,0 +1,63 @@
...
+	if err := profile.Setup(mgr, ctrl.Log.WithName("profile")); err != nil {
+		setupLog.Error(err, "cannot setup profile controller")
+		os.Exit(1)
+	}
+
+	setupLog.Info("starting manager")
+	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+		setupLog.Error(err, "controller manager error")
+		os.Exit(1)
+	}
```
saschagrunert, line 62: I would refactor that on top of a new function which returns an `error`, but we can do that as a follow up. :+1:

controllers/profile/profile.go
pjbgf: I think that's a fair point. 

In terms of feature gates and feature life cycle, we can hold off dealing with them until we get a MVP out.

controllers/profile/profile.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Reconcile reconciles a ConfigMap representing a seccomp profile.
+func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+	logger := r.log.WithValues("profile", req.Name, "namespace", req.Namespace)
+
+	ctx, cancel := context.WithTimeout(context.Background(), reconcileTimeout)
+	defer cancel()
+
+	profile := &corev1.ConfigMap{}
+	if err := r.client.Get(ctx, types.NamespacedName{Name: req.Name, Namespace: req.Namespace}, profile); err != nil {
```
pjbgf, line 86: Would this accommodate all namespaces?

On a side note, when saving on disk potentially we could use:
```seccomp-root/<configmap-namespace>/<configmap-name>/<configmap-filename>```

main.go
hasheddan: Good catch! We are just using the controller-runtime default scheme for now :+1:

main.go
hasheddan: Agreed :+1: We will also probably want to pull out setting up controllers, but since there was only one here I just kept it all together

controllers/profile/profile.go
hasheddan: Yep :+1: All `ConfigMap` in any namespace will be watched here :+1:

controllers/profile/profile_test.go
hasheddan: Fine by me :)

hasheddan:
@saschagrunert @pjbgf I think all comments have been addressed :) Let me know if we need any further changes here :+1:

pjbgf:
LGTM

PR #6, opened 2020-06-16
Save config maps into disk

pjbgf:
Save config maps into disk whilst blocking file traversal at different components.
This may require some improvements around testing, specially around E2E.

### DIFF
--- a/controllers/profile/profile.go
+++ b/controllers/profile/profile.go
@@ -18,6 +18,10 @@ package profile
 
 import (
 	"context"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
 	"time"
 
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
@@ -38,7 +42,23 @@ const (
 
 	longWait = 1 * time.Minute
 
-	errGetProfile = "cannot get profile"
+	errGetProfile                = "cannot get profile"
+	errConfigMapNil              = "config map cannot be nil"
+	errConfigMapWithoutNamespace = "config map must have a namespace"
+	errConfigMapWithoutName      = "config map must have a name"
+	errSavingProfile             = "cannot save profile"
+	errCreatingOperatorDir       = "cannot create operator directory"
+
+	seccompOperatorSuffix string      = "operator"
+	filePermissionMode    os.FileMode = 0o600
+
+	// MkdirAll won't create a directory if it does not have the execute bit.
+	// https://github.com/golang/go/issues/22323#issuecomment-340568811
+	dirPermissionMode os.FileMode = 0o700
+)
+
+var (
+	kubeletSeccompRootPath string = "/var/lib/kubelet/seccomp"
 )
 
 // SeccompProfileAnnotation is the annotation on a ConfigMap that specifies its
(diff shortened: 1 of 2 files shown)


controllers/profile/profile.go
```diff
@@ -38,7 +42,23 @@ const (
...
-	errGetProfile = "cannot get profile"
+	errGetProfile                = "cannot get profile"
+	errConfigMapNil              = "config map cannot be nil"
+	errConfigMapWithoutNamespace = "config map must have a namespace"
+	errConfigMapWithoutName      = "config map must have a name"
+	errSavingProfile             = "cannot save profile"
+	errCreatingOperatorDir       = "cannot create operator directory"
+
+	seccompOperatorSuffix string      = "operator"
+	filePermissionMode    os.FileMode = 0600
```
saschagrunert, line 53:
```suggestion
	filePermissionMode    os.FileMode = 0o600
```

controllers/profile/profile.go
```diff
@@ -38,7 +42,23 @@ const (
...
+
+	// MkdirAll won't create a directory if it does not have the execute bit.
+	// https://github.com/golang/go/issues/22323#issuecomment-340568811
+	dirPermissionMode os.FileMode = 0700
```
saschagrunert, line 57:
```suggestion
	dirPermissionMode os.FileMode = 0o700
```

controllers/profile/profile.go
```diff
@@ -38,7 +42,23 @@ const (
...
+)
+
+var (
+	kubeletSeccompRootPath string = "/var/lib/kubelet/seccomp"
```
saschagrunert, line 61: This can be `const` I guess.

controllers/profile/profile.go
```diff
@@ -88,10 +108,60 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		return reconcile.Result{}, errors.Wrap(ignoreNotFound(err), errGetProfile)
 	}
 
+	for name, contents := range profile.Data {
+		profilePath, err := getProfilePath(name, profile)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+
+		saveProfileOnDisk(profilePath, contents)
```
saschagrunert, line 117: I think we should handle the returned error.

controllers/profile/profile.go
pjbgf: Initially I created it as a `const`, but then changed it to make it testable - allowing tests to fiddle file/folder permissions without needing to run the tests as root. 

Assuming we will initially only support the operator to run as a `DaemonSet` inside the cluster (as opposed to running directly into the host nodes), this becomes less of a link to the actual seccomp root path on the host, but rather what is the mapped path inside the container. So I was wondering whether is there a better way to tackle this - any suggestions?

controllers/profile/profile.go
pjbgf: good spot, I completely overlooked this before - it should now be handled now.

controllers/profile/profile.go
saschagrunert: Okay let's leave it like it is for now. Later on we can put this value into a configuration struct to ensure easier testing.

controllers/profile/profile.go
```diff
@@ -88,10 +108,64 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+			return errors.Wrap(err, errCreatingOperatorDir)
+		}
+		if os.IsNotExist(err) {
+			if err := os.MkdirAll(dirPath, dirPermissionMode); err != nil {
+				return errors.Wrap(err, errSavingProfile)
+			}
+		}
+	}
+
+	if err := ioutil.WriteFile(fileName, []byte(contents), filePermissionMode); err != nil {
```
hasheddan, line 141: Do we want to validate that the contents are actually a valid profile configuration?

pjbgf: That is a very good point. I was already planning a follow-up PR with that.

saschagrunert:
Cool, merged that one for now. :)

PR #7, opened 2020-06-17
Create MVP deployment files

pjbgf:
- Add the yaml files required to deploy the operator.
- Add Dockerfile.
- Create Usage.md to explain current features and how to install the operator.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,20 @@
+FROM golang:1.14-alpine AS build
+WORKDIR /go/src/saschagrunert/seccomp-operator
+RUN apk --no-cache add build-base git gcc
+
+# Speed up build by leveraging docker layer caching
+COPY go.mod go.sum ./
+RUN go mod download
+
+ADD . /go/src/saschagrunert/seccomp-operator
+RUN go build -ldflags '-s -w' -o /seccomp-operator
+
+
+
+FROM alpine
+LABEL name="Seccomp Operator" \
+      version="v0.0.1" \
+      description="The Seccomp Operator makes it easier for cluster admins to manage their seccomp profiles and apply them to Kubernetes' workloads."
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.14-alpine AS build
+WORKDIR /go/src/saschagrunert/seccomp-operator
+RUN apk --no-cache add build-base git bzr mercurial gcc
+
+# Speed up build by leveraging docker layer caching
+COPY go.mod go.sum ./
+RUN go mod download
+
+ADD . /go/src/saschagrunert/seccomp-operator
+RUN go build -o /seccomp-operator
```
saschagrunert, line 10:
```suggestion
RUN go build -ldflags '-s -w' -o /seccomp-operator
```

deploy/operator.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: apps/v1
+kind: Deployment
```
saschagrunert, line 2: The operator should be a DaemonSet, right?

Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+
+
+
+FROM alpine
```
saschagrunert, line 14: Do we want to use `scratch` and link statically?

deploy/operator.yaml
pjbgf: Very good point, just changed it to DaemonSet.

Dockerfile
pjbgf: I had some issues running from scratch, so was thinking on optimising (and also locking it down) once we have some the MVP functionality done and some automated E2E. What do you think?

saschagrunert: Sounds good, let's leave it like it is for now.

PR #8, opened 2020-06-18
Restrict ConfigMap to single namespace plus security improvements

pjbgf:
- Add RBAC for service principal
- Add usage and fix example
- Drop caps and no_new_privs
- Restrict configmaps to a single namespace

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -9,6 +9,30 @@ metadata:
   name: seccomp-operator
   namespace: security-operators
 ---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: config-map-reader
+  namespace: security-operators
+rules:
+- apiGroups: [""]
+  resources: ["configmaps"]
+  verbs: ["get", "watch", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: config-map-reader-binding
+  namespace: security-operators
+subjects:
+- kind: ServiceAccount
+  name: seccomp-operator
+  namespace: security-operators
+roleRef:
+  kind: Role
+  name: config-map-reader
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -27,13 +27,15 @@ import (
 )
 
 var (
-	sync     = time.Second * 30
-	setupLog = ctrl.Log.WithName("setup")
+	sync                    = time.Second * 30
+	setupLog                = ctrl.Log.WithName("setup")
+	namespaceToWatch string = "security-operators"
```
hasheddan, line 32: Do we want this to be configurable?

deploy/operator.yaml
```diff
@@ -29,10 +53,14 @@ spec:
           image: "securityoperators/seccomp-operator:latest"
           imagePullPolicy: Always
           volumeMounts:
-          - mountPath: /var/lib/kubelet/seccomp
-            name: seccomp-root-path
+          - mountPath: /var/lib/kubelet
+            name: kubelet-root-path
+          securityContext:
+            allowPrivilegeEscalation: false
```
saschagrunert, line 59: That’s the default right? If so I guess we can remove it.

pjbgf: It actually defaults to true: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privilege-escalation

main.go
pjbgf: Is it worth also making reconciliation intervals configurable?

deploy/operator.yaml
saschagrunert: Whoot.

main.go
pjbgf: Fixed it - although I am not sure about the ENV name I used: `NAMESPACE_TO_WATCH`.

saschagrunert:
Thank you!

PR #9, opened 2020-06-22
Create yaml for non-root user and apply runtime/default seccomp.

pjbgf:
- Adds an extra `.yaml` to deploy the seccomp-operator running the main container as non `root` user.
- Makes all containers to be bound by the default seccomp.

In the future, once we have profiled the operator, we may create its own tailor made seccomp which we would be able to apply to anything but the init-container.

### DIFF
--- a/deploy/operator-non-root.yaml
+++ b/deploy/operator-non-root.yaml
@@ -50,6 +50,28 @@ spec:
         name: seccomp-operator
     spec:
       serviceAccountName: seccomp-operator
+      initContainers:
+        - name: non-root-enabler
+          image: busybox 
+          # Creates folder /var/lib/kubelet/seccomp/operator and sets 2000:2000 as its owner.
+          # This is required to allow the main container to run as non-root.
+          command: ["/bin/sh", "-c"]
+          args:
+            - >
+              if [ ! -d "/var/lib/kubelet/seccomp/operator" ]; then 
+                /bin/mkdir -p /var/lib/kubelet/seccomp/operator && 
+                chmod 0744 /var/lib/kubelet/seccomp/operator && 
+                /bin/chown -R 2000:2000 /var/lib/kubelet/seccomp; 
+              fi
+          volumeMounts:
+          - mountPath: /var/lib/kubelet
+            name: kubelet-path
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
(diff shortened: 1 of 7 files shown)


deploy/operator-non-root.yaml
```diff
@@ -0,0 +1,89 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: security-operators
```
saschagrunert, line 4: We probably should stick to `seccomp-operator` here…

deploy/operator-non-root.yaml
```diff
@@ -0,0 +1,89 @@
...
+      labels:
+        name: seccomp-operator
+    spec:
+      serviceAccountName: seccomp-operator
+      initContainers:
+        - name: non-root-enabler
+          image: busybox 
+          # Creates folder /var/lib/kubelet/seccomp/operator and sets 2000:2000 as its owner.
+          # This is required to allow the main container to run as non-root.
+          command: ['sh', '-c', 'if [ ! -d "/var/lib/kubelet/seccomp/operator" ]; then /bin/mkdir -p /var/lib/kubelet/seccomp/operator && chmod 0744 /var/lib/kubelet/seccomp/operator && /bin/chown -R 2000:2000 /var/lib/kubelet/seccomp; fi']
```
saschagrunert, line 58: I suggest using somthing like:
```suggestion
          command:
            - sh -c
            - |
              if [ …]; then …
```

deploy/operator-non-root.yaml
```diff
@@ -0,0 +1,89 @@
...
+            name: kubelet-path
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
+            capabilities:
+              drop: ["ALL"]
+              add: ["CHOWN"]
+      containers:
+        - name: seccomp-operator
+          image: "securityoperators/seccomp-operator:latest"
```
saschagrunert, line 70:
```suggestion
          image: securityoperators/seccomp-operator:latest
```

deploy/operator-non-root.yaml
pjbgf: Sure thing - just reverted that across the board.

pjbgf:
> Yeah I think that's a good path to move forward, just found some nits.

Thank you for reviewing it. Made the changes now. :+1:

PR #10, opened 2020-06-24
Add Makefile, go build and mod verify target

saschagrunert:
This introduces a new Makefile to collect common build and test targets.
The first one is `make [all,build/seccomp-operator]` which builds the
operator in build/seccomp-operator. This allows us to collect further
artifacts in a global `build/` dir in the future.

Beside that, the `go-mod` target verifies the integrity of the go
modules and can be used for later CI verification.


pjbgf:
Good stuff!

/lgtm

PR #14, opened 2020-06-24
Add verify-go-mod Makefile target

saschagrunert:
#### What type of PR is this?

/kind cleanup

The verification target can be used of the go modules are in sync with
the repository.

None
None

```release-note
None
```


pjbgf:
The idea here is to link this to some CI process, right? It may be worth adding it as a single piece.

Makefile
```diff
@@ -23,3 +23,7 @@ go-mod:
 	export GO111MODULE=on \
 		$(GO) mod tidy && \
 		$(GO) mod verify
+
+.PHONY: verify-go-mod
+verify-go-mod: go-mod
+	hack/tree-status
```
hasheddan, line 29: This looks good for now. In the future we likely want a separate target that will do all steps that produce any generated code, then run `hack/tree-status`, but I believe `verify-go-mod` is the only place we are doing this for now.

saschagrunert:
> The idea here is to link this to some CI process, right?

Yep, those `verify-*` Makefile targets should be executed during CI or locally.

> It may be worth adding it as a single piece.
What do you mean when referring to "single piece"? As a dedicated script?

Makefile
saschagrunert: Yeah I thought that we may have multiple verification scripts that rely on `hack/tree-status`. For example if we shell format scripts via shfmt.

pjbgf:
> > It may be worth adding it as a single piece.
> 
> What do you mean when referring to "single piece"? As a dedicated script?

I was referring more into adding the E2E CI processes, but I suppose we probably need to wait until we get the new sig repo. 
Let's go ahead with this for now.

PR #17, opened 2020-06-24
Move controller to internal/pkg and main.go to cmd

saschagrunert:
#### What type of PR is this?

/kind cleanup

This follows a more default golang project layout by moving the binary
into `cmd/seccomp-operator` and the library into `internal/pkg`.
None

None

```release-note
None
```


cmd/seccomp-operator/main.go
```diff
@@ -24,7 +24,7 @@ import (
 	"k8s.io/klog/klogr"
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	"github.com/saschagrunert/seccomp-operator/controllers/profile"
+	"github.com/saschagrunert/seccomp-operator/internal/pkg/controllers/profile"
```
hasheddan, line 27: Any particular reason that we need to be `internal/pkg` rather than just `pkg`?

saschagrunert: I don’t know, I just thought that we do not want to provide a library for others in the first place. So if move it to `internal` to be in-importable. Do you prefer `pkg`?

hasheddan: Yeah, that sounds fine to me. I usually opt to expose more with the disclaimer of "use this at your own risk", but I know that many folks disagree, including Dave Cheney: https://dave.cheney.net/2019/10/06/use-internal-packages-to-reduce-your-public-api-surface

PR #20, opened 2020-06-25
Add version subcommand and CLI parser

saschagrunert:
#### What type of PR is this?

/kind feature

To be able to identify the seccomp operator build appropriately we now
provide a `version,v` command.

None

None

```release-note
- Added `version,v` subcommand and CLI parser (`--version` works too now)
```

### DIFF
--- a/internal/pkg/version/version_test.go
+++ b/internal/pkg/version/version_test.go
@@ -1,0 +1,41 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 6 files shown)


saschagrunert:
@hasheddan @pjbgf please take a look :)

internal/pkg/version/version_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestVersionText(t *testing.T) {
+	sut := Get()
+	require.NotEmpty(t, sut.String())
+}
+
+func TestVersionJSON(t *testing.T) {
```
pjbgf, line 30: nit: would be nice to have tests for unknown versions and also when the version is known.

saschagrunert: Hm this is done via the `$LDVARS` so I'm not sure how exactly test both cases. In any case I added the vars to the `test-unit` target so we now can check that they're not empty per default.

PR #23, opened 2020-06-25
Add verify-go-lint Makefile target

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now use golangci-lint to lint the code based on the .golangci.yml
file.

None
None

```release-note
None
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,131 @@
+---
+run:
+  build-tags:
+    - netgo
+  concurrency: 6
+  deadline: 5m
+linters:
+  disable-all: true
+  enable:
+    - asciicheck
+    - bodyclose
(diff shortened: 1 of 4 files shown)


.golangci.yml
```diff
@@ -0,0 +1,131 @@
...
+    - varcheck
+    - whitespace
+    - gocognit
+    - godox
+    - gomnd
+    - gomodguard
+    - gosec
+    - lll
+    - nestif
+    # - funlen
```
hasheddan, line 51: Do we think we will want to enable these commented out rules in the future?

saschagrunert: Yes, I wanted to follow up to fix some of those as well. :)

PR #26, opened 2020-06-27
Make kubelet seccomp root path const

saschagrunert:
#### What type of PR is this?

/kind cleanup

We had it as a `var` because of the tests but we can use a const because
the tests do not rely on changing that global variable.
None

cc @hasheddan @pjbgf 

```release-note
None
```

### DIFF
--- a/internal/pkg/controllers/profile/profile.go
+++ b/internal/pkg/controllers/profile/profile.go
@@ -54,10 +54,8 @@ const (
 	// MkdirAll won't create a directory if it does not have the execute bit.
 	// https://github.com/golang/go/issues/22323#issuecomment-340568811
 	dirPermissionMode os.FileMode = 0o744
-)
 
-var (
-	kubeletSeccompRootPath string = "/var/lib/kubelet"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -55,7 +55,7 @@ verify-go-mod: go-mod
 
 .PHONY: test-unit
 test-unit: $(BUILD_PATH)
-	$(GO) test -v -test.coverprofile=$(BUILD_PATH)/coverage.out ./...
+	$(GO) test -v -test.coverprofile=$(BUILD_PATH)/coverage.out ./internal/...
```
hasheddan, line 58: Do we need to restrict test coverage to “internal”?

saschagrunert: Yeah otherwise we would also execute the e2e tests in `./test` 🤭

hasheddan: Ahh I see! I have solved this before with build tags on integration test files (https://mickey.dev/posts/go-build-tags-testing/), but this seems like an equally valid approach :)

saschagrunert: Hm, let me revert that change, I like the build tags approach more and will follow-up in another PR. :+1:

PR #28, opened 2020-06-28
Add operator deploy e2e test

saschagrunert:
#### What type of PR is this?

/kind design

The e2e test will be the base for further enhance tests, right now we
just build the container image and deploy the operator into the kind
cluster.

None

None

```release-note
None
```


Dockerfile
```diff
@@ -28,13 +27,6 @@ RUN adduser \
     --uid "${UID}" \
     "${USER}"
 
-# Speed up build by leveraging docker layer caching
-COPY go.mod go.sum ./
-RUN go mod download
-
-ADD . /work
-RUN make
```
saschagrunert, line 36: We now build locally, which should be fine since we're building a static binary.

saschagrunert:
Ready to be reviewed, PTAL @hasheddan @pjbgf

Dockerfile
```diff
@@ -28,21 +27,14 @@ RUN adduser \
...
-
-FROM alpine
+FROM scratch
 LABEL name="Seccomp Operator" \
       version="v0.0.1" \
       description="The Seccomp Operator makes it easier for cluster admins to manage their seccomp profiles and apply them to Kubernetes' workloads."
 
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /etc/passwd /etc/passwd
```
pjbgf, line 36: We could follow the same logic and make this simpler by amending the files directly:
```suggestion
RUN echo "2000:x:2000:2000::/nonexistent:/sbin/nologin" >> /etc/passwd &&
    echo "2000:x:2000:" >> /etc/group
```

Dockerfile
```diff
@@ -28,21 +27,14 @@ RUN adduser \
...
 COPY --from=build /etc/group /etc/group
```
pjbgf, line 37:
```suggestion
```

Dockerfile
```diff
@@ -14,7 +14,6 @@
 
 FROM golang:1.14-alpine AS build
 WORKDIR /work
-RUN apk --no-cache add build-base git gcc
 
 ENV USER=secuser
```
pjbgf, line 18: Conversely, we could remove the entire first stage image.

saschagrunert:
> If we are to stop using multi-staged images to build the project, we probably should drop the concept altogether and amend `/etc/passwd/` and `/etc/group` directly on the last image.
> 
> I personally prefer the multi-staged approach to build as it decreases the entry effort to contribute to the project. New contributors only need docker installed in their machines. It can also be handy to have the same errors between machines and CI.

Hm okay I revert my changes back and stick to the multi-layer build approach. I think we can leave those parts as they currently are. :+1:

Dockerfile
saschagrunert: See https://github.com/saschagrunert/seccomp-operator/pull/28#issuecomment-651650217

Dockerfile
saschagrunert: See https://github.com/saschagrunert/seccomp-operator/pull/28#issuecomment-651650217

Dockerfile
saschagrunert: See https://github.com/saschagrunert/seccomp-operator/pull/28#issuecomment-651650217

saschagrunert:
PTAL @hasheddan @pjbgf

test/suite_test.go
```diff
@@ -35,50 +52,83 @@ func (e *e2e) SetupSuite() {
...
 				kindVersion+"/kind-linux-amd64",
-		).RunSuccess())
+		)
 		e.Nil(os.Chmod(e.kindPath, 0o755))
 	}
 }
 
 // SetupTest starts a fresh kind cluster for each test
 func (e *e2e) SetupTest() {
+	// Deploy the registry
```
hasheddan, line 73: Could we skip doing this and just load the image into the kind cluster? https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster

saschagrunert: Hah, yeah I change the implementation as suggested and will choose a pull policy of `Never`.

hasheddan: This is one of my favorite features of `kind` :)

saschagrunert:
Done changing to pre-load the images instead of deploying a registry. PTAL @hasheddan :)

saschagrunert:
Thank you for the review!

PR #35, opened 2020-07-14
Add scripts for image build and e2e test jobs

saschagrunert:
#### What type of PR is this?

/kind cleanup

We have to ensure kubectl to not have a need to include it into the test
image. This will now be done by the test suite itself to make them more
independent from the environment.
Ref https://github.com/kubernetes/test-infra/pull/18301

None

```release-note
None
```

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -33,13 +33,16 @@ import (
 
 const (
 	kindVersion = "v0.8.1"
+	kindSHA512  = "ff71adddbe043df84c4dee82f034c6856bfc51c931bb7839d9c09d02fca93bfe961a9c05d7c657371963cc81febee175133598bba50fb1481a9faa06b42abdc3"
 	kindImage   = "kindest/node:v1.18.2"
-	testImage   = "securityoperators/seccomp-operator:latest"
+
+	testImage = "securityoperators/seccomp-operator:latest"
 )
 
 type e2e struct {
 	suite.Suite
 	kindPath    string
+	kubectlPath string
 	clusterName string
 }
 
@@ -47,7 +50,7 @@ func TestSuite(t *testing.T) {
 	suite.Run(t, &e2e{})
 }
 
-// SetupSuite downloads kind
+// SetupSuite downloads kind and searches for kubectl in $PATH
 func (e *e2e) SetupSuite() {
 	cwd, err := os.Getwd()
 	e.Nil(err)
@@ -64,8 +67,12 @@ func (e *e2e) SetupSuite() {
 			"https://github.com/kubernetes-sigs/kind/releases/download/"+
 				kindVersion+"/kind-linux-amd64",
 		)
-		e.Nil(os.Chmod(e.kindPath, 0o755))
+		e.Nil(os.Chmod(e.kindPath, 0o700))
+		e.verifySHA256(e.kindPath, kindSHA512)
 	}
+
+	e.kubectlPath, err = exec.LookPath("kubectl")
+	e.Nil(err)
 }
 
 // SetupTest starts a fresh kind cluster for each test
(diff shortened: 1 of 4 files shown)


saschagrunert:
/hold 
for some fix

saschagrunert:
/hold cancel

Works, we just have to run as privileged and on host network to be able to access the kind cluster.

test/suite_test.go
```diff
@@ -32,22 +32,24 @@ import (
...
+	kubectlPath string
 	clusterName string
 }
 
 func TestSuite(t *testing.T) {
 	suite.Run(t, &e2e{})
 }
 
-// SetupSuite downloads kind
+// SetupSuite downloads kind and kubectl
```
hasheddan, line 52: In practice won't these be present in the kubekins image?

saschagrunert: Hm yeah, just saw that they're part of the image. Should I drop the download? I think this way we could decide on our own which version of kubectl we wanna use (probably always the latest)…

test/suite_test.go
```diff
@@ -66,6 +68,16 @@ func (e *e2e) SetupSuite() {
 		)
 		e.Nil(os.Chmod(e.kindPath, 0o755))
 	}
+
+	e.kubectlPath = filepath.Join(buildDir, "kubectl")
+	if !util.Exists(e.kubectlPath) {
+		e.run(
+			"curl", "-o", e.kubectlPath,
```
pjbgf, line 75: When downloading and executing binaries, it is good practice to validate their checksum first. Here's an example:

`echo 5a99ce73a14da6ab561016b83906c398f082dd7571771b6b16430445f6668587a2ad2f5ca9e5a330bcc3fad5f4a7f21109042defcd674acae8cd7170b49ed947 ~/Downloads/kubectl | sha512sum -c -`

If the checkshum does not match it will return a non-zero exit code - which you can use to fail the setup. 

I would recommend that the checksum value is kept on a variable to be updated every time the kubectl version is changed. To find the correct checksum for a given version:

https://storage.googleapis.com/kubernetes-release/release/v1.18.5/bin/linux/amd64/kubectl.sha512

test/suite_test.go
```diff
@@ -66,6 +68,16 @@ func (e *e2e) SetupSuite() {
...
+			"https://storage.googleapis.com/kubernetes-release/release/"+
+				kubectlVersion+"/bin/linux/amd64/kubectl",
+		)
+		e.Nil(os.Chmod(e.kubectlPath, 0o755))
```
pjbgf, line 79: Given the nature of this setup I would probably keep the ACL focused on the current user.
```suggestion
		e.Nil(os.Chmod(e.kubectlPath, 0o700))
```

test/suite_test.go
```diff
@@ -32,22 +32,24 @@ import (
 )
 
 const (
-	kindVersion = "v0.8.1"
-	kindImage   = "kindest/node:v1.18.2"
-	testImage   = "securityoperators/seccomp-operator:latest"
+	kindVersion    = "v0.8.1"
+	kubectlVersion = "v1.18.5"
+	kindImage      = "kindest/node:v1.18.2"
+	testImage      = "securityoperators/seccomp-operator:latest"
```
pjbgf, line 38: more of a question: would it be worth having the current branch's image being used?

test/suite_test.go
saschagrunert: Sure, I added a new SHA verification helper function and added it to the download.

test/suite_test.go
saschagrunert: Fixed :+1:

test/suite_test.go
saschagrunert: It will be used, we build it locally and kind loads the image into the node :) This means every code change should be part of the e2e test.

saschagrunert:
Please take a look again @hasheddan @pjbgf

test/suite_test.go
hasheddan: It feels like it would be cleaner, do you know how any other projects handle this?

saschagrunert: Alright, I just changed to to lookup `kubectl` in `$PATH`.

PR #36, opened 2020-07-14
Add cloudbuild.yaml for images in gcr.io

saschagrunert:
#### What type of PR is this?
/kind feature

This allows us to utilize Google Cloud Build to build and publish the seccomp-operator images.
None

A periodic job in test-infra should later take care in building and publishing the image.

```release-note
None
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,19 @@
+---
+# See https://cloud.google.com/cloud-build/docs/build-config
+timeout: 1200s
+options:
+  substitution_option: ALLOW_LOOSE
+steps:
+  - name: gcr.io/cloud-builders/docker
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/seccomp-operator:$_GIT_TAG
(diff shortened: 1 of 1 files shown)


cloudbuild.yaml
```diff
@@ -0,0 +1,20 @@
...
+    args:
+      - build
+      - --tag=gcr.io/$PROJECT_ID/seccomp-operator:$_GIT_TAG
+      - --tag=gcr.io/$PROJECT_ID/seccomp-operator:latest
+      - .
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form
+  # vYYYYMMDD-hash, and can be used as a substitution
+  _GIT_TAG: '12345'
+  _PULL_BASE_REF: 'dev'
```
hasheddan, line 17: What does setting this value do?

saschagrunert: It's the ref name of the base branch when using in prow. I think we do not need to set this substitution at all.

https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#job-environment-variables

PR #49, opened 2020-07-31
Watch all namespaces by default

### LINKED ISSUE #40: Make seccomp profile applicable per namespace
We decided that the operator should watch every namespace, therefore we can deploy profiles from every namespace.

rhafer:
#### What type of PR is this?

> /kind feature

We decided that the operator should watch every namespace, therefore we can deploy profiles from every namespace.

Fixes #40 

Different to what we discussed by opted for not pushing the profiles of the "seccomp-operator" namespace in to the top-level directory. They'll end up in a namespace dir just as any other profile. IMO handling the "seccomp-operator" differently would just make the implementation more difficult without adding too much value.

```release-note
seccomp profiles can be created in any namespace now.  Profiles end up in different subdirectories per namespace.
```

### DIFF
--- a/cmd/seccomp-operator/main.go
+++ b/cmd/seccomp-operator/main.go
@@ -30,9 +30,8 @@ import (
 )
 
 const (
-	appName             string = "seccomp-operator"
-	namespaceToWatchKey string = "NAMESPACE_TO_WATCH"
-	jsonFlag            string = "json"
+	appName  string = "seccomp-operator"
+	jsonFlag string = "json"
 )
 
 var (
@@ -98,23 +97,18 @@ func run(*cli.Context) error {
 	)
 	cfg, err := ctrl.GetConfig()
 	if err != nil {
-		return errors.Wrap(err, "cannot get config")
+		return errors.Wrap(err, "get config")
 	}
 
-	opts := ctrl.Options{
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{
 		SyncPeriod: &sync,
-	}
-	if os.Getenv(namespaceToWatchKey) != "" {
-		opts.Namespace = os.Getenv(namespaceToWatchKey)
-	}
-
-	mgr, err := ctrl.NewManager(cfg, opts)
+	})
 	if err != nil {
-		return errors.Wrap(err, "cannot create manager")
+		return errors.Wrap(err, "create manager")
 	}
 
 	if err := profile.Setup(mgr, ctrl.Log.WithName("profile")); err != nil {
-		return errors.Wrap(err, "cannot setup profile controller")
+		return errors.Wrap(err, "setup profile controller")
 	}
 
 	setupLog.Info("starting manager")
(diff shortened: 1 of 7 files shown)


cmd/seccomp-operator/main.go
```diff
@@ -30,10 +30,9 @@ import (
 )
 
 const (
-	appName                 string = "seccomp-operator"
-	namespaceToWatchDefault string = appName
-	namespaceToWatchKey     string = "NAMESPACE_TO_WATCH"
-	jsonFlag                string = "json"
+	appName             string = "seccomp-operator"
+	namespaceToWatchKey string = "NAMESPACE_TO_WATCH"
```
saschagrunert, line 34: Do we wanna remove this completely?

internal/pkg/controllers/profile/profile.go
```diff
@@ -149,8 +150,13 @@ func getProfilePath(profileName string, cfg *corev1.ConfigMap) (string, error) {
 		return "", errors.New(errConfigMapWithoutName)
 	}
 
+	if cfg.ObjectMeta.Namespace == "" {
+		return "", errors.New(errConfigMapWithoutNamespace)
```
saschagrunert, line 154: I don’t think that this is possible in general.

rhafer:
Marked this as a Draft for now as I'll be afk for some time. @saschagrunert  will hopefully take over from here :wink: otherwise I'll followup when I am back.

saschagrunert:
> Marked this as a Draft for now as I'll be afk for some time. @saschagrunert will hopefully take over from here 😉 otherwise I'll followup when I am back.

Sure thing. 👍

cmd/seccomp-operator/main.go
rhafer: Yeah, good question. I just left it intact for now because I couldn't answer that myself ;)

internal/pkg/controllers/profile/profile.go
rhafer: Actually, yeah. I think it defaults to `default`.

I guess that makes the check for `name` being enpty moot as well. Cause I think name can't be empty either.

saschagrunert: Yep, empty namespaces should be forbidden by the API server.

saschagrunert:
Ready for review, PTAL @pjbgf @hasheddan

cmd/seccomp-operator/main.go
saschagrunert: Removed it. :+1:

saschagrunert:
The go mod download seems to flake from time to time

/retest

deploy/operator.yaml
```diff
@@ -29,7 +29,7 @@ subjects:
   name: seccomp-operator
   namespace: seccomp-operator
 roleRef:
-  kind: Role
+  kind: ClusterRole
   name: config-map-reader
   apiGroup: rbac.authorization.k8s.io
```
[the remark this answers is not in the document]
saschagrunert: Yes, I agree. We probably have to think about configuration of the operator in a next step.

PR #52, opened 2020-07-31
Add operator seccomp profile

### LINKED ISSUE #29: Provide seccomp profile for the operator

I think a default profile for the operator itself could be a good starting point for a demo. It should be highly restrictive and we could create some documentation around how to build such profiles as well as providing additional application profiles out of the box. 

#### Why is this needed:

We should aim for the maximum amount of security for the operator deployment, a seccomp profile will help with that.

pjbgf:
#### What type of PR is this?

/kind feature

Creates a seccomp profile that is specific to the seccomp-operator.

Fixes #29

For a seccomp profile to be applied it first need to exist on disk, therefore such a task needs to be carried out within the init container.

```release-note
Operator now runs under a specific seccomp profile.
```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -36,6 +36,85 @@ roleRef:
   name: config-map-reader
   apiGroup: rbac.authorization.k8s.io
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: seccomp-operator-profile
+  namespace: seccomp-operator
+data:
+  seccomp-operator.json: |-
+    {
+      "defaultAction": "SCMP_ACT_ERRNO",
+      "architectures": [
+          "SCMP_ARCH_X86_64",
+          "SCMP_ARCH_X86",
+          "SCMP_ARCH_X32"
+      ],
+      "syscalls": [
+          {
+              "names": [
+                "accept4",
+                "arch_prctl",
+                "bind",
+                "brk",
+                "clone",
+                "close",
+                "connect",
+                "epoll_create1",
+                "epoll_ctl",
+                "epoll_pwait",
+                "execve",
+                "exit",
+                "exit_group",
+                "fcntl",
+                "fstat",
+                "futex",
+                "getcwd",
+                "getgid",
+                "getpeername",
+                "getpgrp",
+                "getpid",
+                "getppid",
+                "getrandom",
+                "getsockname",
+                "getsockopt",
+                "gettid",
+                "getuid",
+                "listen",
+                "madvise",
+                "membarrier",
+                "mkdirat",
+                "mlock",
+                "mmap",
+                "mprotect",
+                "nanosleep",
+                "newfstatat",
+                "open",
+                "openat",
+                "pipe2",
+                "pread64",
+                "read",
+                "readlinkat",
+                "rt_sigaction",
+                "rt_sigprocmask",
+                "rt_sigreturn",
+                "sched_getaffinity",
+                "sched_yield",
+                "setgid",
+                "setsockopt",
+                "set_tid_address",
+                "setuid",
+                "sigaltstack",
+                "socket",
+                "tgkill",
+                "uname",
+                "write"
+              ],
+              "action": "SCMP_ACT_ALLOW"
+          }
+      ]
+    }
+---
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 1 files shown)


pjbgf:
@hasheddan @saschagrunert what are your thoughts about merging this? if we do it now, as we develop features we may experience some problems locally which would mean we would have to amend the profile as we go.

Otherwise I am also happy to just wait until we are feature complete to reassess the profile.

deploy/operator.yaml
```diff
@@ -33,6 +33,75 @@ roleRef:
   name: config-map-reader
   apiGroup: rbac.authorization.k8s.io
 ---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: seccomp-operator-profile
+  namespace: seccomp-operator
+data:
+  seccomp-operator-v0.1.json: |-
```
saschagrunert, line 42: What was the reason the version this? Do you think we can omit the version string here?

deploy/operator.yaml
```diff
@@ -46,6 +115,7 @@ spec:
     metadata:
       annotations:
         seccomp.security.alpha.kubernetes.io/pod: runtime/default
+        container.seccomp.security.alpha.kubernetes.io/seccomp-operator: localhost/seccomp-operator-v0.1.json
```
hasheddan, line 118: Seems like we should probably have deployment options for both alpha and GA seccomp (though alpha will always work for now). Also, I think we should make it easy for folks to override the operator's profile in case it breaks something for their container runtime (not sure how likely this is to happen?). Both of these could be handled in a follow-up though.

deploy/operator.yaml
pjbgf: The idea behind this was to avoid overwriting profiles when deploying. On such instances a rollback would still have side-effects (specially if the new profile removed allowed syscalls).

I think we can remove that version for the time being, and do something around those lines when a new profile is not backwards compatible. What do you reckon?

deploy/operator.yaml
pjbgf: I completely agree with the GA syntax, just thought that doing that now would mean that we either would 1) only support v1.19+ or 2) would have to maintain two deployment files. So thought that perhaps it was better to defer that decision to later on with follow-up PRs.

In terms of overriding the operator profile, that is an interesting point. Currently the operator's profile is isolated, which allows for a cluster admin to isolate it in such a way that cluster's users can't fiddle with it. This may require changes if the community has some different use cases.

deploy/operator.yaml
saschagrunert: I agree. OTOH could versionize this profile on its own, so a v1 should a should probably also work. I’m not opinionated any more and leave it up to you.

pjbgf:
@saschagrunert I put the profile in "complain mode", the issue may be related to a specific syscall being blocked. I will try to reproduce the error to amend it.

saschagrunert:
Ah okay there is an issue in the DaemonSet:
```
The DaemonSet "seccomp-operator" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "host-operator-vol"
    suite_test.go:121: 
```

pjbgf:
> Ah okay there is an issue in the DaemonSet:
> 
> ```
> The DaemonSet "seccomp-operator" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: "host-operator-vol"
>     suite_test.go:121: 
> ```

Thank you for the heads up. There were also a few syscalls missing that I added now.

pjbgf:
@hasheddan @saschagrunert This profile worked fine on `Ubuntu 18.04` and `Ubuntu 20.04`. PTAL

Are we happy to go ahead with this in "enforce mode"? As we develop new features this may break the build, in which case we may need to check the missing syscalls:

`cat /var/log/syslog | grep audit | grep seccomp-operator`

I wonder whether is worth adding something to that effect on the E2E tests.

saschagrunert:
> @hasheddan @saschagrunert This profile worked fine on `Ubuntu 18.04` and `Ubuntu 20.04`. PTAL
> 
> Are we happy to go ahead with this in "enforce mode"? As we develop new features this may break the build, in which case we may need to check the missing syscalls:
> 
> `cat /var/log/syslog | grep audit | grep seccomp-operator`
> 
> I wonder whether is worth adding something to that effect on the E2E tests.

I think that's fine, we have the CI to get early feedback and adjust the profile accordingly during development :+1:

saschagrunert:
Ah snap this needs another rebase :-/

pjbgf:
rebased: PTAL

PR #54, opened 2020-08-03
Add default profile generation into operator deployment

saschagrunert:
#### What type of PR is this?

/kind feature

We now generate the default profiles from the `profiles/` directory into
the operator deployment. This includes an additional verification step
in the CI, where we check if the profiles in the deployment are
up-to-date.

Fixes https://github.com/kubernetes-sigs/seccomp-operator/issues/42

None

```release-note
- Add default operator profiles to the deployment. For now we added an nginx:1.19.1 profile
```


saschagrunert:
Working on a fix for:
```
The ConfigMap "default-profiles" is invalid: data[nginx:1.19.1.json]: Invalid value: "nginx:1.19.1.json": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
```

deploy/operator.yaml
```diff
@@ -111,4 +111,15 @@ spec:
...
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: seccomp-operator
+  name: default-profiles
+  annotations:
+    seccomp.security.kubernetes.io/profile: "true"
+data:
+  nginx-1.19.1.json: |-
```
hasheddan, line 124: I have mentioned this in a few PR's, but we should think about making some of the deployment (such as the install of default profiles) configurable 👍

Makefile
```diff
@@ -100,6 +104,10 @@ $(BUILD_DIR)/verify_boilerplate.py: $(BUILD_DIR)
 verify-go-mod: go-mod ## Verify the go modules
 	hack/tree-status
 
+.PHONY: verify-default-profiles
+verify-default-profiles: default-profiles ## Verify the generated default profiles
```
hasheddan, line 108: Would it make sense to clear the profiles before regenerating them? That would guard against the case where the profiles were manually updated to be correct, but the generator was actually failing to run or not changing anything

profiles/build.go
```diff
@@ -0,0 +1,123 @@
...
+
+type matcher struct {
+	foundConfigMapType bool
+	foundConfigMapName bool
+	foundData          bool
+	replace            bool
+}
+
+func main() {
+	file, err := os.Open(deploymentYAML)
```
hasheddan, line 47: I might take a pass after this is merged and working to see if we can make this a more general purpose tool for converting any seccomp profiles to a `ConfigMap`

deploy/operator.yaml
saschagrunert: I'll follow-up on that. I have no clue how we wanna do that without adding another level of abstraction. Do we wanna make it configurable via environment variables, Helm or a config file? I think a config file could be a good first step, which could be deployed in parallel to the deploy/operator.yaml. WDYT?

profiles/build.go
saschagrunert: Yeah that would be cool, thank you! :pray:

Makefile
saschagrunert: Hm, we indirectly clear them by regeneration all at once. If the generation outputs the same as the state of the repository, then we should be fine in any case. :shrug:

saschagrunert:
Thanks for the review @hasheddan, let's iterate on the configuration and the tooling. :pray: 
/hold cancel

PR #56, opened 2020-08-03
Add initial eventing machinery

### LINKED ISSUE #41: Surface events for seccomp profiles created via ConfigMaps
Users should be made aware if their seccomp profile was successfully added to nodes in the cluster or not.

hasheddan:
#### What type of PR is this?

/kind feature

Adds event recorder and emits warning events on errors.

Signed-off-by: hasheddan <georgedanielmangum@gmail.com>

Fixes #41 

```release-note
Profile controller will emit warning events on failure to get profile path or save profile to disk
```


go.sum
```diff
@@ -133,6 +133,7 @@ github.com/go-logfmt/logfmt v0.4.0/go.mod h1:3RMwSq7FuexP4Kalkev3ejPJsZTpXXBr9+V
 github.com/go-logr/logr v0.1.0 h1:M1Tv3VzNlEHg6uyACnRdtrploV2P7wZqH8BoQMtz0cg=
 github.com/go-logr/logr v0.1.0/go.mod h1:ixOQHD9gLJUVQQ2ZOR7zLEifBX6tGkNJF4QyIY7sIas=
 github.com/go-logr/zapr v0.1.0/go.mod h1:tabnROwaDl0UNxkVeFRbY8bwB37GwRv0P8lg6aAiEnk=
+github.com/go-logr/zapr v0.1.1 h1:qXBXPDdNncunGs7XeEpsJt8wCjYBygluzfdLO0G5baE=
```
hasheddan, line 136: hmmm not sure how these `go.sum` changes happened. `make-verify` passed locally for me though :thinking:

internal/pkg/controllers/profile/profile.go
```diff
@@ -116,11 +119,13 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 		profilePath, err := getProfilePath(name, profile)
 		if err != nil {
 			logger.Error(err, "cannot get profile path")
+			r.record.Event(profile, event.Warning(event.Reason("cannot get profile path"), err))
```
pjbgf, line 122: What's your thoughts on adding this into a new implementation of `logger`?

hasheddan: And having a combined logger / eventer? So it would look something like `logger.Event()`?

hasheddan: Just noticed the passed in logger is no longer being used

saschagrunert: Logging an error automatically throws an event? Sound good! :)

hasheddan: Hmm I'm not sure we will always want to have our logging and eventing tied together in that manner. @pjbgf did you have a specific use case in mind?

pjbgf: The idea was just to remove repetition as I can easily imagine those two lines being used quite frequently together.
I thought more like an EventLogger that implemented the same interface as logger but had an extra side effect.
Thinking a bit more about it, this would violate SRP and probably would make things a bit more complex for new contributors to understand.

Let's keep it simple for now, if needs be we can always review and refactor. ;) 

/lgtm

PR #59, opened 2020-08-04
Fix permission issues when kubelet folder is set to 0700

pjbgf:
#### What type of PR is this?

/kind bug

Add support for when `/var/lib/kubelet` is set to chmod `0700`.

This is related to https://github.com/kubernetes/kubernetes/issues/93652

None

```release-note
NONE
```

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -53,27 +53,29 @@ spec:
       initContainers:
         - name: non-root-enabler
           image: busybox
-          # Creates folder /var/lib/kubelet/seccomp/operator and sets 2000:2000 as its owner.
+          # Creates folder /var/lib/seccomp-operator, sets 2000:2000 as its owner and symlink it to /var/lib/kubelet/seccomp/operator.
           # This is required to allow the main container to run as non-root.
           command: ["/bin/sh", "-c"]
           args:
             - >
-              KUBELET_DIR=/var/lib/kubelet &&
-              OPERATOR_DIR=$KUBELET_DIR/seccomp/operator &&
-              chmod +x $KUBELET_DIR &&
-              if [ ! -d $OPERATOR_DIR ]; then
-                /bin/mkdir -m 0744 -p $OPERATOR_DIR &&
-                /bin/chown -R 2000:2000 $OPERATOR_DIR;
-              fi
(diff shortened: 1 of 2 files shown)


deploy/operator.yaml
```diff
@@ -53,7 +53,7 @@ spec:
       initContainers:
         - name: non-root-enabler
           image: busybox
-          # Creates folder /var/lib/kubelet/seccomp/operator and sets 2000:2000 as its owner.
+          # Creates folder /var/lib/seccomp-operator, sets 2000:2000 as its owner and symlink it to /var/lib/kubelet/seccomp/operator.
           # This is required to allow the main container to run as non-root.
           command: ["/bin/sh", "-c"]
           args:
```
saschagrunert, line 59: Oops, the linking seems missing now :see_no_evil: 
/hold

pjbgf: Yep, it was an issue on conflict resolution.. PTAL

saschagrunert:
Alright, I think the e2e tests have to be fixed to check for the right paths there:
https://github.com/kubernetes-sigs/seccomp-operator/blob/f8af1bb40e6e3cdb15bfde0d0b38541c582bdc34/test/e2e_test.go#L71-L74

pjbgf:
> Alright, I think the e2e tests have to be fixed to check for the right paths there:
> https://github.com/kubernetes-sigs/seccomp-operator/blob/f8af1bb40e6e3cdb15bfde0d0b38541c582bdc34/test/e2e_test.go#L71-L74

Thank you for pointing out, I think the issue may be that we need to follow the symlinks, so I added the `-L` switch.

test/e2e_test.go
```diff
@@ -87,7 +87,7 @@ func (e *e2e) TestSeccompOperator() {
 		// General path verification
 		e.logf("Verifying seccomp operator directory on node: %s", node)
 		statOutput := e.execNode(
-			node, "stat", "-c", `%a,%u,%g`, profile.DirTargetPath(),
+			node, "stat", "-L", "-c", `%a,%u,%g`, profile.DirTargetPath(),
 		)
 		e.Contains(statOutput, "744,2000,2000")
```
saschagrunert, line 92:
```suggestion
		e.Contains(statOutput, "755,2000,2000")
```

PR #62, opened 2020-08-04
Fix double symlink creation on operator restart

saschagrunert:
#### What type of PR is this?

/kind bug

If we do not check the symlink before creating it we might be end-up
creating another symlink inside of the already linked directory.

The issue can be reproduced by deploying the operator and removing one
of the pods. Before this patch the init would fail.

None

None

```release-note
None
```


saschagrunert:
> /lgtm
> 
> I may circle back here and add an e2e test for this +1

Yes this would be very nice! :)

saschagrunert:
Timeout
/retest

Issue #66, opened 2020-08-04
Validate profiles before writing them to disk


saschagrunert:
We could add a pre validation step to parse the seccomp profiles before writing them to disk. This way we can ensure that only valid profiles are written down. I would report errors with events as well as directly at reconciliation level. WDYT? Would this be suitable for v1?

/area security

Refers to https://github.com/seccomp/containers-golang/issues/34

saschagrunert:
Let's close this now
/close

PR #72, opened 2020-08-05
Add basic seccomp profile validation

saschagrunert:
#### What type of PR is this?

/kind feature

We now unmarshal the profile to do a basic validation. The build will
now change to include the `seccomp` build tag required for the new
dependency.

I also left a `TODO` to consider further validation steps later on.

Refers to #66 

One drawback is now that we need a static libseccomp.a to build locally.

```release-note
- Added basic seccomp profile validation before syncing them on disk
```

### DIFF
--- a/internal/pkg/controllers/profile/profile.go
+++ b/internal/pkg/controllers/profile/profile.go
@@ -18,6 +18,7 @@ package profile
 
 import (
 	"context"
+	"encoding/json"
 	"io/ioutil"
 	"os"
 	"path"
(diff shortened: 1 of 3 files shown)


saschagrunert:
Rebased to resolve the conflicts.

saschagrunert:
> @saschagrunert if we are only doing validation by marshaling right now does it make sense to introduce this dependency? We could alternatively just add the types here: https://github.com/seccomp/containers-golang/blob/master/types.go

Yeah I think this would make sense. How about now? I guess an e2e test would be sufficient, too :D

PR #77, opened 2020-08-06
Initialize the operator with a dedicated subcommand

saschagrunert:
#### What type of PR is this?

/kind feature

Instead of relying on a bunch of bash code we now write the
initialization phase directly as operator subcommand. This way we can
reuse the already built image and test the code paths.

A new `config` package now contains global consts shared across the
whole application. This way we can decouple it from the `profile`
package.

None

None

```release-note
- Reuse operator container image for rootless initialization of the deployment
```

### DIFF
--- a/cmd/seccomp-operator/main.go
+++ b/cmd/seccomp-operator/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/pkg/errors"
@@ -25,14 +26,13 @@ import (
 	"k8s.io/klog/klogr"
 	ctrl "sigs.k8s.io/controller-runtime"
 
+	"sigs.k8s.io/seccomp-operator/internal/pkg/config"
 	"sigs.k8s.io/seccomp-operator/internal/pkg/controllers/profile"
+	"sigs.k8s.io/seccomp-operator/internal/pkg/initialize"
 	"sigs.k8s.io/seccomp-operator/internal/pkg/version"
 )
 
-const (
-	appName  string = "seccomp-operator"
-	jsonFlag string = "json"
-)
+const jsonFlag string = "json"
 
 var (
 	sync     = time.Second * 30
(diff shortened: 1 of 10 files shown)


cmd/seccomp-operator/main.go
```diff
@@ -75,6 +75,19 @@ func main() {
 				return nil
 			},
 		},
+		&cli.Command{
+			Name:    "initialize",
+			Aliases: []string{"init", "i"},
+			Usage:   "initialize the operator directories and set the right permissions",
+			Action: func(c *cli.Context) error {
+				return errors.Wrap(initialize.SetupRootless(
+					config.KubeletSeccompRootPath,
```
hasheddan, line 84: Might be nice to be able to override this if desired. I'm fine with deferring that until later though 👍

internal/pkg/config/config.go
```diff
@@ -0,0 +1,37 @@
...
+*/
+
+package config
+
+const (
+	// OperatorName is the name when referring to the operator.
+	OperatorName = "seccomp-operator"
+
+	// OperatorNamespace is the default namespace for the operator deployment.
+	OperatorNamespace = OperatorName
```
hasheddan, line 24: How would you feel about moving theses in `test/`? It appears that is the only place they are used and they could easily get out of sync with the actual deployment if the values were changed there. I would prefer not to the illusion that these values are always correct and instead use the downward API in places where these values are needed within the operator.

hasheddan: Happy to follow with this if you prefer :)

saschagrunert: Yep, I completely removed it and sicked to `OperatorName` in the `test` package.

cmd/seccomp-operator/main.go
saschagrunert: Yeah, I can follow-up on that. We could add a `Config` structure to the `config` package and let it be populated by CLI flags and/or environment variables. Then we could pass them there…

internal/pkg/config/config.go
hasheddan: Hmm I think I have the same feeling about `OperatorName` it is not strictly guaranteed to be correct, so it seems safer to be in `test/`

saschagrunert:
@hasheddan do we wanna get this in? I can work on the config enhancements afterwards. :upside_down_face:

PR #81, opened 2020-08-07
Revert init command and stick to simple bash image

saschagrunert:
#### What type of PR is this?
/kind cleanup

To reduce the attack surface of the application we now stick to a simple
bash script init instead of the dedicated subcommand.

None

Partially reverts https://github.com/kubernetes-sigs/seccomp-operator/pull/77

```release-note
None
```


saschagrunert:
> @saschagrunert is main concern here base image vulnerabilities?

I think it's the shared codebase which could case. Having the init container (more privileged) coupled to the operator seems to increase the general attack surface.

I'm pretty sure @pjbgf has some other vulnerable cases in mind :innocent:

pjbgf:
It is in moments like this that people realise how unpopular and unsexy security can be. =)

By having an init container to do the "privileged tasks", we reduced the time in which we will be running as `root`, on every single node and with write access to `/var/lib`, which is a great thing. However, we must remember that during this time window we cannot apply any security controls over it (i.e. seccomp, apparmor, etc).

Now what we execute within that opportunity window also makes a difference. It should be drastically smaller the attack surface of executing 4 pretty stable linux commands (mkdir, chmod, ln and chown) running inside a standard image, when compared to a custom application with [hundreds of dependencies](https://github.com/kubernetes-sigs/seccomp-operator/blob/master/go.sum) (considering dependencies of dependencies too) that may be in active development. 

In terms of "what is the worse that a dependency can do", any of them, even without being directed called from our code could hijack the execution: [here are a few examples](https://itnext.io/golang-stop-trusting-your-dependencies-a4c916533b04). 
So realistically, it would only take a single one of our dependencies to be misbehave for the impact to be quite serious.

Issue #76, opened 2020-08-05
Check if a node has seccomp support. 


saschagrunert:
We could verify if a node supports seccomp by:

```go
func IsEnabled() bool {
	// Check if Seccomp is supported, via CONFIG_SECCOMP.
	if err := unix.Prctl(unix.PR_GET_SECCOMP, 0, 0, 0, 0); err != unix.EINVAL {
		// Make sure the kernel has CONFIG_SECCOMP_FILTER.
		if err := unix.Prctl(unix.PR_SET_SECCOMP, unix.SECCOMP_MODE_FILTER, 0, 0, 0); err != unix.EINVAL {
			return true
		}
	}
	return false
}
```

The question is, what would we do in the case that this function returns `false`? We could make the reconciliation a noop, throw a warning event. Right now this would not bring much benefit to the user. How could we prevent that a workload with a profile gets scheduled on a node which is not supporting seccomp?

hasheddan:
> How could we prevent that a workload with a profile gets scheduled on a node which is not supporting seccomp?

Seems like a good use-case for an admission webhook.

saschagrunert:
> > How could we prevent that a workload with a profile gets scheduled on a node which is not supporting seccomp?
> 
> Seems like a good use-case for an admission webhook.

Yeah, so probably a good topic for post v1.

pjbgf:
I think if it returns `false` we could tag the node with something meaningful, which could then be used with [node selectors](https://raw.githubusercontent.com/kubernetes/website/master/content/en/examples/pods/pod-nginx.yaml).
If we have the capacity, we potentially could do now:

1. Implement the validation and tagging.
2. Document so users can already benefit from being able to not schedule pods that must run with seccomp.

And then once the admission webhook is done:

3. Mutate pods to automatically have the node selector based on some configuration or tag/annotation (TBC).

Or we could just do three of them post v1.

saschagrunert:
Sounds good, let's move it to v2 to decrease the complexity.

saschagrunert:
#85 contains a first RFC, but does not necessarily land in v1.

hasheddan:
@saschagrunert do you know how the kubelet handles if a pod is scheduled to it with a seccomp profile and seccomp is not enabled? I would imagine it likely behaves similar to if the profile does not exist?

saschagrunert:
[...]

PR #85, opened 2020-08-08
Verify if nodes do support seccomp and do not reconcile in that case

saschagrunert:
#### What type of PR is this?

/kind design
/kind feature

The new `seccomp` package can be used to check if seccomp is enabled on
the host.

Refers to #76 

This is a RFC to ensure that we're moving into the right direction. If we wanna use something like this directly in the operator, then we could throw an event on startup to inform the user if the node supports seccomp or not.

To be able to throw the event for the node, the operator would have to require permissions to list the nodes on the cluster. The node name could be injected via an environment variable, whereas the client could retrieve the node object.

```release-note
- Nodes not supporting seccomp will not reconcile profiles to disk. Additionally a warning event will be thrown for the config map.
```


internal/pkg/controllers/profile/profile.go
```diff
@@ -108,23 +110,35 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+	if !seccomp.New().IsSupported(logger) {
+		err := errors.New("profile not added")
+		reason := fmt.Sprintf(
+			"node %q does not support seccomp",
+			os.Getenv(config.NodeNameEnvKey),
+		)
+		logger.Error(err, "reason", reason)
+		r.record.Event(configMap, event.Warning(event.Reason(reason), err))
+		return reconcile.Result{}, errors.Wrap(err, reason)
+	}
```
saschagrunert, line 123: _Following-up on discussion https://github.com/kubernetes-sigs/seccomp-operator/pull/85#pullrequestreview-463875699_

@hasheddan like this? Do you think we should put the operator in some `excluded` state to not spam the events? This way we could stop the operator reconciling if seccomp is disabled on a node…

hasheddan: @saschagrunert definitely 👍 If we just don't return the error here then it should stop trying to reconcile. controller-runtime behaves in the following manner:
1. If an `error` is returned, requeue immediately
2. If `reconcile.Result{ Requeue: true }`, requeue immediately
3. If `reconcile.Result{ RequeueAfter: someTime }`, requeue after `someTime` has elapsed
4. If `reconcile.Result{}, nil`, do not requeue (will be requeued if a change to the object is observed, or after the usually very long reconcile timeout configured for the controller manager)

saschagrunert: Awesome, thank you for the clarification! :+1:

internal/pkg/controllers/profile/profile.go
```diff
@@ -90,13 +92,18 @@ func Setup(mgr ctrl.Manager, l logr.Logger) error {
...
+	log             logr.Logger
+	record          event.Recorder
+	stopReconciling bool
 }
 
 // Reconcile reconciles a ConfigMap representing a seccomp profile.
 func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+	if r.stopReconciling {
+		return reconcile.Result{}, nil
+	}
```
saschagrunert, line 105: @hasheddan do you think it makes sense to generally stop reconciling like this? The only way to restart the reconciling would be to restart the workload now.

hasheddan: I don't think the benefit of skipping running the checks on the node on very infrequent reconciles is worth the added complexity of managing this / having to restart the workload

saschagrunert: Sounds good, let me revert that change.

saschagrunert:
/hold
If we can move forward with https://github.com/seccomp/containers-golang/pull/38, then I'll rebase my PR on top of that.

pjbgf:
> @saschagrunert this is looking pretty good to me! I wonder if instead of eventing on the node we could just event on the profile `ConfigMap` with something like `Profile not added to node ____ because seccomp is not enabled`. It seems more likely that someone would be looking there than looking for events on the node after the operator was installed. What do you think?

I have mixed feeling on this: I love the traceability it gives us, only slightly concerned how noisy this may get in very large deployments (hundreds of nodes). However, I am happy with that approach for the time being and we revisit it once we land the CRD.

internal/pkg/node/seccomp/seccomp.go
```diff
@@ -0,0 +1,132 @@
...
+
+//counterfeiter:generate . verifer
+type verifer interface {
+	ParseStatusFile(logr.Logger, string) (map[string]string, error)
+	Prctl(int, uintptr, uintptr, uintptr, uintptr) error
+}
+
+// Seccomp is the main structure of this package.
+type Seccomp struct {
+	verifier verifer
```
pjbgf, line 42:
```suggestion
	verifier verifier
```

internal/pkg/node/seccomp/seccomp.go
```diff
@@ -0,0 +1,132 @@
...
+	perrors "github.com/pkg/errors"
+	"golang.org/x/sys/unix"
+)
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+const statusFilePath = "/proc/self/status"
+
+//counterfeiter:generate . verifer
+type verifer interface {
```
pjbgf, line 35:
```suggestion
type verifier interface {
```

internal/pkg/node/seccomp/seccomp.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ParseStatusFile is the status file parse implementation for the default
+// verifer.
+func (d *defaultVerifier) ParseStatusFile(
+	l logr.Logger, filePath string,
+) (map[string]string, error) {
+	return ParseStatusFile(l, filePath)
+}
+
+// Prctl is the Prctl implementation for the default verifer.
```
pjbgf, line 61:
```suggestion
// Prctl is the Prctl implementation for the default verifier.
```

internal/pkg/node/seccomp/seccomp.go
```diff
@@ -0,0 +1,132 @@
...
+func (d *defaultVerifier) Prctl(
+	option int, arg2, arg3, arg4, arg5 uintptr,
+) error {
+	return unix.Prctl(option, arg2, arg3, arg4, arg5)
+}
+
+// SetVerifier can be used to replace the internal verifier implementation.
+func (s *Seccomp) SetVerifier(v verifer) {
```
pjbgf, line 69:
```suggestion
func (s *Seccomp) SetVerifier(v verifier) {
```

pjbgf:
> /hold
> If we can move forward with [seccomp/containers-golang#38](https://github.com/seccomp/containers-golang/pull/38), then I'll rebase my PR on top of that.

@saschagrunert given the value this adds, I would not be opposed to merge this and then have a follow-up PR to use the `IsSupported` from upstream when that gets merged.

saschagrunert:
> > /hold
> > If we can move forward with [seccomp/containers-golang#38](https://github.com/seccomp/containers-golang/pull/38), then I'll rebase my PR on top of that.
> 
> @saschagrunert given the value this adds, I would not be opposed to merge this and then have a follow-up PR to use the `IsSupported` from upstream when that gets merged.

Okay cool, let me just rebase and fix the typos :+1:

saschagrunert:
/retitle Verify if nodes do support seccomp and do not reconcile in that case

PR #86, opened 2020-08-10
Remove internal seccomp types in favor of containers-golang package

saschagrunert:
#### What type of PR is this?

/kind cleanup

The package types are available even without having the `seccomp` build
tag specified. This means we can use them safely by relying on the
dependency.

None

None

```release-note
None
```


go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/crossplane/crossplane-runtime v0.9.0
 	github.com/go-logr/logr v0.2.1-0.20200730175230-ee2de8da5be6 // TODO: switch to v0.2.1 if released
 	github.com/pkg/errors v0.9.1
+	github.com/seccomp/containers-golang v0.6.0
```
hasheddan, line 9: @saschagrunert if we expanded usage of this package in the future could it result in requiring the build tags? If so, might be good to leave a note somewhere if possible

saschagrunert: Yep, added not about the build tag.

saschagrunert:
@hasheddan wanna move forward with this change? :innocent:

PR #89, opened 2020-08-11
Verify that invalid profile gets not reconciled on disk

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now additionally verify that the invalid profile does not exist on
the node in the corresponding e2e test.

None

As a follow-up I'd like to re-patch the configmap to be valid again and then check if it gets reconciled to disk.

```release-note
None
```

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -19,8 +19,11 @@ limitations under the License.
 package e2e_test
 
 import (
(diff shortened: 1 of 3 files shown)


saschagrunert:
@hasheddan @pjbgf PTAL

PR #92, opened 2020-08-12
Refactor event creating into own method

saschagrunert:
#### What type of PR is this?

/kind cleanup

To reduce code duplication we now use a new method on the reconciler to
push the events for configMaps.

None
None

```release-note
None
```

### DIFF
--- a/internal/pkg/controllers/profile/profile.go
+++ b/internal/pkg/controllers/profile/profile.go
@@ -110,29 +110,34 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) 
 
 	for profileName, profileContent := range configMap.Data {
 		if err := validateProfile(profileContent); err != nil {
-			reason := "cannot validate profile " + profileName
-			logger.Error(err, reason)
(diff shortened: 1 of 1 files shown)


internal/pkg/controllers/profile/profile.go
```diff
@@ -110,23 +110,19 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
 
 	for profileName, profileContent := range configMap.Data {
 		if err := validateProfile(profileContent); err != nil {
-			reason := "cannot validate profile " + profileName
-			logger.Error(err, reason)
-			r.record.Event(configMap, event.Warning(event.Reason(reason), err))
+			r.pushEventForConfigMap(logger, configMap, "cannot validate profile", err)
```
hasheddan, line 113: Hmm currently we are only emitting events for errors, but I don't think this would be the case in the future. I am inclined to keep the code a bit more verbose since this change is only removing one line in each instance. What do you think?

saschagrunert: Hm yeah I agree. Let's close this for now :+1:

PR #101, opened 2020-08-13
Update eventing and reconcile logic

hasheddan:
#### What type of PR is this?

/kind feature

Updates eventing style to be more aligned with Kubernetes style. Also avoids constant reconciles when there is an error writing a profile to disk.

None

Also reduced `wait` to 30 seconds now that we are only using it in error cases.

```release-note
Do not requeue after successfully writing profile to disk and do not immediately requeue on errors.
```


saschagrunert:
Ah okay let’s fix the lint issue and tests. :)

hasheddan:
@saschagrunert I think this is ready now :)

PR #105, opened 2020-08-14
Remove leading `v` from image version

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should not add leading `v`s to container images.
None

```release-note
None
```

### DIFF
--- a/deploy/namespace-operator.yaml
+++ b/deploy/namespace-operator.yaml
@@ -187,7 +187,7 @@ spec:
               ephemeral-storage: "50Mi"
       containers:
         - name: seccomp-operator
-          image: k8s.gcr.io/seccomp-operator/seccomp-operator:v0.1.0
+          image: k8s.gcr.io/seccomp-operator/seccomp-operator:0.1.0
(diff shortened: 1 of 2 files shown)


deploy/namespace-operator.yaml
```diff
@@ -187,7 +187,7 @@ spec:
               ephemeral-storage: "50Mi"
       containers:
         - name: seccomp-operator
-          image: k8s.gcr.io/seccomp-operator/seccomp-operator:v0.1.0
+          image: k8s.gcr.io/seccomp-operator/seccomp-operator:0.1.0
```
hasheddan, line 190: Don't we tag k8s releases with a `v` prefix?

saschagrunert: Yes the tag will contain the prefix, but I prefer having the container image without it. I think it's more common. WDYT?

saschagrunert: Ah no I see we do it the same way in Kubernetes, which should be fine.

PR #114, opened 2020-08-20
Use kustomize for deployment generation

### LINKED ISSUE #107: Use kustomize for base deployment and derive namespaced one
The idea is to combine both deployment variants to increase their maintainability. 

We could create one kustomize `base` and change the namespaced deployment to be an overlay of the base. We also could generate those two deployments statically for the end users, so that they do not have to rely on the kustomize binary itself (like now, but with increased maintainability). The verify CI job can then check if the deployments are in sync.

Alternatively, we could also create a helm chart. WDYT?

saschagrunert:
#### What type of PR is this?

/kind feature

This is the first approach in splitting-up the deployment into a base
and an overlay for the namespaced deployment. Later on we can further
split-up the base and overwrite the image names for example.

Fixes #107 
None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,11 @@ go-mod: ## Cleanup and verify go modules
 default-profiles: ## Generate the default profiles
 	$(GO) run ./profiles
 
+.PHONY: deployments
+deployments: ## Generate the deployment files with kustomize
+	kustomize build --reorder=none deploy/base -o deploy/operator.yaml
+	kustomize build --reorder=none deploy/overlays/namespaced -o deploy/namespace-operator.yaml
+
 .PHONY: image
 image: ## Build the container image
 	$(CONTAINER_RUNTIME) build --build-arg version=$(VERSION) -t $(IMAGE) .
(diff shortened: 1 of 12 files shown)


deploy/operator.yaml
```diff
@@ -1,247 +1,271 @@
...
 kind: ConfigMap
 metadata:
-  name: seccomp-operator-profile
+  annotations:
+    seccomp.security.kubernetes.io/profile: "true"
+  labels:
+    app: seccomp-operator
+  name: default-profiles
   namespace: seccomp-operator
```
hasheddan, line 66: nitpick: any idea why the `ConfigMap` gets generated with `data` before `kind`? Is it just alphabetical?

saschagrunert: I think so, not sure if this is configurable by kustomize… :thinking:

Makefile
```diff
@@ -85,14 +85,19 @@ go-mod: ## Cleanup and verify go modules
 default-profiles: ## Generate the default profiles
 	$(GO) run ./profiles
 
+.PHONY: deployments
+deployments: ## Generate the deployment files with kustomize
+	kustomize build --reorder=none deploy/base -o deploy/operator.yaml
+	kustomize build --reorder=none deploy/overlays/namespaced -o deploy/namespace-operator.yaml
```
saschagrunert, line 91: I guess this reorder flag applies to the resources and not the fields… :thinking:

hasheddan: I am fine moving ahead as-is, then circling back to clean up, what do you think?

saschagrunert: I agree, thank you! :)

PR #121, opened 2020-08-28
Validate profile by building the seccomp filter

### LINKED ISSUE #119: Enhance the profile validation by re-using the libcontainer implementation
I'm working right now to move the validation code from runc (libcontainer) into [containers/common](https://github.com/containers/common), which could be then re-used by the operator to do a tighter validation of the profiles.

If we have a CRD, then we could think about possibilities to validate the profile on CRD creation.

saschagrunert:
#### What type of PR is this?

/kind feature

The new `containers/common` `ValidateProfile()` API can be used to build
a libseccomp filter without loading it into the kernel. This can be now
used by the operator to have a more in-depth profile validation. This
now requires the `seccomp` build tag and a (statically) compiled version
of libseccomp.

Fixes #119 
None

```release-note
- Changed profile validation to build the whole filter without loading it into the kernel
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@
 
 FROM golang:1.15-alpine AS build
 WORKDIR /work
-RUN apk --no-cache add build-base git gcc
+RUN apk --no-cache add build-base git gcc libseccomp-dev
 
 ENV USER=secuser
 ENV UID=2000
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,8 +1,9 @@
 ---
 run:
   build-tags:
-    - netgo
     - e2e
+    - netgo
+    - seccomp
   concurrency: 6
   deadline: 5m
 linters:
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ GIT_COMMIT := $(shell git rev-parse HEAD 2> /dev/null || echo unknown)
 GIT_TREE_STATE := $(if $(shell git status --porcelain --untracked-files=no),dirty,clean)
 VERSION := $(shell cat VERSION)
 
-BUILDTAGS := netgo
+BUILDTAGS := netgo seccomp
 BUILD_FILES := $(shell find . -type f -name '*.go' -or -name '*.mod' -or -name '*.sum' -not -name '*_test.go')
 GO_PROJECT := sigs.k8s.io/$(PROJECT)
 LDVARS := \
@@ -129,9 +129,9 @@ $(BUILD_DIR)/golangci-lint:
 
 .PHONY: test-unit
 test-unit: $(BUILD_DIR) ## Run the unit tests
-	$(GO) test -ldflags '$(LDVARS)' -v -test.coverprofile=$(BUILD_DIR)/coverage.out ./...
+	$(GO) test -ldflags '$(LDVARS)' -tags '$(BUILDTAGS)' -v -test.coverprofile=$(BUILD_DIR)/coverage.out ./...
 	$(GO) tool cover -html $(BUILD_DIR)/coverage.out -o $(BUILD_DIR)/coverage.html
 
 .PHONY: test-e2e
 test-e2e: ## Run the end-to-end tests
-	$(GO) test -timeout 20m -tags e2e -count=1 ./test/... -v
+	$(GO) test -timeout 20m -tags 'e2e $(BUILDTAGS)' -count=1 ./test/... -v
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module sigs.k8s.io/seccomp-operator
 go 1.15
 
 require (
-	github.com/containers/common v0.20.3-0.20200827091701-a550d6a98aa3 // can be built with the `seccomp` build tag to support more features
+	github.com/containers/common v0.21.0
 	github.com/crossplane/crossplane-runtime v0.9.0
 	github.com/go-logr/logr v0.2.1
 	github.com/pkg/errors v0.9.1
@@ -15,3 +15,5 @@ require (
 	k8s.io/release v0.4.0
 	sigs.k8s.io/controller-runtime v0.6.2
 )
+
+replace github.com/containers/common => github.com/saschagrunert/common v0.20.3-0.20200908080721-143cdae3b077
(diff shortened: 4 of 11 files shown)


saschagrunert:
Now it gets interesting:
```
2s          Warning   InvalidSeccompProfile     configmap/test-profile   build seccomp filter: create filter for default action Action: Log system call: could not create filter
```

Locally reproducible, but if I write a unit tests to init the seccomp filter with ACT_LOG, then it works…scary.

saschagrunert:
It's a permission problem, sure.

saschagrunert:
Looks like that this is the first time that the seccomp profile of the operator blocks a necessary syscall for the LOG action. @pjbgf do you have any hint which syscall we have to add here?

saschagrunert:
It was the `seccomp(2)` syscall which had to be added.

Dockerfile
```diff
@@ -14,7 +14,7 @@
 
 FROM golang:1.15-alpine AS build
 WORKDIR /work
-RUN apk --no-cache add build-base git gcc
+RUN apk --no-cache add build-base git gcc libseccomp-dev pkgconfig
```
hasheddan, line 17: What is `pkgconfig` used for here? Should it be `pkg-config`?

pjbgf:
> It was the `seccomp(2)` syscall which had to be added. Ready for review.

If this change requires the `seccomp` syscall, that means we are not only "building" the profile, but we are also applying it. Going through upstream code it does look like that: 
https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L82

On that note, as part of calling upstream we are also setting the filter's NoNewPrivs bit to false. I am not sure we would want to do that as a side effect of "building" the profile. What are your thoughts around this?
https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L72

Dockerfile
saschagrunert: The package is called `pkgconfig`, but the resulting binary is `pkg-config`, yeah :grin:. Looks like that it is not necessary at all any more, removing it.

saschagrunert:
> > It was the `seccomp(2)` syscall which had to be added. Ready for review.
> 
> If this change requires the `seccomp` syscall, that means we are not only "building" the profile, but we are also applying it. Going through upstream code it does look like that:
> https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L82

The funny thing is that it was only required by reconciling the profile with the contents of `{ "defaultAction": "SCMP_ACT_LOG" }`. I'm not exactly sure which libseccomp internals triggered this requirement.

> On that note, as part of calling upstream we are also setting the filter's NoNewPrivs bit to false. I am not sure we would want to do that as a side effect of "building" the profile. What are your thoughts around this?
> https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L72

We only set the bit to the new filter. Do you think that this already modifies some state of the currently running process? :thinking:

pjbgf:
> The funny thing is that it was only required by reconciling the profile with the contents of `{ "defaultAction": "SCMP_ACT_LOG" }`, tiggered by the e2e tests. I'm not exactly sure which libseccomp internals made this requirement.

My take is that by using `SCMP_ACT_LOG` in this context, we are logging but still allowing all the things we haven't allowed on the operator's seccomp. Reason being that a process can have multiple filters, and we are effectively bypassing ours.

I think this is the flow that is making the seccomp syscall:

[matchSyscall](https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L82) -> [AddRule](https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L114) (or AddRuleConditional) -> [libseccomp-golang] [f.addRuleGeneric](https://github.com/seccomp/libseccomp-golang/blob/847368b35ebf24980814276c545776d22f66d71e/seccomp.go#L869) -> [AddRuleWrapper](https://github.com/seccomp/libseccomp-golang/blob/847368b35ebf24980814276c545776d22f66d71e/seccomp_internal.go#L321) -> [C.seccomp_rule_add_...](https://github.com/seccomp/libseccomp-golang/blob/847368b35ebf24980814276c545776d22f66d71e/seccomp_internal.go#L328)

Would you be opposed to stop the process [here](https://github.com/containers/common/blob/e8021ed05dc2fb314b2e0dd26699de1d01d7cb19/pkg/seccomp/filter.go#L70) and return the filter before executing it? We would not have syscall names validated then, so we would have to extend that to cover syscall names.


> We only set the bit to the new filter. Do you think that this already modifies some state of the currently running process? 🤔

Well this is a very good question. 😆  My understanding is that once set, you can't unset it, so feels like that code is a bit redundant. Here's from the kernel documentation: 
_Once the bit is set, it is inherited across fork, clone, and execve and cannot be unset._
https://www.kernel.org/doc/html/latest/userspace-api/no_new_privs.html

saschagrunert:
@pjbgf I switched to a [modified version of containers/common](https://github.com/containers/common/pull/303/files) and it looks like the ACT_LOG still complains that it would like to run the `seccomp()` syscall (see e2e test result)

I think it will happen there:
https://github.com/containers/common/pull/303/files#diff-570d2b26b4e989d337a2ecdd2fff07afR65

:thinking:

pjbgf:
> I think it will happen there:
> https://github.com/containers/common/pull/303/files#diff-570d2b26b4e989d337a2ecdd2fff07afR65

you are right, the `seccomp_init` is [called](https://github.com/seccomp/libseccomp-golang/blob/f1ae5650e34b7a744bd25c9a834f712c164b76bc/seccomp.go#L523) within that method. I am assuming this call may also be filtered by the `seccomp` syscall filter.

Given the implications, I think we better off keeping the existing validation and then start moving towards the CRD.
What are your thoughts?

saschagrunert:
> > I think it will happen there:
> > https://github.com/containers/common/pull/303/files#diff-570d2b26b4e989d337a2ecdd2fff07afR65
> 
> you are right, the `seccomp_init` is [called](https://github.com/seccomp/libseccomp-golang/blob/f1ae5650e34b7a744bd25c9a834f712c164b76bc/seccomp.go#L523) within that method. I am assuming this call may also be filtered by the `seccomp` syscall filter.
> 
> Given the implications, I think we better off keeping the existing validation and then start moving towards the CRD.
> What are your thoughts?

I agree :+1:

PR #123, opened 2020-08-28
Update Pod example for Kubernetes 1.19

cmurphy:
Add an example to the Installation and Usage document using the
seccompProfile field of the pod securityContext, which is now available
in Kubernetes 1.19. Keep the annotation example for older Kubernetes
versions.

#### What type of PR is this?

/kind documentation

```release-note
NONE
```

### DIFF
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -55,7 +55,25 @@ path inside of the kubelet root `/var/lib/kubelet/seccomp/operator`.
 
 ### 3. Apply profile to pod
 
-Create a pod using one of the created profiles:
+Create a pod using one of the created profiles. On Kubernetes >= 1.19, the
+profile can be specified as part of the pod's security context:
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+spec:
+  securityContext:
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
Ah cool! Thank you for the contribution @cmurphy :partying_face: 

/ok-to-test

installation-usage.md
```diff
@@ -55,7 +55,25 @@ path inside of the kubelet root `/var/lib/kubelet/seccomp/operator`.
...
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: test-pod
+spec:
+  containers:
+    - name: test-container
+      image: nginx
+      securityContext:
```
hasheddan, line 70: nitpick: this would be adding it to a container, while the existing example is for the whole pod. Looks like the title of this section suggests adding to the pod. How would you feel about updating it to that?

saschagrunert:
/approve
too :)

installation-usage.md
cmurphy: done

hasheddan: Thanks!

PR #124, opened 2020-08-29
Fix error messages

saschagrunert:
#### What type of PR is this?

/kind bug

The logger.Error() API already accepts an `error` and a `message
(string)`, there is no reason to add another "reason" to it.

None
None

```release-note
- Fixed error messages in operator log to be displayed correctly, without any additional "reason" field.
```


pjbgf (approved):
/lgtm

PR #125, opened 2020-09-03
Implement SeccompProfile CRD and controller

### LINKED ISSUE #117: Custom Resource Definition (CRD) for profiles
The idea is to create a CRD on top of profiles, which helps to:

- focus on a better user experience
- provides a useful abstraction, like the minimum amount of syscalls needed for container runtimes like runc
- easier create profiles on a higher level

cmurphy:
This change implements a SeccompProfile custom resource API. A new
SeccompProfile type is created and the controller Reconciler is modified
to handle either a SeccompProfile Kind or a ConfigMap describing a
seccomp profile. The CRD manifest is generated by the new type.

#### What type of PR is this?

/kind feature
/kind api-change

Fixes #117 

```release-note
Added new Custom Resource Definition `seccompprofiles.seccomp-operator.k8s-sigs.io` as an alternative to an annotated ConfigMap for defining seccomp profiles.
```


deploy/operator.yaml
```diff
@@ -36,6 +36,16 @@ rules:
...
+- apiGroups:
+  - seccomp-operator.k8s-sigs.io
+  resources:
+  - seccompprofiles
+  verbs:
+  - create
+  - patch
+  - get
+  - watch
+  - list
```
saschagrunert, line 48: Yep, let's add this to https://github.com/kubernetes-sigs/seccomp-operator/blob/master/deploy/base/rbac.yaml and run `make deployments` (kustomize has to be installed) to sync it up.

saschagrunert (commented):
Great work so far, just two nits on the deployment.

deploy/operator.yaml
cmurphy: done

deploy/operator.yaml
```diff
@@ -378,3 +388,100 @@ metadata:
     app: seccomp-operator
   name: default-profiles
   namespace: seccomp-operator
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  name: seccompprofiles.seccomp-operator.k8s-sigs.io # ??
```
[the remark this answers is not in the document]
cmurphy: done

deploy/namespace-operator.yaml
```diff
@@ -36,6 +36,16 @@ rules:
   verbs:
   - create
   - patch
+- apiGroups:
+  - seccomp-operator.k8s-sigs.io
+  resources:
+  - seccompprofiles
+  verbs:
+  - create
```
hasheddan, line 44: Do we need create permissions on the `SeccompProfile` resource?

deploy/base/crd.yaml
```diff
@@ -0,0 +1,94 @@
+---
+apiVersion: apiextensions.k8s.io/v1
```
hasheddan, line 2: In regards to generating, I think it would be good to do this from the beginning to make sure we stay in sync. This is fairly straightforward with `controller-tools`. Here is an example of doing so with go generate --> https://github.com/crossplane/crossplane/blob/3ea5d036778e0c257960983d0cdb9271f12dc895/apis/generate.go#L26

cmd/seccomp-operator/main.go
```diff
@@ -38,8 +41,14 @@ const (
 var (
 	sync     = time.Second * 30
 	setupLog = ctrl.Log.WithName("setup")
+	scheme   = runtime.NewScheme()
 )
 
+func init() {
+	_ = clientgoscheme.AddToScheme(scheme)
```
hasheddan, line 48: Do we need to add the `clientgoscheme` here?

cmd/seccomp-operator/main.go
```diff
@@ -103,6 +112,7 @@ func run(*cli.Context) error {
 
 	ctrlOpts := ctrl.Options{
 		SyncPeriod: &sync,
+		Scheme:     scheme,
```
hasheddan, line 115: nitpick: you can leave this empty and `ctrl.NewManager()` will create a `*runtime.Scheme` for you, which you can then add types to with something like:
```
if err := seccompoperatorv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
	return errors.Wrap(err, "Cannot add core Seccomp APIs to scheme")
}
```

hasheddan (commented):
Thanks for the awesome work here @cmurphy! Excited to see this coming along! 🎉

deploy/namespace-operator.yaml
cmurphy: Nope, will remove

saschagrunert:
@cmurphy feel free to drop the `WIP` if you think this is ready for another round of review :blush:

cmurphy:
@saschagrunert I will once I get the tests fixed :smile:

api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1alpha1 contains API Schema definitions for the seccomp-operator v1alpha1 API group
+// +kubebuilder:object:generate=true
+// +groupName=seccomp-operator.k8s-sigs.io
```
cmurphy, line 19: Need input on what the group name for this resource should be. `seccomp-operator.sigs.k8s.io` would make sense but trying to use it produces this error:

```
The CustomResourceDefinition "seccompprofiles.seccomp-operator.sigs.k8s.io" is invalid: metadata.annotations[api-approved.kubernetes.io]: Required value: protected groups must have approval annotation "api-approved.kubernetes.io", see https://github.com/kubernetes/enhancements/pull/1111
```

cmurphy (commented):
This is more or less ready for feedback.

api/v1alpha1/groupversion_info.go
saschagrunert: I'm not sure if we would get approval if we request it from the API reviewers. :thinking:  Otherwise we don't have other options to use a pseudo name, right?

api/v1alpha1/seccompprofile_types.go
```diff
@@ -0,0 +1,104 @@
...
+	Names []string `json:"names"`
+	// the action for seccomp rules
+	//nolint:lll
+	// +kubebuilder:validation:Enum=SCMP_ACT_KILL;SCMP_ACT_KILL_PROCESS;SCMP_ACT_KILL_THREAD;SCMP_ACT_TRAP;SCMP_ACT_ERRNO;SCMP_ACT_TRACE;SCMP_ACT_ALLOW;SCMP_ACT_LOG
+	Action seccomp.Action `json:"action"`
+	// the errno return code to use. Some actions like SCMP_ACT_ERRNO and
+	// SCMP_ACT_TRACE allow to specify the errno code to return
+	ErrnoRet string `json:"errnoRet,omitempty"`
+	// the specific syscall in seccomp
+	Args []*Arg `json:"args,omitempty"`
```
saschagrunert, line 61: Let's add a validation here too: Linux system calls can have at most 6 arguments.

cmd/seccomp-operator/main.go
```diff
@@ -114,6 +115,10 @@ func run(*cli.Context) error {
 		return errors.Wrap(err, "create manager")
 	}
 
+	if err := seccompoperatorv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		return errors.Wrap(err, "Cannot add core Seccomp APIs to scheme")
```
saschagrunert, line 119: Let's start with lowercase errors and keep the description as short as possible to keep the noise low in the error stack:
```suggestion
		return errors.Wrap(err, "add core seccomp APIs to scheme")
```

internal/pkg/controllers/profile/profile.go
```diff
@@ -85,42 +87,66 @@ func isProfile(obj runtime.Object) bool {
...
 }
 
+type saver func(string, []byte) error
+
 // A Reconciler reconciles seccomp profiles.
 type Reconciler struct {
 	client client.Client
 	log    logr.Logger
 	record event.Recorder
+	save   saver
```
saschagrunert, line 120: Is this `saver` still required?

go.mod
```diff
@@ -12,6 +12,7 @@ require (
 	golang.org/x/sys v0.0.0-20200806125547-5acd03effb82
 	k8s.io/api v0.18.6
 	k8s.io/apimachinery v0.18.6
+	k8s.io/client-go v0.18.6
```
saschagrunert, line 15: Do we need this? It looks like it's not imported anywhere.

saschagrunert:
I think we should add e2e tests around this feature, too. But we can do it as follow-up PR. :)

internal/pkg/controllers/profile/profile.go
cmurphy: I added it to help with the unit tests. Without it, [this new test](https://github.com/kubernetes-sigs/seccomp-operator/pull/125/files#diff-e4e0872e41f6b6c837b6e05708f07bf2R146-R154) tries to write the profile to /var/lib/. I'm new to testing and mocking in Go so open to other suggestions for this.

go.mod
cmurphy: fixed

cmd/seccomp-operator/main.go
cmurphy: fixed

api/v1alpha1/seccompprofile_types.go
cmurphy: done

internal/pkg/controllers/profile/profile.go
saschagrunert: Ah okay, yeah this seems reasonable. Another option would be to use an interface, but this function seems a good leightweight alternative to me. 👍

saschagrunert (approved):
/lgtm
/approve 
/hold 
PTAL @hasheddan @pjbgf

cmurphy:
Noticed errors on the validation of flags and architectures, latest revision fixes them

cmurphy:
Added an e2e test.

saschagrunert (approved):
Thank you! PTAL @hasheddan @pjbgf

api/v1alpha1/groupversion_info.go
pjbgf: Tricky one! Would you be opposed to use something internal (without k8s-sigs.io) and open an issue for us to resolve this later on depending on whether or not the project gets renamed? I'd rather we go for a name submission/approval only once if possible.

We probably could start with `seccompprofiles.seccomp-operator.io` and then change and submit for approval when/if we finally decide on the project renaming.

pjbgf (changes requested):
Really good stuff, thank you for this contribution. 
Overall it is a lgtm from me once we change the group to a temporary one.

api/v1alpha1/groupversion_info.go
saschagrunert: Got a response in the [slack message where I raised the question](https://kubernetes.slack.com/archives/C5P3FE08M/p1600968242008600)

So I guess `x-k8s.io` is the namespace we have to choose.

Xref: https://github.com/kubernetes-sigs/cluster-api/issues/1073#issuecomment-514265501

cmurphy: Done https://github.com/kubernetes-sigs/seccomp-operator/issues/135

cmurphy: Fixed

pjbgf (approved):
/lgtm

pjbgf (approved):
/lgtm

hasheddan (approved):
/lgtm

Thanks for the awesome work here @cmurphy!!

saschagrunert:
Great work @cmurphy!  Are there any follow ups we can consider?

cmurphy:
@saschagrunert yes, I just propose https://github.com/kubernetes-sigs/seccomp-operator/pull/138 as an idea to address #131. I think we might also want another PR to address part of #117 that is not addressed here, "provides a useful abstraction, like the minimum amount of syscalls needed for container runtimes like runc". Also, documentation :smile:

PR #129, opened 2020-09-04
Improve thrown events

saschagrunert:
#### What type of PR is this?

/kind feature

We now do include the profile name in the events. Beside that, we do not
throw a "success" (normal) event if the reconciliation did not succeed
because of the profile validation.

None

None

```release-note
- Added profile name to events
```


hasheddan (approved):
/lgtm

Thanks @saschagrunert!

PR #138, opened 2020-10-01
Remove support for ConfigMap seccomp profiles

### LINKED ISSUE #131: Define behaviour for `CRD` and `ConfigMap` containing the same profile
The operator is gaining [full support](https://github.com/kubernetes-sigs/seccomp-operator/pull/125/) to represent seccomp profiles in `CRD`. 

Used in conjunction, and having the same profile name used in both CRD and ConfigMap forms, may lead to a non-deterministic results. Whatever resource is created last would overwrite the other former. Here's a few ideas on how we could tackle it:

1. Only allow a single method to represent profiles per cluster. When `CRD` is enabled, `ConfigMap` is disabled and vice-versa. 

2. Create slightly different path for each (i.e. `cm/myprofile.json` and `crd/myprofile.json`).

cmurphy:
Remove the ability to use ConfigMaps to define seccomp profiles, and instead only support the SeccompProfile custom resource.

#### What type of PR is this?

/kind feature
/kind api-change

```release-note
Seccomp profiles can now no longer be configured using the ConfigMap native resource, and instead may now only be defined using the provided SeccompProfile custom resource.
```


saschagrunert:
> Notes for discussion:
> 
> * Having the controller application be configurable does not prevent a user from uploading a ConfigMap or a SeccompProfile, so
>   I tried to provide some transparency by having the controller monitor both resources and emits events when it discovers a resource that it is not configured to respect. But that is tending to make the code extra spaghetti-like. Would it be better to just silently ignore the resources it is not configured for? Am I overthinking it?

I like the idea of getting feedback by a warning if the configure is used in a different way than configured. Not sure if a refactoring of the logic makes sense at this point, but we could split-up the logic into two controllers, select the right one at startup and verify the behavior with a shared code base package.

> * Since it's now strictly one-or-the-other, there's not a smooth migration path from one form to the other, so changing the default value of the new config option will be disruptive.

Would like to hear @pjbgf thoughts on that. From my perspective this "should be fine" since the feature has not released yet, right?

> * Before this can be merged we need to rewrite the two (so far) default profiles as `SeccompProfile`s.

:+1: 

/ok-to-test

cmurphy:
> but we could split-up the logic into two controllers, select the right one at startup and verify the behavior with a shared code base package.

I think the controllers both need to be running in order to detect both kinds of resources and emit a warning when it finds the wrong kind. Not bothering with a warning would allow us to simplify and only run one after startup.

cmurphy:
Added the two default profiles, but found I needed to add them as a separate manifest because there's some small delay between when the CRD is applied and when it can actually be POSTed to, so it unfortunately becomes a two-step process. I'm not sure if there's a better way to handle this.

`SeccompProfiles` right now are written under `<namespace>/custom-profiles` which doesn't feel right for "default" profiles, any thoughts on a better way to handle the directory structure? Should there be a `directory` field in the `spec` object, or some kind of annotation?

How do people feel about emitting warnings versus silently ignoring configmaps/crds? If we want to continue the approach of emitting a warning then it might be good to split out the `ConfigMap`s for the default profiles also in their own manifest.

saschagrunert:
> Added the two default profiles, but found I needed to add them as a separate manifest because there's some small delay between when the CRD is applied and when it can actually be POSTed to, so it unfortunately becomes a two-step process. I'm not sure if there's a better way to handle this.

Hm, yeah those kind of races are tricky. Can we wait somehow on the existence/creation of the default profiles and then start the reconciler? I think solving it by code would be a good option (maybe a good follow up).

> `SeccompProfiles` right now are written under `<namespace>/custom-profiles` which doesn't feel right for "default" profiles, any thoughts on a better way to handle the directory structure? Should there be a `directory` field in the `spec` object, or some kind of annotation?

We had the default profiles in the `seccomp-operator` namespace which is kinda our system namespace for the operator. I'm wondering where the `custom-profiles` directory comes from.
 
> How do people feel about emitting warnings versus silently ignoring configmaps/crds? If we want to continue the approach of emitting a warning then it might be good to split out the `ConfigMap`s for the default profiles also in their own manifest.

Yes I think we could solve it by a kustomize overlay. A drawback is that we have multiple deployment variants of the operator but I'm fine with that if we document it and have the shiny new CRD the default. :+1: 

cc @hasheddan @pjbgf

cmurphy:
> We had the default profiles in the seccomp-operator namespace which is kinda our system namespace for the operator. I'm wondering where the custom-profiles directory comes from.

It comes from [the controller](https://github.com/kubernetes-sigs/seccomp-operator/blob/ae1dbf2c98b5d8a282208993cb6c45b14aba6c14/internal/pkg/config/config.go#L37) - ConfigMap uses the name of the ConfigMap resource as the directory name, but there's no natural equivalent for SeccompProfile since the name of the resource becomes the name of the `.json` file, so it's (currently) lumped together in one directory called `custom-profiles`.

cmurphy:
> Hm, yeah those kind of races are tricky. Can we wait somehow on the existence/creation of the default profiles and then start the reconciler? I think solving it by code would be a good option (maybe a good follow up).

It doesn't have to do with the reconciler, it's the API server itself that is delayed in recognizing the new CRD and so can't accept the resource being POSTed to it:

```
$ kubectl apply -f deploy/operator.yaml 
namespace/seccomp-operator created
serviceaccount/seccomp-operator created
clusterrole.rbac.authorization.k8s.io/config-map-reader created
clusterrolebinding.rbac.authorization.k8s.io/config-map-reader-binding created
daemonset.apps/seccomp-operator created
customresourcedefinition.apiextensions.k8s.io/seccompprofiles.seccomp-operator.x-k8s.io created
configmap/seccomp-operator-profile created
configmap/default-profiles created
Error from server (NotFound): error when creating "deploy/operator.yaml": the server could not find the requested resource (post seccompprofiles.seccomp-operator.x-k8s.io)
Error from server (NotFound): error when creating "deploy/operator.yaml": the server could not find the requested resource (post seccompprofiles.seccomp-operator.x-k8s.io)
```

internal/pkg/controllers/profile/profile.go
```diff
@@ -87,6 +97,15 @@ func isProfile(obj runtime.Object) bool {
...
 
+	switch os.Getenv(config.ProfileControllerEnvKey) {
+	case "", "ConfigMap": // defaults to ConfigMap if no value provided
+		activeController = configMapKind
+	case "SeccompProfile":
+		activeController = seccompProfileKind
+	default:
+		return errors.New("invalid profile controller configured")
+	}
+
```
pjbgf, line 108: Following up the [slack discussion](https://kubernetes.slack.com/archives/C013FQNB0A2/p1602176399038600?thread_ts=1602163933.038000&cid=C013FQNB0A2) we can use this PR to deprecate the `ConfigMap` support.

pjbgf (commented):
@cmurphy thanks again for your contribution. Do you mind amending this PR  to also remove `ConfigMap` support?

pjbgf:
> `SeccompProfiles` right now are written under `<namespace>/custom-profiles` which doesn't feel right for "default" profiles, any thoughts on a better way to handle the directory structure? Should there be a `directory` field in the `spec` object, or some kind of annotation?

By removing `ConfigMap` support we can also remove the `custom-profiles`, which would then lead to have the same directory structure as we had before - nice and clean.

cmurphy:
[**@pjbgf**](https://github.com/pjbgf) commented on [Oct 8, 2020, 12:32 PM PDT](https://github.com/kubernetes-sigs/security-profiles-operator/pull/138#issuecomment-705778265 "2020-10-08T19:32:09Z - Replied by Github Reply Comments"):
> > `SeccompProfiles` right now are written under `<namespace>/custom-profiles` which doesn't feel right for "default" profiles, any thoughts on a better way to handle the directory structure? Should there be a `directory` field in the `spec` object, or some kind of annotation?
> 
> By removing `ConfigMap` support we can also remove the `custom-profiles`, which would then lead to have the same directory structure as we had before - nice and clean.

It's not quite the same directory structure. The directory structure with ConfigMap is `<namespace>/<configmap name>/<configmap key name>`. With SeccompProfile the structure is `<namespace>/custom-profiles/<seccompprofile name>.json`. So removing `custom-profiles` moves the `.json` file one level up.

deploy/defaultprofiles.yaml
```diff
@@ -0,0 +1,176 @@
...
+    - accept4
+    - access
+    - arch_prctl
+    - bind
+    - brk
+    - capget
+    - capset
+    - chdir
+    - chown
+    - clone
```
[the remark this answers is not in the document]
pjbgf: ignore this as it actually does not make sense. 🤦

pjbgf:
> It's not quite the same directory structure. The directory structure with ConfigMap is `<namespace>/<configmap name>/<configmap key name>`. With SeccompProfile the structure is `<namespace>/custom-profiles/<seccompprofile name>.json`. So removing `custom-profiles` moves the `.json` file one level up.

You are absolutely right, I completely disregarded the fact that each CRD will only store a single profile. 
Considering that, and connecting this to the "auto-apply" of profiles, I think we could have a way to define what workload we are talking about and use that within the directory structure. 

For example, if we had a new field called (target-workload? target-image?) which defined what image this profile is for, we could then build the entire path: `<namespace>/<workload>/<seccompprofile name>.json`

kube-proxy:
`kube-system/k8s.gcr.io/kube-proxy/kube-proxy-v1.18.json`

nginx:
`default/nginx/nginx-1.19.1.json`

what are your thoughts @cmurphy  @saschagrunert @hasheddan ?

cmurphy:
Sounds fine to me :+1:

cmurphy:
So you're envisioning that there could be multiple profiles per workload, e.g. `default/nginx/nginx-1.19.1.json` and `default/nginx/nginx-1.18.0.json`?

pjbgf:
> So you're envisioning that there could be multiple profiles per workload, e.g. `default/nginx/nginx-1.19.1.json` and `default/nginx/nginx-1.18.0.json`?

Not always, but it is highly likely that workloads in active development would experience a change in their seccomp profile as new features start (or stop) depending on other system calls. Therefore, restricting profile to workload as `1:N` a mapping may break workloads.

Conversely, mapping them as `1:1` would lead to too many profiles. The auto apply feature will flash this out a bit further, but as a first thought I was thinking that the map could be based on `target-workload-image` + `target-tag regex`, so it is flexible enough to cover both `1:N` and `1:1` mappings.

saschagrunert:
The general direction sounds good to me. We should block the case where people can overwrite the profiles form other namespaces than they're permitted to act on.

cmurphy:
Okay the latest set of revisions fully remove `ConfigMap` support, including adding a `TargetWorkload` attribute to the CRD
to help with grouping profiles and create a consistent directory structure as discussed.

Fully pivoting from ConfigMap to SeccompProfile is a little tricky, because 1) there is a very slight delay between registering a new CRD with the kubernetes API and it being available to POST a new resource to, so if both the CRD and the default SeccompProfile resources are included in the deployment manifest, creating the default profiles might fail and usually failed for me, and 2) there is a chicken/egg problem with creating the seccomp-operator container which reconciles the seccomp-operator profile, because the controller can't start without the profile already there. To fix both of these, I added another init container which reads a special ConfigMap which contains the default SeccompProfile resources, serializes the seccomp-operator profile to disk, and creates the default SeccompProfile resources which the main container will later reconcile. It seems like init containers do not need to adhere to the pod profile, at least in the annotation - not sure if that's intentional but it was advantageous here. The new init container is based on an external dockerhub image and is just a proof of concept, but if this approach is acceptable I can transform it into its own app so that it doesn't rely on an external image.

I will also update the usage guide, and will need to double check whether this works properly with the namespaced deployment. This will also need to be rebased on top of #139.

deploy/base/operator.yaml
```diff
@@ -18,6 +18,25 @@ spec:
...
+
+            while ! kubectl get crd | grep seccompprofile ; do sleep 1 ; done
+
+            kubectl apply -f /opt/bootstrap-profiles/
+            kubectl --namespace seccomp-operator get seccompprofile seccomp-operator -o jsonpath='{.spec}' > /opt/seccomp-profiles/seccomp-operator.json
+          volumeMounts:
+          - name: profile-configmap-volume
+            mountPath: /opt/bootstrap-profiles
+          - name: seccomp-operator-profile-volume
+            mountPath: /opt/seccomp-profiles
```
saschagrunert, line 39: I think we can do it this way. 

An alternative idea would be:

- Give the operator permissions to write CRDs (we need it anyway later on with the recording feature)
- Deploy the profile on operator startup, containing the same logic for checking if the CRD is deployed (via client-go)
- Let the operator reconcile…

I know that I've mentioned it somewhere in the past, do you think this would work?

internal/pkg/controllers/profile/profile.go
```diff
@@ -120,32 +89,18 @@ type Reconciler struct {
...
-			return reconcile.Result{}, errors.Wrap(ignoreNotFound(err), errGetProfile)
-		}
-	}
-
 	// Pre-check if the node supports seccomp
 	if !seccomp.IsSupported() {
 		err := errors.New("profile not added")
 		logger.Error(err, fmt.Sprintf("node %q does not support seccomp", os.Getenv(config.NodeNameEnvKey)))
+		seccompProfile := &seccompoperatorv1alpha1.SeccompProfile{}
 		r.record.Event(seccompProfile,
```
saschagrunert, line 104: I see no need to bind the variable:
```suggestion
		r.record.Event(&seccompoperatorv1alpha1.SeccompProfile{},
```

internal/pkg/controllers/profile/profile.go
```diff
@@ -251,7 +166,7 @@ func saveProfileOnDisk(fileName string, contents []byte) error {
 
 // GetProfilePath returns the full path for the provided profile name and config.
 func GetProfilePath(profileName, namespace, subdir string) (string, error) {
-	if filepath.Ext(profileName) == "" {
+	if filepath.Ext(profileName) != ".json" {
 		profileName += ".json"
```
saschagrunert, line 170: nit: Let's make `.json` a `extJSON` (or similar) `const`

saschagrunert (commented):
I really like how this simplifies the reconciliation. :green_heart: Just have some nits, and one alternative suggestion to the init container.

deploy/base/operator.yaml
cmurphy: But the operator can't even start if it requires the seccomp-operator profile which doesn't exist yet, that's the chicken-and-egg problem I was describing

https://github.com/kubernetes-sigs/security-profiles-operator/blob/fb21e8a423c7fbc824754b73ed625d65e382e7f9/deploy/operator.yaml#L79

saschagrunert: Yes, makes sense. Can we merge both init containers into one?

cmurphy: Hmm yes I think they could be merged. But it would mean using the same `securityContext` for both parts, where currently this init container doesn't need any `securityContext`, is that okay?

saschagrunert: Yes we can extend the existing security context to match the new requirements.

cmurphy:
Fixed the namespaced deployment, reimplemented the init containers in Go, and fixed nits. Still needs a rebase/rename and docs.

saschagrunert:
> Fixed the namespaced deployment, reimplemented the init containers in Go, and fixed nits. Still needs a rebase/rename and docs.

Thanks! From my side this looks good so far. Sorry for the rebase, I hope this does not introduce to much trouble on your side.

cmurphy:
This is pretty much ready now. The verify test was failing on the generated namespace-operator.yaml manifest though I couldn't reproduce locally. I can clean up the fixup commits after you guys have had a chance to take a look.

cmd/security-profiles-operator-init/main.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	// Set up directories for non-root access
+	kubeletSeccompRoot := os.Getenv(kubeletSeccompRootEnvKey)
+	operatorRoot := os.Getenv(operatorRootEnvKey)
+	operatorLink := os.Getenv(operatorLinkEnvKey)
+	f, err := os.Lstat(kubeletSeccompRoot)
+	if os.IsNotExist(err) || !f.IsDir() {
+		if err = os.Mkdir(kubeletSeccompRoot, 0744); err != nil {
```
saschagrunert, line 158: More a style than a functional nit:
```suggestion
		if err = os.Mkdir(kubeletSeccompRoot, os.FileMode(0o744)); err != nil {
```

saschagrunert (approved):
/lgtm

/hold
PTAL @hasheddan @pjbgf 

I think we have to change the image promotion to include the init image as well, I can take care of that afterwards.

cmd/security-profiles-operator-init/main.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	profiles, err := getProfiles(c.String("path"))
+	if err != nil {
+		return errors.Wrap(err, "getting profiles")
+	}
+
+	// Create SeccompProfile resources
+	ctx := context.Background()
+	for name, sp := range profiles {
```
hasheddan, line 134: Is the main reason for maintaining our own init container such that all of the "bootstrap" profiles will be reflected as CRDs in the cluster as well? I think we can modify deployment such that instances of the CRD are not created until the CRD itself is. For instance, helm handles this with a dedicated `crds` directory. Not strictly against maintaining our own init image, but as has been mentioned before, it is a larger attack surface.

cmurphy: The main reason is item number 2 in [this commit message](https://github.com/kubernetes-sigs/security-profiles-operator/pull/138/commits/92481d97f21d7c978bb6f329ed15eab09b7309ea) which is that the controller container itself cannot start if the seccomp profile for the container image itself doesn't exist on the host, so the controller can't reconcile the `SeccompProfile` because it hasn't started yet. I'm certainly open to other ideas here.

cmd/security-profiles-operator-init/main.go
cmurphy: done

cmd/security-profiles-operator-init/main.go
pjbgf: @cmurphy We currently resolve that chicken-egg situation by not relying on any of the operator's functionality to deploy its profile, we simply use the built-in "ConfigMap to Volume" mount and copy the profile with `cp`, that's why that specific `ConfigMap` does not have the `seccomp.security.kubernetes.io/profile: "true"`. Here's where we do it:

https://github.com/kubernetes-sigs/security-profiles-operator/blob/49f5464dc2e05fbfaa9d176ca0afb156034824a8/deploy/base/operator.yaml#L43

As @hasheddan mentioned, we took that approach to limit the attack surface running without any profile-based enforcement. Do you think we can get away without changing that?

pjbgf (commented):
@cmurphy thank you again for the great effort on this one. For me we just need to iron out the "init changes" that I commented on above and the nit from @saschagrunert.

cmurphy:
@pjbgf thanks for the hint, I didn't notice that that ConfigMap was special and not annotated. Keeping that will be much easier.

> and the nit from @saschagrunert 

I think I already addressed the filemode nit but let me know if I missed something else.

cmurphy:
Updated to remove the unnecessary init container. Added the following to the main commit message, will copy here for visibility:

```
    This requires modifications to the default profiles. The
    security-profiles-operator profile remains a ConfigMap because the
    controller's init container needs to mount and copy it in order to
    bootstrap the controller, but the special annotation is no longer
    needed. This has the consequence that this particular profile cannot be
    managed as a SeccompProfile resource, it can only be managed as a
    ConfigMap. The nginx-1.19.1 profile is transformed to a SeccompProfile
    resource. Since there is a slight delay between a CRD being registered
    with the API server and it being available to POST to, the CRD
    definition is split into its own manifest so it can be deployed first
    before the default profiles.
```

saschagrunert (approved):
/lgtm

saschagrunert:
Ah there seems an issue in the e2e test:
```
I1027 02:08:39.689955    6163 suite_test.go:164]  "msg"="Verifying security-profiles-operator-profile profile on node so-e2e-1603764250-worker"  
    tc_default_profiles_test.go:82: 
        	Error Trace:	tc_default_profiles_test.go:82
        	            				tc_default_profiles_test.go:53
        	            				e2e_test.go:62
        	Error:      	Expected nil, but got: &errors.errorString{s:"invalid syntax"}
        	Test:       	TestSuite/TestSeccompOperator
```

cmurphy:
[**@saschagrunert**](https://github.com/saschagrunert) commented on [Oct 27, 2020, 1:05 AM PDT](https://github.com/kubernetes-sigs/security-profiles-operator/pull/138#issuecomment-717063154 "2020-10-27T08:05:59Z - Replied by Github Reply Comments"):
> Ah there seems an issue in the e2e test:

Oops, pushed a little too soon. Should be fixed now.

pjbgf (approved):
@cmurphy really good stuff, thank you again for the effort on this one. 

/lgtm

pjbgf:
Based on the thread, I don't think there is anything else holding this to be merged.

/unhold

PR #140, opened 2020-10-07
WIP: PoC for Recording seccomp profiles

### LINKED ISSUE #46: Record seccomp profiles
The idea is to allow the recording of seccomp profiles directly inside the Kubernetes cluster. To achieve that we could utilize the [oci-seccomp-bpf-hook](https://github.com/containers/oci-seccomp-bpf-hook), which has the following requirements:

- Kernel headers have to be installed (could be handled by the operator or the sysadmin)
- An OCI hook compatible container runtime

We could think about annotating a node, then install the kernel headers (may be a prerequisite) and the hook. After that, the operator would have to annotate the target workload correctly (that's the way how the hook knows what to do). The recorded profile would have to be written into a configmap afterwards, which would cause that all nodes sync the profile to disk.

WDYT? It's overall a bit complicated but seems like a valuable feature after the initial release.

---

Edit: We could also use the seccomp notifier feature to think about a more elegant way to record profiles. This needs more evaluation but might be achievable by creating a PID namespace-shared sidecar.

rhafer:
#### What type of PR is this?

/kind feature

This PoC adds a controller to the controller that will collect seccomp profiles generated by the `seccomp-bpf` oci-hook. It works by watching for pods carrying the "io.containers.trace-syscall" annotation which is used by the above oci hook to enable tracing of syscalls issued by a pod and generating a seccomp profiles from that.
After pod completion the controller will collect the generated profile and create a "SeccompProfile" resource in the namespace that the traced pod was running in.

This is still in it's early stages and there are still quite a few things to work on:

- Better error handling
- Documentation
- Tests

And a ton of open questions to be detailed out:

- What to do with the generated profiles? Currently the collected profiles will be just created as a "SeccompProfile" resource and will unconditionally be installed on every node. For that point the profile is generally usable by everybody who can access it. Should we maybe add some flag (label?, annotation?) to the SeccompProfile CRD that will mark a profile as being disabled until e.g. a manual review?
- In the non-namespaced deployment the operatore now needs pretty wide (read) access to resources (namely pods and nodes). It also needs to be able to create "SeccompProfiles" in any namespace. Can this be further restricted?
- Deeper integration with the oci-hook? Currently the integration with the tracing mechanism is pretty rough and based on specifying the right filename/directory in the `.trace-syscall` annotation. This is not only a bad UX but also raises question about how to cleanup the directory and what to when e.g. to pods specify the same output file. 
  - Instead of watching for the hook annotation and giving the user control over the output file I guess we could create a mutating webhook that triggers on a different annotation (something specific to the seccomp operator)  and injects the oci-hook annotation into the pod, which would give us better control over the output.
  - The oci-hook could be modified to "submit" the profile directly to the operator (e.g. via grpc) instead of writing a file to disk
- Trying out different approach for the syscall tracing. (See also https://github.com/kubernetes-sigs/security-profiles-operator/issues/46#issuecomment-702288267) and https://kubernetes.slack.com/archives/C013FQNB0A2/p1600971007007100)

Fixes #46

### DIFF
--- a/internal/pkg/controllers/profilerecorder/profilerecorder.go
+++ b/internal/pkg/controllers/profilerecorder/profilerecorder.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package profilerecorder
 
 import (
--- a/internal/pkg/controllers/profilerecorder/profilerecorder.go
+++ b/internal/pkg/controllers/profilerecorder/profilerecorder.go
@@ -1,0 +1,252 @@
+package profilerecorder
+
+import (
+	"context"
+	"encoding/json"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/security-profiles-operator/api/v1alpha1"
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
+)
+
+const (
+	// default reconcile timeout.
+	reconcileTimeout = 1 * time.Minute
+
+	maxTraceAnnotations = 2
+
+	errGetClient         = "cannot get client connection"
+	errGetNode           = "cannot get node object"
+	errGetPod            = "cannot get pod"
+	errInvalidAnnotation = "invalid Annotation"
+
+	// traceAnnotation is the annotation on a Pod that triggers the
+	// oci-seccomp-bpf-hook to trace the syscalls of a Pod and created a
+	// seccomp profile.
+	traceAnnotation = "io.containers.trace-syscall"
+
+	reasonProfileRecording      event.Reason = "SeccompProfileRecording"
+	reasonProfileCreated        event.Reason = "SeccompProfileCreated"
+	reasonProfileCreationFailed event.Reason = "CannotCreateSeccompProfile"
+)
+
+type RecorderReconciler struct {
+	client      client.Client
+	log         logr.Logger
+	record      event.Recorder
+	nodeAddress string
+	podsToWatch map[string]string
+}
+
+func (r *RecorderReconciler) isPodOnLocalNode(obj runtime.Object) bool {
+	p, ok := obj.(*corev1.Pod)
+
+	if !ok {
+		return false
+	}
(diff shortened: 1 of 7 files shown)


pjbgf:
@rhafer really good job so far! Here's my 2 pence on the open questions: 

> What to do with the generated profiles? Currently the collected profiles will be just created as a "SeccompProfile" resource and will unconditionally be installed on every node. For that point the profile is generally usable by everybody who can access it. Should we maybe add some flag (label?, annotation?) to the SeccompProfile CRD that will mark a profile as being disabled until e.g. a manual review?

Very good question. Maybe it would be worth keeping it simple for now and just create the CRD profile, which will lead to sync/saving into all nodes. At this point, the user would always need to actively map a profile to a workload, which they should be allowed to. We can refine this further once we have the auto apply in place. 

On that point, it would be nice to auto populate the new field that defines what image is this profile for. This is still WIP, just linking here for future reference: https://github.com/kubernetes-sigs/security-profiles-operator/pull/138#issuecomment-705837622

> In the non-namespaced deployment the operatore now needs pretty wide (read) access to resources (namely pods and nodes). It also needs to be able to create "SeccompProfiles" in any namespace. Can this be further restricted?

My understanding is that now that we are deprecating the `ConfigMap` support, we won't have cluster level profiles anymore. So we probably want to have a default target `namespace` for the profiles to be generated into so we can have a tight RBAC by default. @cmurphy please correct me if my namespace comment is gibberish. 😄

rhafer:
> On that point, it would be nice to auto populate the new field that defines what image is this profile for. This is still WIP, just linking here for future reference: #138 (comment)

:+1:

> My understanding is that now that we are deprecating the `ConfigMap` support, we won't have cluster level profiles anymore. So we probably want to have a default target `namespace` for the profiles to be generated into so we can have a tight RBAC by default.

Hm, not sure I completely understand that. But I think the recorded profile should endup in the same namespace as the pod that it was recorded from. Otherwise it would just be too easy to have two pods from different namespaces overwrite each other's recorded profiles.

cmurphy:
>> In the non-namespaced deployment the operatore now needs pretty wide (read) access to resources (namely pods and nodes). It also needs to be able to create "SeccompProfiles" in any namespace. Can this be further restricted?

> My understanding is that now that we are deprecating the ConfigMap support, we won't have cluster level profiles anymore.

I didn't fully understand this before, but I guess the "cluster level profiles" are, e.g. the `seccomp-operator-profile` configmap which gets mounted and copied by the initcontainer to `/var/lib/kubelet/seccomp/seccomp-operator.json`? Right now `SeccompProfile` doesn't do that, but maybe it should?

> So we probably want to have a default target namespace for the profiles to be generated into so we can have a tight RBAC by default.

I agree with @rhafer that the profile should go in the namespace it was recorded from, and as such I'm not sure how the operator could be further restricted: it needs broad permissions to write SeccompProfiles to any namespace.

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,252 @@
...
+
+	errGetClient         = "cannot get client connection"
+	errGetNode           = "cannot get node object"
+	errGetPod            = "cannot get pod"
+	errInvalidAnnotation = "invalid Annotation"
+
+	// traceAnnotation is the annotation on a Pod that triggers the
+	// oci-seccomp-bpf-hook to trace the syscalls of a Pod and created a
+	// seccomp profile.
+	traceAnnotation = "io.containers.trace-syscall"
```
saschagrunert, line 45: Did you use the CRI-O annotation patch to make this work? I'm thinking if we can choose another way to pass the data than the annotation…

rhafer: Yeah. For now this depends on the annotation patch. (For reference https://github.com/cri-o/cri-o/pull/4138)

pjbgf:
> Hm, not sure I completely understand that. But I think the recorded profile should endup in the same namespace as the pod that it was recorded from. Otherwise it would just be too easy to have two pods from different namespaces overwrite each other's recorded profiles.

@rhafer  I am not opposed to get the recorded profile into the namespace the recorded pod was running from, I just thought this probably would happen as two separate steps to start with:

1) A pod is selected for profile recording, a temporary seccomp profile may be created and will be amended through this process.
2) A user promotes the generated profile by pushing it to a given namespace and applies it to a workload.

Given how difficult it is to get a complete profile, I was thinking on restricting the first part to a given namespace, which we could then enforce simply by RBAC. Then second would be the user's responsibility until this feature matured.

However, I think it is more important to have an end-to-end process we can play around with then get bog down on this details, so feel free to go ahead and we will then tackle such concerns as they appear.

rhafer:
rebased

saschagrunert:
> rebased

Do you think that we can demo this at the community meeting this Thursday?

rhafer:
> Do you think that we can demo this at the community meeting this Thursday?

Yeah, that should work. Even if just for discussing some of the road blocks I've hit recently with the current approach :-)

saschagrunert:
@rhafer I reused most of your work in #247, thank you again for the contribution. I assume that you will not continue this PR now. Feel free to reopen if that is not the case. :pray: 

/close

PR #152, opened 2020-11-03
Add base profile example and mergeable profiles

### LINKED ISSUE #150: Decide whether and how to add default syscalls to the SeccompProfile CRD
Part of the idea of #117 was to perhaps ensure that a standard set of default syscalls were present in a profile so that a minimum viable pod could be launched in kubernetes on e.g. runc. This issue is to evaluate whether this is worthwhile and discuss how to implement it.

The first consideration is that it's possible that the syscalls used by runc could change over time, and in fact they haven't been fully documented (https://github.com/opencontainers/runc/issues/2097) so we would need to maintain our own list. Moreover, a container runtime other than runc might be used, and container managers on top of runc like docker/cri-o might also call other syscalls. Also, the list might be different depending on the architecture.

Another consideration is the pause container, which seems to use an additional set of syscalls on top of what is required for runc. Should those syscalls be included in the defaults?

Finally, the actions to take for a syscall aren't a blanket "allow" or "deny", there is also e.g. "log", "trace", "kill", etc. If a user creates a complex profile with different actions for different syscalls, should the default syscalls use the "allow" action or the "log" action? What happens if the user adds one of the default syscalls to their "log" rule, but the controller is automatically adding it to the "allow" rule?

One idea could be to simply maintain an example SeccompProfile that contains the minimum required syscalls instead of building it into the CRD/controller. Thoughts?

cmurphy:
#### What type of PR is this?

/kind feature

This PR has two commits, one to add a base `SeccompProfile` resource containing the minimum required syscalls to launch a container on runc in kubernetes, and another to add a `BaseProfileName` attribute to the `SeccompProfile` CRD to allow users to construct profiles whose syscalls take the union of two profiles.

Fixes #150 

```release-note
Added a new example `SeccompProfile` to provide a starting point on which to build custom profiles, and an attribute `BaseProfileName` to the `SeccompProfile` kind to allow merging syscalls from two profiles.
```

### DIFF
--- a/examples/baseprofile.yaml
+++ b/examples/baseprofile.yaml
@@ -1,0 +1,38 @@
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: SeccompProfile
+metadata:
+  name: runc-v1.0.0-rc92
+spec:
+  defaultAction: SCMP_ACT_ERRNO
+  targetWorkload: base
+  architectures:
+  - SCMP_ARCH_X86_64
+  syscalls:
+  - action: SCMP_ACT_ALLOW
+    names:
+    - capget
+    - capset
+    - chdir
+    - close
+    - epoll_ctl
+    - epoll_pwait
+    - execve
+    - fchown
+    - fcntl
+    - fstat
+    - fstatfs
+    - futex
+    - getdents64
+    - getpid
+    - getppid
+    - nanosleep
+    - newfstatat
+    - openat
+    - prctl
+    - read
+    - sched_yield
+    - setgid
+    - setgroups
+    - setuid
+    - write
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -81,6 +81,13 @@ spec:
       image: nginx
 ```
 
+#### Base syscalls for a container runtime
+
+An example of the minimum required syscalls for a runtime such as runc (tested
+on version 1.0.0-rc92) to launch a container can be found in [the
+examples](./examples/baseprofile.yaml). Use this example as a starting point for
+creating custom profiles for your application.
+
 ## Restricting to a Single Namespace
 
 The security-profiles-operator can optionally be run to watch SeccompProfiles in
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -85,8 +85,27 @@ spec:
 
 An example of the minimum required syscalls for a runtime such as runc (tested
 on version 1.0.0-rc92) to launch a container can be found in [the
-examples](./examples/baseprofile.yaml). Use this example as a starting point for
-creating custom profiles for your application.
+examples](./examples/baseprofile.yaml). You can use this example as a starting
+point for creating custom profiles for your application. You can also
+programmatically combine it with your custom profiles in order to build
+application-specific profiles that only specify syscalls that are required on
+top of the base calls needed for the container runtime. For example:
+
+```yaml
+apiVersion: v1
+kind: SeccompProfile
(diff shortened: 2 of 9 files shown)


examples/baseprofile.yaml
```diff
@@ -0,0 +1,38 @@
+---
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: SeccompProfile
+metadata:
+  name: base
```
saschagrunert, line 5: Do you think we need some kind of description field for a profile, too? This way we could still name it `base`, but explain for which runtime (runc) and version it has been built and for which purpose.

installation-usage.md
```diff
@@ -81,6 +81,32 @@ spec:
       image: nginx
 ```
 
+#### Base syscalls for a container runtime
+
+An example of the minimum required syscalls for a runtime such as runc to launch
```
saschagrunert, line 86: Let's include the runc version:
```suggestion
An example of the minimum required syscalls for a runtime such as runc v1.0.0-rc92 to launch
```

internal/pkg/controllers/profile/profile.go
```diff
@@ -120,14 +121,76 @@ func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error)
...
+		for n := range names {
+			syscall.Names = append(syscall.Names, n)
+		}
+		finalSyscalls = append(finalSyscalls, &syscall)
+	}
+	return finalSyscalls
+}
+
+func (r *Reconciler) mergeBaseProfile(
+	ctx context.Context, sp *v1alpha1.SeccompProfile, l logr.Logger) (OutputProfile, error) {
```
saschagrunert, line 160: Just a style nit:
```suggestion
	ctx context.Context, sp *v1alpha1.SeccompProfile, l logr.Logger,
) (OutputProfile, error) {
```

saschagrunert (approved):
Just some nits, otherwise LGTM :)

examples/baseprofile.yaml
cmurphy: Maybe we could just include it in the name, combined with `targetWorkload`? e.g. `name: runc-v1.0.0-rc92` `targetWorkload: base`

installation-usage.md
cmurphy: I didn't happen to test it on `v1.0.0-rc92`, have you? I've only tried `v1.0.0-rc10`

examples/baseprofile.yaml
saschagrunert: Yes this would also work from my perspective :+1:

installation-usage.md
saschagrunert: No, I just assumed the latest version. I would not expect changes in the setup/syscalls, but who knows…

examples/baseprofile.yaml
cmurphy: done

installation-usage.md
cmurphy: tested on rc92, added the version here and to the yaml

internal/pkg/controllers/profile/profile.go
cmurphy: done

cmurphy (commented):
Tested on runc v1.0.0-rc92, also fixed the e2e test to actually use a security context :facepalm:

saschagrunert (approved):
/lgtm
Thank you!

PR #155, opened 2020-11-06
Add seccomp profile deletion

### LINKED ISSUE #106: Clean up profiles when ConfigMap deleted

Currently profiles are not cleaned up on a node when their `ConfigMap` is deleted. The lifecycle of the profile on a node should match that of the `ConfigMap` that created it.

#### Why is this needed:

In most cases it is actually fine for the profile to be orphaned. In fact, cleaning up a profile after a workload has started using it can be a problem if that workload restarts and the profile is no longer there. However, there are a few cases in which orphaning profiles can be a problem:
- A user creates a profile that is overly permissive or has a typo and does not want it to be able to be applied to containers anymore.
- A user creates a new profile with a similar name to an old one that no longer exists as a `ConfigMap`, but still exists on the node. They then create a workload and accidentally apply the old profile. The operator would not block this because it would be using a valid profile, but it would be unexpected and potentially difficult to troubleshoot.
- A user cannot get an accurate global list of seccomp profiles without interacting with each node.

cmurphy:
#### What type of PR is this?

/kind feature
/kind api-change

Three commits here: the first is a bugfix to correct the way conflicting updates are handled when the resource status is updated, the second is a trivial fix for the e2e tests to make them more reliable, and the main commit adds finalizers to the resource in order to ensure the profile is deleted from every node. This also adds new fields to the Status subresource of the CRD to help facilitate deleting profiles.

Fixes #106 

```release-note
Added the ability to delete seccomp profiles from nodes by deleting `SeccompProfile` resources. Added new fields `activeWorkloads` and `status` to the `status` subresource of the `SeccompProfile` kind.
```


internal/pkg/controllers/profile/profile.go
```diff
@@ -204,15 +208,45 @@ func (r *Reconciler) reconcileSeccompProfile(
 		r.record.Event(sp, event.Warning(reasonCannotGetProfilePath, err))
 		return reconcile.Result{RequeueAfter: wait}, nil
 	}
+	nodeFinalizerString := os.Getenv(config.NodeNameEnvKey) + "-delete"
+	if !sp.ObjectMeta.DeletionTimestamp.IsZero() { // object is being deleted
```
saschagrunert, line 212: Do we still want to remove the profile if a workload references it? Restarting the workload would not work any more.

internal/pkg/controllers/profile/profile.go
```diff
@@ -223,6 +257,79 @@ func (r *Reconciler) reconcileSeccompProfile(
...
+	}
+	return errors.Wrap(err, fmt.Sprintf("exceeded %d retries removing finalizer from seccompprofile", retries))
+}
+
+func handleDeletion(sp *v1alpha1.SeccompProfile, l logr.Logger) error {
+	profilePath, err := GetProfilePath(sp.ObjectMeta.Name, sp.ObjectMeta.Namespace, sp.Spec.TargetWorkload)
+	if err != nil {
+		return err
+	}
+	if _, e := os.Stat(profilePath); !os.IsNotExist(e) {
```
saschagrunert, line 324: I'm wondering why we `Stat` here before the `Remove`. I think remove does return `os.IsNotExist(e)` too in case that the file does not exist.

saschagrunert (commented):
Really cool! I think we should discuss the "deletion if still referenced" question below :point_down:

internal/pkg/controllers/profile/profile.go
cmurphy: What do you think the behavior should be? Should the `SeccompProfile` add an error condition but remain in an active state? Should it go into a terminating state until all workloads are deleted?

internal/pkg/controllers/profile/profile.go
cmurphy: fixed

internal/pkg/controllers/profile/profile.go
saschagrunert: Sticking in termination could be a good option. I’m not sure, WDYT?

@hasheddan @pjbgf any thoughts?

cmurphy: I think the implementation for causing it to go into a terminating state to wait for all workloads to be deleted would involve getting every pod that references it to add a finalizer to it, which i think would be a little messy. I think it would be easier to simply fail to delete it and emit an event/condition saying why it failed to delete.

saschagrunert: Sounds good. :+1:

cmurphy: In working through it I found that there wasn't really a way to just block deletion except with finalizers, so this latest update indexes pods that are using the profile and adds a finalizer until they are all removed.

saschagrunert (approved):
LGTM :+1: 

PTAL @hasheddan @pjbgf

cmurphy:
Squashed commits, removed WIP, added release note.

pjbgf (approved):
@cmurphy another great PR, thanks for that!

/lgtm

PR #162, opened 2020-12-02
Run e2e tests for namespaced deployment

### LINKED ISSUE #108: Make e2e tests (partially) run with namespaced deployment
The [namespaced deployment](https://github.com/kubernetes-sigs/seccomp-operator/blob/19835a6/deploy/namespace-operator.yaml) is currently not e2e tested. One thing to mention is that they will not run out of the box, because the `default-profiles` configMap is part of the target namespace and not the `seccomp-operator` one.

cmurphy:
#### What type of PR is this?

/kind cleanup

This change re-deploys the operator as a namespaced deployment and
re-runs the test cases for them. The test cases were adjusted to use a
configurable namespace instead of hardcoding "default" and
"security-profiles-operator".

Fixes #108

```release-note
NONE
```

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -25,13 +25,17 @@ import (
 )
 
 const (
-	manifest        = "deploy/operator.yaml"
-	defaultProfiles = "deploy/profiles/default-profiles.yaml"
+	manifest                 = "deploy/operator.yaml"
+	namespaceManifest        = "deploy/namespace-operator.yaml"
+	defaultProfiles          = "deploy/profiles/default-profiles.yaml"
+	namespaceDefaultProfiles = "deploy/profiles/namespace-default-profiles.yaml"
+	testNamespace            = "test-ns"
+	defaultNamespace         = "default"
 )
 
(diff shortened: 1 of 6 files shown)


test/e2e_test.go
```diff
@@ -63,13 +68,43 @@ func (e *e2e) TestSeccompOperator() {
...
+		namespaceManifest,
+	)
+	e.run(
+		"sed", "-i", fmt.Sprintf("s/NS_REPLACE/%s/", testNamespace),
+		namespaceDefaultProfiles,
+	)
+	defer e.run("git", "checkout", namespaceManifest)
+	defer e.run("git", "checkout", namespaceDefaultProfiles)
+	// All following operations such as create pod will be in the test namespace
+	e.kubectl("config", "set-context", "--current", "--namespace", testNamespace)
```
hasheddan, line 98: Potential future optimization: using unstructured client for these operations instead of kubectl directly

hasheddan (approved):
Thanks @cmurphy!

/lgtm

PR #164, opened 2020-12-08
Allow setting linkmode in Makefile

JAORMX:
Allow setting linkmode in Makefile

The current Makefile implementation assumes that the developer is using cgo.
This is not the case for some folks.

This change makes this optional by enabling folks to set a
`LINKMODE_EXTERNAL` environment variable when running Make. If set to anything
else than "true" (which is the default)  the go command will not pick up cgo as
the compiler implementation


saschagrunert (approved):
/lgtm

PR #165, opened 2020-12-08
Integrate selinux-operator into the SPO

### LINKED ISSUE #163: SELinux profile support
Some time ago, we noticed a gap in how folks install SELinux modules for their workloads, and came up with an operator to do just that: https://github.com/JAORMX/selinux-operator . The intent of that project is to be able to create SELinux policies as Custom Resources, and to be able to install them and take them into use on a per-namespace basis. So it comes with the controller that reads policies and schedules workloads to install them, as well as a webhook to validate that folks aren't able to use policies that are not installed in the namespaces. So, to some extent, it has a very similar model to the security-profile-operator.

If you'd like a bit more of an explanation and to see it in action, we gave a talk about it in DevConf this year: https://www.youtube.com/watch?v=iMO6rwA-i_s

### Let's merge the projects!

This would come in the form of adding the SELinux policy types, controllers, and webhook from the selinux-operator to be part of the selinux-profiles-operator instead.

## Why is this needed:

It would add SELinux support.

JAORMX:
This integrates the selinux-operator [1] into the SPO by copying in the
reconcilers/controllers and setting them up in the code-base.

Note that since the selinux-operator was meant to run as part of a
deployment, I had to create two managers in the main function. One for
the daemonset that handles the seccomp profiles, and another one for the
Selinux controllers which will do leader election, and thus effectively
only run one.

Note that this only includes the controllers and doesn't include the webhook
yet.

Closes #163 

```release-note
Initial SELinux policy support is implemented. This adds a CRD called `SelinuxPolicy`, which the operator uses to ensure policies are installed on the nodes.
```

[1] https://github.com/JAORMX/selinux-operator


JAORMX:
Pull-request updated, HEAD is now f0a947cb4c99265923310d957d6cf0d65f360675

JAORMX:
Pull-request updated, HEAD is now 6080c1eaa35d35d7544bcd5c1b057638a94ed90a

JAORMX:
Pull-request updated, HEAD is now e07f2c2f95692b492ab597c7a1bb8b3b5e0e0e06

JAORMX:
Pull-request updated, HEAD is now 0e865d7af293f1c78338c0d12829dbcfc73413f4

JAORMX:
Pull-request updated, HEAD is now 0e865d7af293f1c78338c0d12829dbcfc73413f4

JAORMX:
Pull-request updated, HEAD is now 658eb03b26054fe67b208485cd2022daa73ab0d8

JAORMX:
Pull-request updated, HEAD is now f1824e0fed2880d1d52f5526053dcee89077a098

JAORMX:
Pull-request updated, HEAD is now f1824e0fed2880d1d52f5526053dcee89077a098

JAORMX:
Pull-request updated, HEAD is now c12486d4ea710c22f20147691c66d8baac9c628d

JAORMX:
Pull-request updated, HEAD is now b4c5f250e5f91c1fa8ee1caad1be511d5813c532

JAORMX:
Pull-request updated, HEAD is now 0d184dd45b28e079b94b55774a9609f03543d46e

JAORMX:
Pull-request updated, HEAD is now b46b3d563709edc420a7a9c01fb517e22add8cdf

JAORMX:
Pull-request updated, HEAD is now 1c374e1dfa0c3a016bbc4de5e2604958b6f76ca9

saschagrunert (approved):
This looks pretty promising! I have to trust on the `selinuxpolicy` controller implementation defaults. I assume that we integrate unit tests for the controller too and setup the docs later on. Since this PR is already huge, let's get it in.

/lgtm
/assign @pjbgf 
/hold

@pjbgf feel free to lift the hold if you think this can be merged.

JAORMX:
@saschagrunert yep! tests and improvements are coming.

deploy/base/operator.yaml
```diff
@@ -11,6 +11,7 @@ spec:
   template:
     metadata:
       annotations:
+        openshift.io/scc: privileged
```
cmurphy, line 14: Is it common to have distro-specific configuration like this in general-purpose k8s operators?

internal/pkg/controllers/selinuxpolicy/utils.go
```diff
@@ -0,0 +1,125 @@
...
+	return err
+}
+
+// IgnoreAlreadyExists ignores "AlreadyExists" errors.
+func IgnoreAlreadyExists(err error) error {
+	if errors.IsAlreadyExists(err) {
+		return nil
+	}
+	return err
+}
```
cmurphy, line 125: There are similar functions for the seccomp controller, in the future it might be nice to split them into a shared utils package

internal/pkg/controllers/selinuxpolicy/configmap_controller.go
```diff
@@ -0,0 +1,290 @@
...
+			if containerStatus.State.Terminated != nil {
+				termState := containerStatus.State.Terminated
+				return termState.ExitCode, true
+			}
+		}
+	}
+	return -1, false
+}
+
+// newPodForPolicy returns a busybox pod with the same name/namespace as the cr.
```
cmurphy, line 148: Is it a busybox pod? It looks like it's using the udica image for both containers

cmd/security-profiles-operator/main.go
```diff
@@ -102,31 +104,76 @@ func run(ctx *cli.Context) error {
...
+		}
+
+		setupLog.Info("starting manager")
+		if err := mgr.Start(sigHandler); err != nil {
+			fatalErrors <- errors.Wrap(err, "controller manager error")
+		}
+		close(done)
+	}()
+
+	// cluster managers (need leader election)
```
cmurphy, line 141: The operator currently runs as a Daemonset and doesn't need leader election because it just runs the same operation on every node. If I'm getting this right, the new configmap controller spins up a separate pod to run the installer/uninstaller, and so it doesn't really need to be a Daemonset. I wonder if it would make sense to run the new controllers as a separate application, maybe in a Deployment? That would also allow us to split the RBAC and service accounts.

cmurphy (commented):
Super cool to see this! I added some questions inline, but I'm a new contributor too so I defer to others for their expertise :wink: Would be great to get this in and keep iterating with docs, tests, and TODO cleanups

cmd/security-profiles-operator/main.go
JAORMX: That's actually how the original implementation of the selinux operator worked. It was a deployment that spawned the installer pods as a daemonset. This was just how I was able to adapt the code to this new code base. Ideally we would switch back to having the controllers on a deployment as that's what package managers such as the OLM (for Operator Hub) expect and require.

deploy/base/operator.yaml
```diff
@@ -60,6 +61,10 @@ spec:
             capabilities:
               drop: ["ALL"]
               add: ["CHOWN", "FOWNER", "FSETID", "DAC_OVERRIDE"]
+            runAsUser: 0
```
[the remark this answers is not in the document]
JAORMX: Some distros require you to explicitly set that as far as I've understood.

deploy/base/rbac.yaml
```diff
@@ -42,3 +42,91 @@ roleRef:
...
+  - apps
+  resources:
+  - replicasets
+  - deployments
+  verbs:
+  - get
+- apiGroups:                  # Needed for using privileged containers
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
```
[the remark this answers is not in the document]
JAORMX: It's implicit in OpenShift and used when you set any security setting on a pod.

internal/pkg/controllers/selinuxpolicy/configmap_controller.go
JAORMX: I actually don't remember isf it's busybox or UBI haha. We'll actually change this implementation of the operator to spawn a single workload as opposed to a workload per policy. I'll start updating that when we have a decent proposal. First I'd like to start adding unit tests though

saschagrunert:
/hold cancel

Let’s follow up on the other points.

internal/pkg/controllers/selinuxpolicy/configmap_controller.go
```diff
@@ -0,0 +1,290 @@
...
+			Namespace: GetOperatorNamespace(),
+			Labels:    labels,
+		},
+		Spec: corev1.PodSpec{
+			Containers: []corev1.Container{
+				{
+					Name: "policy-installer",
+					// TODO(jaosorior): Add udica image or equivalent to a
+					// more formal registry
+					Image:   "quay.io/jaosorior/udica",
```
pjbgf, line 168: @JAORMX given how privileged this runs, and that is a direct dependency, it would be great if we could build a version of that image in this repo too, and maybe distribute as `gcr.io/k8s-staging-sp-operator/udica`? 

@saschagrunert @cmurphy any thoughts?

pjbgf (commented):
@JAORMX really good contribution, thanks for that. Although this is already merged, I added a few points we can progress as follow-up PRs.

internal/pkg/controllers/selinuxpolicy/configmap_controller.go
saschagrunert: Is this the Dockerfile for the image: https://github.com/containers/udica/blob/master/Dockerfile ?

Since the source code is part of a different repository I'm wondering what the best appraoch would be… Maybe we could just integrate the quay.io auto-build into contianers/udica to build the image for each push? I'm concerned about security vulnerabilities, for example in the python packages. If we hit that case, quay will not rebuild if there is no push into the repository, right?

JAORMX: @saschagrunert @pjbgf we're actually planning to move away from the udica implementation and have a go-based daemon instead. But yeah, we should figure out a way to deal with external container images.

PR #170, opened 2020-12-14
Enable e2e tests to run on OpenShift

JAORMX:
Enable e2e tests to run on OpenShift

This changes the e2e tests to not only run on a kind cluster, but to
also be able to run on a running OpenShift cluster.

To enable this, there is now the `E2E_CLUTER_TYPE` environment variable
which the Makefile can take into use and that will be taken into account
by the test itself.

Note that in order for the security-profiles-operator to work on
OpenShift, I temporarily added spc_t as the SELinux type. This ensures
that the operator is allowed to place files on the folder it's trying to
access. This is a placeholder while we work on a better mechanism to
consume both workload and bootstrap SELinux policies [1]. This mechanism
will eventually replace the tiny workloads that we currently spawn to
install policies and will be included as part of the main daemonset.

[1] https://github.com/JAORMX/selinuxd

This allows us to run the e2e tests on an OpenShift cluster. We can
subsequently enable the OpenShift infrastructure to run tests here and thus get
coverage there as well.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,13 +43,15 @@ else
   LDFLAGS := -s -w -extldflags "-static" $(LDVARS)
 endif
 
-CONTAINER_RUNTIME ?= docker
+export CONTAINER_RUNTIME ?= docker
 IMAGE ?= $(PROJECT):latest
 
 CRD_OPTIONS ?= "crd:crdVersions=v1"
 
 GOLANGCI_LINT_VERSION = v1.32.0
 REPO_INFRA_VERSION = v0.1.2
+
+export E2E_CLUSTER_TYPE ?= kind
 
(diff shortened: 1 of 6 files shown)


JAORMX:
Pull-request updated, HEAD is now b2c6d2078496917118efae33c5c8e8484a34b627

JAORMX:
Pull-request updated, HEAD is now 20624798d834a1177f66c9538d802a18cbdb2019

JAORMX:
Pull-request updated, HEAD is now ee54383746fe5429a01bd7a9a083ea82f1dc2ee2

pjbgf (approved):
/lgtm

PR #172, opened 2020-12-15
Enable UBI-based image

JAORMX:
Enable UBI-based image

This enables the deployment of a UBI-based image which can optionally be
used instead of the default scratch one.

It also introduces an variable in the Makefile called `DOCKERFILE` which
one can use to specify the Dockerfile to use.

#### What type of PR is this?

/kind feature

```release-note
Adds UBI-based Dockerfile.
```


pjbgf (approved):
/lgtm

PR #173, opened 2020-12-16
Add UBI Dockerfiles

JAORMX:
Add UBI Dockerfiles
    
These were forgotten from the PR that was supposed to introduce them
(https://github.com/kubernetes-sigs/security-profiles-operator/pull/172)

```release-note
NONE
```

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -1,0 +1,54 @@
+# Copyright 2020 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


Dockerfile.ubi
```diff
@@ -0,0 +1,54 @@
...
+    --shell "/sbin/nologin" \
+    --no-create-home \
+    --uid "${UID}" \
+    "${USER}"
+
+COPY --from=build /work/build/security-profiles-operator /usr/bin/
+
+ENTRYPOINT ["/usr/bin/security-profiles-operator"]
+
+USER ${USER_UID}
```
pjbgf, line 54:
```suggestion
USER ${UID}
```

The idea is to use the USER ID from line 40, right?

pjbgf (commented):
small nit around `UID`, otherwise it is a lgtm from me

Dockerfile.ubi
JAORMX: yeah, that's the idea... I wonder why the container build didn't fail here. Thanks for the review! let me fix this.

JAORMX:
Pull-request updated, HEAD is now 84f2e8e68ff1ea68d9ebbc1f81ec57749bd2f7d3

pjbgf (approved):
/lgtm

PR #174, opened 2020-12-16
Add SELinux e2e test

JAORMX:
This adds a couple of e2e test cases that verify that the base
functionality for the selinux bits work.

Each test is now prefixed with the component that's being tested.

On the other hand, to enable integration with other kube utilities (such as
`kubectl wait`) which would make testing easier, support for conditions
was added in the `SelinuxPolicy` object.

Some fixes related to object creation/deletion and status reporting had
to be done in the selinux-related controllers.

Note that the SELinux tests are being skipped for everything except
OpenShift clusters, since that's where I've been able to test this.
If someone has access to another distro that uses SELinux where we
could test this, we can integrate that too.

```release-note
Conditions were added to the SelinuxPolicy object's status.
```

### DIFF
--- a/api/v1alpha1/selinuxpolicy_types.go
+++ b/api/v1alpha1/selinuxpolicy_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	rcommonv1 "github.com/crossplane/crossplane-runtime/apis/common/v1"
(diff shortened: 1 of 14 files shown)


JAORMX:
Pull-request updated, HEAD is now bc6f5b8062789918dd1fc2db2e1fd1ad25cd0ab5

pjbgf (approved):
/lgtm

Issue #118, opened 2020-08-27
Expand the scope of the operator to AppArmor


saschagrunert:
Loading AppArmor profiles could be done easily by the operator, too. We probably have to rename the project to something like "security-operator", too.

saschagrunert:
cc @mrostecki

hasheddan:
@saschagrunert @pjbgf what are y'all thinking about potential rename? Would likely be good to do sooner rather than later if it is something we want to do.

saschagrunert:
> @saschagrunert @pjbgf what are y'all thinking about potential rename? Would likely be good to do sooner rather than later if it is something we want to do.

Yes I think that's a good idea. Do you wanna go ahead and create the issue? :+1:

pjbgf:
Yep, completely agree we rather do that soon. Will create a few issues on this now.

pjbgf:
Initial support done via #680

PR #176, opened 2020-12-16
Kh support apparmor 0

Kaizhe:
Add AppArmor profile support

```release-note
Add AppArmor profile support

```

### DIFF
--- a/api/v1alpha1/apparmorprofile_deepcopy.go
+++ b/api/v1alpha1/apparmorprofile_deepcopy.go
@@ -1,3 +1,19 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package v1alpha1
 
 import "k8s.io/apimachinery/pkg/runtime"
--- a/api/v1alpha1/apparmorprofile_deepcopy.go
+++ b/api/v1alpha1/apparmorprofile_deepcopy.go
@@ -13,7 +13,7 @@ func (in *AppArmorProfile) DeepCopyInto(out *AppArmorProfile) {
 	}
 }
 
-// DeepCopyObject returns a generically typed copy of an object
+// DeepCopyObject returns a generically typed copy of an object.
 func (in *AppArmorProfile) DeepCopyObject() runtime.Object {
 	out := AppArmorProfile{}
 	in.DeepCopyInto(&out)
@@ -21,7 +21,7 @@ func (in *AppArmorProfile) DeepCopyObject() runtime.Object {
 	return &out
 }
 
-// DeepCopyObject returns a generically typed copy of an object
+// DeepCopyObject returns a generically typed copy of an object.
 func (in *AppArmorProfileList) DeepCopyObject() runtime.Object {
 	out := AppArmorProfileList{}
 	out.TypeMeta = in.TypeMeta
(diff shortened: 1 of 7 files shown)


pjbgf:
Relates to #118

api/v1alpha1/apparmorprofile_types.go
```diff
@@ -0,0 +1,50 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+type AppArmorProfileSpec struct {
+	Rules    string `json:"rules"`
```
pjbgf, line 22: I think we could break this down into the different types of rules - similarly to what we have on the Seccomp CRD. This would allow us to validate profiles to quite a good degree.

pjbgf (commented):
@Kaizhe I added a few comments around the CRD, please have a look.

It would be great to get some views from @cmurphy and @saschagrunert on this too. :+1:

api/v1alpha1/apparmorprofile_types.go
Kaizhe: @pjbgf can you point me to a few examples?

api/v1alpha1/apparmorprofile_types.go
```diff
@@ -0,0 +1,50 @@
...
+	Enforced bool   `json:"enforced"`
```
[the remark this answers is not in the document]
Kaizhe: @pjbgf , this is a good discussion. The reason I didn't include different relative and attach is to simplify the usage. And to confine a process inside a pod/container: `attach_disconnect` is used which is the apparmor-manager's implementation. 

Happy to discuss more.

api/v1alpha1/apparmorprofile_types.go
pjbgf: On Seccomp we define the different types here: https://github.com/kubernetes-sigs/security-profiles-operator/blob/b9715d8142237f0ea4b64ff9c569370369d6e2cd/api/seccompprofile/v1alpha1/seccompprofile_types.go#L25

That allows us to statically verify that a profile is valid, which thinking about it is a slightly easier job considering that seccomp profiles are generally represented in `json` already. Some tools (e.g. `aa-easyprof`) support json [representation](http://manpages.ubuntu.com/manpages/xenial/man8/aa-easyprof.8.html) of AppArmor profiles, but to my knowledge users tend to use the unstructured approach more often.

We don't need to take the same approach for the AppArmor policy content, but it is worth discussing how can we strike a balance between the user experience on using existing policies and being able to statically verify the policy at creation time.

api/v1alpha1/apparmorprofile_types.go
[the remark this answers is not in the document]
pjbgf: Fair point, I think the other flags would depend on how we are representing the policy. If we go ahead and define the entire policy in an unstructured manner inside `Rules`, the users would define such flags (e.g. `Relative`, `Attach`) as per a typical AppArmor profile.
If we were to have a full representation of the profile, we would need to expose each one of those.

saschagrunert:
@Kaizhe thank you for the PR! I'm happy to move forward with the approach and iterate on the implementation. Can you please update your kustomize version and run `make deployments` to make the verification step within the CI happy?

.gitignore
```diff
@@ -1,2 +1,3 @@
 *.orig
 build
+.idea/
```
cmurphy, line 3: This could be removed, if you need to ignore IDE-specific files you can add them to your global .gitignore https://docs.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

api/v1alpha1/apparmorprofile_deepcopy.go
```diff
@@ -0,0 +1,54 @@
+/*
```
cmurphy, line 1: API files were moved under type-specific packages in https://github.com/kubernetes-sigs/security-profiles-operator/pull/178, so this will need to be moved and the Makefile updated

api/v1alpha1/apparmorprofile_types.go
cmurphy: I think this would be worthwhile too

cmurphy (commented):
This defines the CRD, is the plan to add controller code in another PR?

Please squash the commits :)

Kaizhe:
Need to work on a more fine grained data structure for the CRD

PR #179, opened 2020-12-17
Add ProfileBinding CRD and webhook

### LINKED ISSUE #145: Bind CRD profiles to workloads
Target is to find a way to make applying profiles easier rather than specifying the `LocalhostPath` in the security context.

#### Why is this needed:
Users may not know where the profile is actually saved on disk. We could implement a new resource like a `ProfileBinding` which uses a selector to mutate workloads and set the right path to the profile in it's annotation/securityContext.

It is questionable how to handle deletions of those bindings. The annotations/fields for seccomp and AppArmor behave differently with respect to their mutability.

cmurphy:
To do:

- [ ] Create a custom seccomp profile for the webhook deployment
- [x] As with RoleBindings, maybe this should have a profileRef attribute rather than type and profile attributes.
- [ ] Add support for SelinuxPolicy (future PR)
- [x] Unit tests
- [x] E2E tests
- [x] Documentation

#### What type of PR is this?

/kind feature
/kind api-change

Add a new CRD called ProfileBinding which defines a relationship between
a container image and a profile resource (currently only
SeccompProfile). When a pod that has a matching container image is
created, a mutating admission webhook injects a SecurityContext into it
to refer to the SeccompProfile. This change adds a new controller and
webhook as a separate Deployment.

Fixes #145 

See discussion questions above.

```release-note
Adds a new CRD `ProfileBinding` to define a relationship between a Pod and a profile resource. Currently only supports the SeccompProfile kind.
```


pjbgf:
@cmurphy great work and awesome description of the current status - inc. talking points and all. :+1:  
Here's my take on a few points:

> Is podSelector a good choice for matching ProfileBindings to pods, or
should the binding be 1:1 between pod and profile? See the comment in
api/v1alpha1/pod_webhook.go about the complications with matching
containers and init containers.
Relatedly to above, should this match to a list of Containers and
InitContainers, or one Container or InitContainer (singular)?

As we discussed on the meeting today, potentially using the actual images may be a way to make this easier to map.
Would we map per image, image + tag, image + tag regex?

> As with RoleBindings and ClusterRoleBindings, should we have
ProfileBindings and ClusterProfileBindings in order to allow bindings
across namespaces?

I would probably start with namespaced only and then progress to cluster-wide when we see the need. 


> A ProfileBinding needs to be created before a matching Pod can be
created, so this doesn't help with all-in-one manifests containing
both a profile and an application; the user would have to delete and
recreate the pod or restart the Deployment/DaemonSet.

What are your thoughts on potentially handling only Pod Templates? My understanding is that the majority of the Pod Templates would automatically apply the changes, unless it is a `Deployment` which is in `paused` state. This combined with using the `image` as the primary factor to link a policy to a Pod, may remove some of the complexities.


> The webhook has a dependency on cert-manager, should we keep that
external dependency or write our own certificate manager?

I would not oppose keeping that dependency to start with. we may keep that to start with.

cmd/security-profiles-operator/main.go
```diff
@@ -225,3 +233,48 @@ func runDaemon(ctx *cli.Context) error {
 	setupLog.Info("ending daemon")
 	return nil
 }
+
+func runBindingWebhook(ctx *cli.Context) error {
```
JAORMX, line 237: Why not run the webhook with the main deployment? as part of the `manager` subcommand. This way it would all be part of the same pod.

internal/pkg/controllers/binding/binding.go
```diff
@@ -0,0 +1,260 @@
...
+		})
+}
+
+type Reconciler struct {
+	client client.Client
+	log    logr.Logger
+	record event.Recorder
+}
+
+func (r *Reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
```
JAORMX, line 96: if we use kubebuilder more, we could set the rbac permissions dynamically here: https://book.kubebuilder.io/reference/markers/rbac.html this can ensure that we always have up-to-date permissions.

internal/pkg/controllers/binding/binding.go
```diff
@@ -0,0 +1,260 @@
...
+	logger := r.log.WithValues("pod", req.Name, "namespace", req.Namespace)
+
+	ctx, cancel := context.WithTimeout(context.Background(), reconcileTimeout)
+	defer cancel()
+
+	podID := req.Namespace + "/" + req.Name
```
JAORMX, line 102: NamespacedName actually "stringifies" on to the same construct: https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go#L37, so you could pass req.NamespacedName.String() instead of creating the podID.

internal/pkg/controllers/binding/binding.go
```diff
@@ -0,0 +1,260 @@
...
+
+	pod := &corev1.Pod{}
+	var err error
+	if err = r.client.Get(ctx, req.NamespacedName, pod); ignoreNotFound(err) != nil {
+		return reconcile.Result{}, errors.Wrap(ignoreNotFound(err), errGetBinding)
+	}
+	if kerrors.IsNotFound(err) { // this is a pod deletion, so update all seccomp profiles that were using it
+		return reconcile.Result{}, r.removeBoundPod(ctx, logger, podID)
```
JAORMX, line 110: I'm feeling a little shaky about this bit. the client we're using is a cached client, and so our references to objects might be a little old. Perhaps this is something better handled in a webhook?

cmurphy (commented):
[**@pjbgf**](https://github.com/pjbgf) commented on [Jan 14, 2021, 3:18 PM PST](https://github.com/kubernetes-sigs/security-profiles-operator/pull/179#issuecomment-760539273 "2021-01-14T23:18:59Z - Replied by Github Reply Comments"):

> What are your thoughts on potentially handling only Pod Templates? My understanding is that the majority of the Pod Templates would automatically apply the changes, unless it is a `Deployment` which is in `paused` state. This combined with using the `image` as the primary factor to link a policy to a Pod, may remove some of the complexities.

I think this would have close to the same effect as patching and recreating a pod on its own, which is a little aggressive and invasive. In the case of a `Deployment` with >1 replicas it might be less disruptive but for something like a `Job` or a `DaemonSet` it might have more of an impact. It would also inhibit the use case that I think we talked about yesterday where maybe a cluster administrator wants to set up a profilebinding to apply to their users' workloads - the users could just create a `Pod` to get around the policy.

I'm coming around to thinking that just documenting that profilebindings need to be applied before workloads can use them and that existing workloads needs to be restarted is not all that bad and this is still a useful way to go.

deploy/base/webhookconfig.yaml
```diff
@@ -0,0 +1,40 @@
+
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
```
saschagrunert, line 6: I think those can be removed, right?
```suggestion
```

installation-usage.md
```diff
@@ -137,6 +138,39 @@ spec:
     - exit_group
 ```
 
+#### Bind workloads to profiles with ProfileBindings
```
saschagrunert, line 141: Do we need to document the cert-manager dependency?

saschagrunert (approved):
Yay, just two nits from my side, otherwise LGTM

installation-usage.md
JAORMX: I think that would be good.

JAORMX:
This looks great! Let's just document the cert-manager dependency and this will be good to go.

installation-usage.md
saschagrunert: Let's also document what exactly happens when cert manager does not exist :+1:

cmurphy: Oops, good catch. Added now as part of the installation instructions.

Currently the webhook is bundled into the regular operator deployment, so failing to install cert-manager will result in the operator.yaml failing to deploy. We had talked about splitting the webhook into a separate installation, do you think we should do that so that cert-manager isn't always required?

saschagrunert: > Currently the webhook is bundled into the regular operator deployment, so failing to install cert-manager will result in the operator.yaml failing to deploy. We had talked about splitting the webhook into a separate installation, do you think we should do that so that cert-manager isn't always required?

I would be in favor if the operator still works either with or without cert-manager to be available. Maybe we can change the controller init logic that it will only warn if cert-manager is not available (aka the cert is not available) and then disable the feature automatically?

I would not split it into a separate deployment variant, but we could also think about making it manually configurable via the configmap.

Let's follow-up on that topic after this PR got merged. WDYT?

saschagrunert (approved):
/lgtm

PR #180, opened 2021-01-07
Change controller manager to be a Deployment

JAORMX:
Change controller manager to be a Deployment

Why?
====

While the current approach of deploying a DaemonSet directly works, it
has some limitations:

* It's not trivial to customize what to deploy (e.g. if we want to
  disable AppArmor)

* It's not directly supported by some operator catalogs (e.g.
  via the Operator Lifecycle Manager)

This instead proposes the usage of a Deployment.

This Deployment listens for a Configuration called "config", which
will allow us to customize the DaemonSet (e.g. enable/disable
features).

On the other hand, with we'll be able to use this Deployment to serve
webhooks (like the one that's currently missing to validate the SELinux
policy namespaces).

What's missing?
===============

Actually add configuration options
----------------------------------

Currently the controller only deploys the DaemonSet... only configuration
is taken into account: `SPOdImagePullPolicy`.

This was required for the kind-based e2e tests.

Handle configuration updates
----------------------------

Once we have configurations, we need to handle updates.

Better bindata management
-------------------------

Currently the DaemonSet is a hardcoded string, it would be better to
have an appropriate bindata solution which would read a yaml file and
update go code to handle it automatically.

Separate RBAC roles
----------------------------

Currently, both the controller manager and the daemon use the same ServiceAccount
with the same role. It would be ideal to separate them.

Merge SELinux to the DaemonSet
------------------------------

Currently SELinux works on a cluster level, and so it's part of the main
Deployment. This needs to be changed so SELinux module installation
should be done as part of the DaemonSet instead. This will require us to
change the inner workings of the SELinux bits (probably using selinuxd
[1])

Discussion
========

* How do folks feel about the configuration being a ConfigMap. Is that fine for everybody?
  or would folks prefer to have a dedicated CustomResource for this?

[1] https://github.com/JAORMX/selinuxd

```release-note
The main deployment method is now a Deployment object that requires a ConfigMap
called "config".
```

### DIFF
--- a/internal/pkg/controllers/spod/configmap_controller.go
+++ b/internal/pkg/controllers/spod/configmap_controller.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package spod
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/crossplane/crossplane-runtime/pkg/event"
+	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
+)
+
+// blank assignment to verify that ReconcileConfigMap implements `reconcile.Reconciler`.
+var _ reconcile.Reconciler = &ReconcileSPOd{}
+
+// ReconcileSPOd reconciles the SPOd DaemonSet object.
+type ReconcileSPOd struct {
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client   client.Client
+	scheme   *runtime.Scheme
+	baseSPOd *appsv1.DaemonSet
+	record   event.Recorder
+	log      logr.Logger
+}
+
+// Reconcile reads that state of the cluster for a ConfigMap object and makes changes based on the state read
+// and what is in the `ConfigMap.Spec`.
+func (r *ReconcileSPOd) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx := context.Background()
+	// Fetch the ConfigMap instance
+	cminstance := &corev1.ConfigMap{}
+	if err := r.client.Get(ctx, request.NamespacedName, cminstance); err != nil {
+		if kerrors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("error getting spod configuration: %w", err)
+	}
+	spodKey := types.NamespacedName{
+		Name:      r.baseSPOd.GetName(),
+		Namespace: config.GetOperatorNamespace(),
+	}
+	foundSPOd := &appsv1.DaemonSet{}
+	if getErr := r.client.Get(ctx, spodKey, foundSPOd); getErr != nil {
+		if kerrors.IsNotFound(getErr) {
+			return r.handleCreate(ctx, cminstance)
+		}
+		return reconcile.Result{}, fmt.Errorf("error getting spod DaemonSet: %w", getErr)
+	}
+
+	// NOTE(jaosorior): We gotta handle updates
+	return reconcile.Result{}, nil
+}
+
+func (r *ReconcileSPOd) handleCreate(ctx context.Context, cfg *corev1.ConfigMap) (reconcile.Result, error) {
+	r.log.Info("Creating SPOd")
+	newSPOd := r.getConfiguredSPOd(cfg)
+
+	if err := controllerutil.SetControllerReference(cfg, newSPOd, r.scheme); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error setting spod controller reference: %w", err)
+	}
+
+	if createErr := r.client.Create(ctx, newSPOd); createErr != nil {
+		if kerrors.IsAlreadyExists(createErr) {
+			return reconcile.Result{}, nil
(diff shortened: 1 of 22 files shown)


JAORMX:
Pull-request updated, HEAD is now ab783b89aa6a1cf145c10778e0f07df61927e033

JAORMX:
Pull-request updated, HEAD is now c953a36e50a0560b0dc79face965b6e5e3ac9905

JAORMX:
Pull-request updated, HEAD is now a53cd0f885438f891f6f1879e89ccd560b24051d

JAORMX:
Pull-request updated, HEAD is now bef1618270b4717a075c5991e16797684660a7a1

JAORMX:
Ready for review. Sorry for the noise, had to fix the e2e tests

saschagrunert:
In general I like the idea of switching to a Deployment. If @pjbgf has no concerns then I think we can get this in.

internal/pkg/controllers/spod/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
+)
+
+// blank assignment to verify that ReconcileConfigMap implements `reconcile.Reconciler`.
+var _ reconcile.Reconciler = &ReconcileConfigMap{}
+
+// ReconcileConfigMap reconciles a ConfigMap object.
```
cmurphy, line 40: I think this is reconciling the DaemonSet, not the ConfigMap, right? The configmap is managed by the user and not meant to be changed by the controller iiuc. This comment could be made clearer.

internal/pkg/controllers/spod/configmap_controller.go
```diff
@@ -0,0 +1,104 @@
...
+
+func (r *ReconcileConfigMap) handleSPOdCreate(cfg *corev1.ConfigMap) (reconcile.Result, error) {
+	r.log.Info("Creating SPOd")
+	newSPOd := r.getConfiguredSPOd(cfg)
+
+	if err := controllerutil.SetControllerReference(cfg, newSPOd, r.scheme); err != nil {
+		return reconcile.Result{}, fmt.Errorf("error setting spod controller reference: %w", err)
+	}
+
+	if createErr := r.client.Create(context.TODO(), newSPOd); createErr != nil {
```
cmurphy, line 86: Would probably be better to either use `context.Background()` or an existing context rather than leave in TODOs?

internal/pkg/controllers/spod/setup.go
```diff
@@ -0,0 +1,103 @@
...
+
+// SPOdTunables defines the parameters to tune/modify for the
+// Security-Profiles-Operator-Daemon.
+type DaemonTunables struct {
+	DaemonImage         string
+	NonRootEnablerImage string
+	WatchNamespace      string
+}
+
+// Setup adds a controller that reconciles seccomp profiles.
```
cmurphy, line 46: reconciles seccomp profiles? Should this be reconciles the operator DaemonSet?

cmurphy (commented):
I'm definitely in favor of splitting the logic between a Deployment and a DaemonSet, that will simplify a lot of things.

Has any thought been given to doing that split in the YAML manifests, rather than moving the whole DaemonSet installation into a controller loop? That would simplify the configuration problem a little. I'm not sure I understand what managing the DaemonSet in a controller buys us.

> How do folks feel about the configuration being a ConfigMap. Is that fine for everybody?
or would folks prefer to have a dedicated CustomResource for this?

Config in a ConfigMap seems the most sensible to me, I don't think a custom CRD would improve usability very much here.

JAORMX:
> I'm definitely in favor of splitting the logic between a Deployment and a DaemonSet, that will simplify a lot of things.
> 
> Has any thought been given to doing that split in the YAML manifests, rather than moving the whole DaemonSet installation into a controller loop? That would simplify the configuration problem a little. I'm not sure I understand what managing the DaemonSet in a controller buys us.

That could have been a possible approach, however, I'm leaning more towards reconciling the DaemonSet based on configuration values since it would have some advantages:
* Being able to turn off/on entire containers from the DS (e.g. SELinux/AppArmor pieces)
* Being able to manage future upgrade logic as part of the reconcile loop

> 
> > How do folks feel about the configuration being a ConfigMap. Is that fine for everybody?
> > or would folks prefer to have a dedicated CustomResource for this?
> 
> Config in a ConfigMap seems the most sensible to me, I don't think a custom CRD would improve usability very much here.

I wasn't sure. The ConfigMap is an easy way to provide configurations, and no other CRD would be needed. But, a CRD could provide us with SPO status, as well as better validation of the configuration. However, I don't have a strong opinion on this, and the ConfigMap was an easy start.

internal/pkg/controllers/spod/configmap_controller.go
JAORMX: You're right. Fixed :+1:

internal/pkg/controllers/spod/configmap_controller.go
JAORMX: done, I need to get into that habit! Thanks! Fixed.

internal/pkg/controllers/spod/setup.go
JAORMX: fixed.

saschagrunert:
> > Config in a ConfigMap seems the most sensible to me, I don't think a custom CRD would improve usability very much here.
> 
> I wasn't sure. The ConfigMap is an easy way to provide configurations, and no other CRD would be needed. But, a CRD could provide us with SPO status, as well as better validation of the configuration. However, I don't have a strong opinion on this, and the ConfigMap was an easy start.

If we're validating/applying the configuration only on startup of the operator, then a ConfigMap seems fine to me. I see the drawback of manually editing structures like JSON or even YAML in a ConfigMap, so I still lean towards a CRD. It might introduce a too big overhead in the first place, but later it may save us painful migration paths. Any thoughts on that?

JAORMX:
> > > Config in a ConfigMap seems the most sensible to me, I don't think a custom CRD would improve usability very much here.
> > 
> > 
> > I wasn't sure. The ConfigMap is an easy way to provide configurations, and no other CRD would be needed. But, a CRD could provide us with SPO status, as well as better validation of the configuration. However, I don't have a strong opinion on this, and the ConfigMap was an easy start.
> 
> If we're validating/applying the configuration only on startup of the operator, then a ConfigMap seems fine to me. I see the drawback of manually editing structures like JSON or even YAML in a ConfigMap, so I still lean towards a CRD. It might introduce a too big overhead in the first place, but later it may save us painful migration paths. Any thoughts on that?

The CRD will also make the reconciler code easier since we would get everything parsed correctly in the first place. But I suggest the following: Let's gather use-cases for the CRD and start drafting a proposal. Once we have those, we can migrate from this ConfigMap to the CRD if we see fit. It could turn out that the use-cases aren't many and we don't need to migrate. But it could also make it clearer why a CRD would be a better choice.

In the meantime, what if we leave this PR as-is so it doesn't become bigger? We're not doing a stable release yet, so we can iterate on further PRs.

saschagrunert:
> In the meantime, what if we leave this PR as-is so it doesn't become bigger? We're not doing a stable release yet, so we can iterate on further PRs.

I agree :+1:

saschagrunert (approved):
/lgtm

PR #181, opened 2021-01-12
Add ability to set GOFLAGS from Makefile

JAORMX:
Add ability to set GOFLAGS from Makefile

This enables us to set custom GOFLAGS as part of the build process. This
is specially useful for environments where there are expected GOFLAGS
and we need to overwrite them.

```release-note
NONE
```


saschagrunert (approved):
/ok-to-test
/lgtm

PR #184, opened 2021-01-13
Add Makefile targets for easy development in OpenShift

JAORMX:
Add Makefile targets for easy development in OpenShift

This adds the following targets:

`push-openshift-dev`
--------------------

Builds the operator image based on UBI, and pushes it to the OpenShift
cluster that your current kubectl context is pointing at.

`do-deploy-openshift-dev`
-------------------------

Builds `deploy/operator.yaml` based on an `openshift-dev` overlay which
contains overrides that point to the local OpenShift container image
registry. This is then deployed and a trigger is set so any changes to
the container image get automatically taken into use by the Deployment.

`deploy-openshift-dev`
----------------------

Triggers both of the aforementioned targets.

Note
----

These targets are only usable for development purposes.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,13 @@ else
 endif
 
 export CONTAINER_RUNTIME ?= docker
+
+ifeq ($(CONTAINER_RUNTIME), podman)
+    LOGIN_PUSH_OPTS="--tls-verify=false"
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):
/ok-to-test
/lgtm

jhrozek (approved):
This works for me!

PR #203, opened 2021-01-18
Fix UBI Dockerfile

JAORMX:
Fix UBI Dockerfile

A recent commit added the seccomp tag to the build [1], however, it
missed to add it to the UBI build. This fixes that. Note that since the
seccomp-static package is not available in UBI, we switched the
build-base to be fedora, and it'll eventually copy the binary to a UBI
based container.

[1] f504483024c68b621f6e43170de59e1cd2f47216

```release-note
NONE
```

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -12,9 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


jhrozek (approved):
This works for me.

saschagrunert (approved):
/lgtm

PR #205, opened 2021-01-18
Vendor dependencies

jhrozek:
Vendor dependencies to speed up builds and avoid waiting for go mod
download. Also makes sure the builds are better reproducible and more
resilient towards cache outages.

I ran:
    go mod tidy
    go mod vendor

Then the Makefile hunk sets `-mod=vendor` explicitly and Dockerfile
drops the two commands that copy go.mod/sum and fetch the modules.

```release-note

```


saschagrunert (commented):
/ok-to-test

JAORMX:
Seems we're having issues with `controller-gen`. Folks in other projects (e.g. metal3) have scripts to download this separately [1]. This way we wouldn't need to vendor it and have to include it in the project's dependencies.

[1] https://github.com/metal3-io/baremetal-operator/blob/master/hack/install-controller-gen.sh

saschagrunert:
> Seems we're having issues with `controller-gen`. Folks in other projects (e.g. metal3) have scripts to download this separately [1]. This way we wouldn't need to vendor it and have to include it in the project's dependencies.
> 
> [1] https://github.com/metal3-io/baremetal-operator/blob/master/hack/install-controller-gen.sh

Sounds reasonable to me :+1:

cmurphy:
I'm personally not in favor of vendoring dependencies, I find the code bloat and maintenance overhead to be irksome and I feel like it's an anti-pattern that the Go community seemed to be moving away from. Am I in the minority?

JAORMX:
> I'm personally not in favor of vendoring dependencies, I find the code bloat and maintenance overhead to be irksome and I feel like it's an anti-pattern that the Go community seemed to be moving away from. Am I in the minority?

I understand the hesitance... vendoring dependencies is indeed a little funky and am not too keen on that either. However, there are a couple of things that this is addressing:
* Faster container build times (not having to pull the go modules on each container build indeed helps with this)
* OpenShift CI integration (they overwrite the build to expect vendored dependencies...)

naveensrinivasan:
>Also makes sure the builds are better reproducible 

I agree with the fact it makes builds reproducible. 


> * Faster container build times (not having to pull the go modules on each container build indeed helps with this)

The docker has cache now if we want to avoid downloading go modules https://www.docker.com/blog/containerize-your-go-developer-environment-part-2/

> and more resilient towards cache outages.

How often have the caches have been down? Not that often AFAIK. 



> I'm personally not in favor of vendoring dependencies, I find the code bloat and maintenance overhead to be irksome and I feel like it's an anti-pattern that the Go community seemed to be moving away from. Am I in the minority? 

I concur. I am not a big fan of vendoring dependencies

pjbgf:
Another positive point I can see on vendoring is the ability to easily identify changes throughout the dependency chain whenever a PR is adding new dependencies or updating existing ones.

Solely from a security point of view, I can see the benefit of being able to easily see all the changes brought into the code base instead of tacitly approving and pushing the responsibility down the dependency chain.

saschagrunert:
@JAORMX can you please fix the verification step:
```
go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen "crd:crdVersions=v1" paths="./api/seccompprofile/..." output:crd:stdout > deploy/base/crd.yaml
cannot find package "." in:
	/home/prow/go/src/github.com/kubernetes-sigs/security-profiles-operator/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen
make: *** [Makefile:162: manifests] Error 1
```

jhrozek:
On Wed, Jan 27, 2021 at 12:00:26AM -0800, Sascha Grunert wrote:
> @JAORMX can you please fix the verification step:
> ```
> go run -tags generate sigs.k8s.io/controller-tools/cmd/controller-gen "crd:crdVersions=v1" paths="./api/seccompprofile/..." output:crd:stdout > deploy/base/crd.yaml
> cannot find package "." in:
> 	/home/prow/go/src/github.com/kubernetes-sigs/security-profiles-operator/vendor/sigs.k8s.io/controller-tools/cmd/controller-gen
> make: *** [Makefile:162: manifests] Error 1
> ```

Sorry it's taking so long to get back to this, I'm taking another look
today.

jhrozek:
On Tue, Jan 26, 2021 at 10:33:12AM -0800, Paulo Gomes wrote:
> Another positive point I can see on vendoring is the ability to easily identify changes throughout the dependency chain whenever a PR is adding new dependencies or updating existing ones.

Not only identify, but I've seen a couple of cases where we had to patch
a dependency while a PR for it was being proposed. Much easier to do with
vendoring.

Makefile
```diff
@@ -35,14 +35,14 @@ BUILDTAGS := netgo osusergo
 endif
 
 ifneq ($(shell uname -s), Darwin)
-LINT_BUILDTAGS := e2e,netgo,osusergo,seccomp
+LINT_BUILDTAGS := e2e,netgo,osusergo,seccomp,-tools
```
jhrozek, line 38: Even if I added `//nolint` to tools.go the linter would still bark at me during `make verify` so I used this as a workaround. If  someone could make the `//nolint` tag work instead, I'd be glad.

JAORMX:
/approve

Will wait for feedback from more folks.

saschagrunert (approved):
/lgtm

saschagrunert:
I think we have to update this PR again:
```
go: inconsistent vendoring in /home/prow/go/src/github.com/kubernetes-sigs/security-profiles-operator:
	k8s.io/klog/v2@v2.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	k8s.io/klog/v2@v2.4.0: is marked as explicit in vendor/modules.txt, but not explicitly required in go.mod
```

saschagrunert (approved):
/lgtm

PR #206, opened 2021-01-18
Use typed bindata for DaemonSet

saschagrunert:
#### What type of PR is this?

/kind cleanup

This allows us more type-safe refactorings and reduces some possible
error paths per design.

None
None

```release-note
None
```


JAORMX (commented):
/lgtm

JAORMX:
One reason why I initially submitted this with yaml format as opposed to the straight library-go implementation was that I was hoping that once we upgrade to go 1.16, we could use the new go file embedding and start linting the daemon yaml with a standard kubernetes linter. But I'm not opposed to this change.

jhrozek (approved):
Looks good, the typed manifest is the same as the previous untyped manifest and basic smoke test succeeded.

PR #210, opened 2021-01-19
Simplify operator init container by using internal vars

saschagrunert:
#### What type of PR is this?

/kind cleanup

If we programmatically create the DaemonSet then we can also use internal variables directly without having a need to set them via Kubernetes environment variables. This patch also clarifies/simplifies some code paths inherited by this change.

None
None

```release-note
None
```


hasheddan (approved):
/lgtm

Issue #198, opened 2021-01-15
Separate ServiceAccount/RBAC permissions for DaemonSet and Deployment


JAORMX:
## Context

With the recent change of making the DaemonSet be deployed by a Deployment workload ( https://github.com/kubernetes-sigs/security-profiles-operator/pull/180 ). The operator is left using the same ServiceAccount with the same permissions for both workloads. 

## What would you like to be added:

We should move the workloads to have individual service accounts so they can have their own RBAC permissions.

## Why is this needed:

If one of these pieces would be compromised, this reduces the attack vector.

JAORMX:
Fixed by https://github.com/kubernetes-sigs/security-profiles-operator/pull/325

PR #213, opened 2021-01-20
WIP: Separate RBAC permissions for spo-manager and spod

JAORMX:
This separates the RBAC permissions for the SPO manager and the SPO
daemon. They now have the following service accounts with matching roles
and cluster roles:

* spo-manager
* spod

They are managed by the `make generate` target which, in turn, calls
`controller-gen` to generate the necessary RBAC roles. These are set as
comments that kubebuilder reads.

#### What type of PR is this?

/kind feature

By splitting the RBAC permissions, the pods are only allowed to do what they
need to function, and thus a compromise of a pod won't do as much damage as it
would if the permissions are shared.

No

Closes: https://github.com/kubernetes-sigs/security-profiles-operator/issues/198

```release-note
The service accounts and (cluster)roles are now split into: `spo-manager` and `spod`
```


naveensrinivasan:
@JAORMX  Was there a particular reason you had closed this PR? The reason I am asking is that to know if you run into any issues. I was thinking of basing of work based on your changes.

JAORMX:
@naveensrinivasan I started it before the webhooks were introduced and some extra changes... so I ditched it cause it was easier than rewriting.

PR #214, opened 2021-01-20
Take selinuxd into use

jhrozek:
Changes SELinux integration by ditching the intermediate configMaps 
previously used by the CM reconcile loop of the selinux controller. 
Instead, the daemon running in the DS now takes an boolean argument that 
signifies if the SELinux feature should be enabled. This is false by 
default.

If true, the DS deploys an additional container running selinuxd which 
listens to any files under /etc/selinux.d and can be queried for status 
over a UNIX socket mounted under /var/run/selinuxd. Both the socket 
directory and the drop-in directory are EmptyDirs shared between selinuxd
and the controller.

In addition, the operator receives a --with-selinux=true flag, which makes
the operator create an additional manager that listens for the 
selinuxpolicy API object. Contents of those objects are reconciled to the
/etc/selinux.d drop-in directory and then the controller queries the state
of selinuxd to make sure the policy was indeed added or removed.

### Release note
```release-note
The  selinux component can now be enabled or disabled through the  CongfiMap named  config by toggling a boolean option called EnableSelinux
 
Since not all Linux distributions support SeLinux, its support is disabled by default.
```

TODO:
   - [ ] The SELinux controller should be on by default for OpenShift
         through kustomize
   - [ ] optimize writing the policy files to write only if the file
          content differs when selinux'd `/ready` endpoint merges
   - [ ] take errors from selinuxd into use, extend the status to present them and add a test to exercise a failing selinux policy
   - [ ] the selinuxd container currently sets no resource requests/limits
   - [ ] make the selinuxd container read-only, needs to mount the selinuxd database in a separate volume
   - [ ] open a ticket to consider adding events

/kind feature


JAORMX:
@jhrozek can you add a release note about the flag to enable/disable SELinux?

deploy/base/kustomization.yaml
```diff
@@ -27,6 +27,10 @@ patchesStrategicMerge:
                 value: gcr.io/k8s-staging-sp-operator/security-profiles-operator:latest
               - name: RELATED_IMAGE_NON_ROOT_ENABLER
                 value: bash:5.0
+              - name: RELATED_IMAGE_POLICIES_COPIER
+                value: quay.io/jaosorior/udica
+              - name: RELATED_IMAGE_SELINUXD
+                value: quay.io/jaosorior/selinuxd
```
JAORMX, line 33: let's unify this by adding the `udica` dependency to the `selinuxd` image. This way we don't need the extra image.

deploy/base/operator.yaml
```diff
@@ -31,6 +31,10 @@ spec:
               value: security-profiles-operator
             - name: RELATED_IMAGE_NON_ROOT_ENABLER
               value: non-root-enabler
+            - name: RELATED_IMAGE_POLICIES_COPIER
+              value: quay.io/jaosorior/udica
+            - name: RELATED_IMAGE_SELINUXD
+              value: quay.io/jaosorior/selinuxd
```
JAORMX, line 37: same comment as above.

cmd/security-profiles-operator/main.go
```diff
@@ -38,8 +41,12 @@ import (
 
 const (
 	jsonFlag               string = "json"
+	seccompFlag            string = "with-seccomp"
```
JAORMX, line 44: This flag doesn't seem to be used anywhere.

deploy/base/config.yaml
```diff
@@ -7,3 +7,4 @@ metadata:
     security-profiles-operator/config: ""
 data:
   SPOdImagePullPolicy: Always
+  EnableSeccomp: "true"
```
JAORMX, line 10: Let's add `EnableSelinux: "false"` here too. It's better to document all the available options.

JAORMX (changes requested):

test/e2e_test.go
```diff
@@ -144,6 +144,13 @@ func (e *e2e) deployOperator(manifest, profiles string) {
 		manifest,
 	)
 
+	if e.platformSupportsSelinux() {
+		e.run(
+			"sed", "-i", "/EnableSeccomp/a \\ \\ EnableSelinux: \"true\"",
```
JAORMX, line 149: This `sed` could be simplified once you always add `EnableSelinux` in the sample config.

deploy/base/kustomization.yaml
jhrozek: I'm not sure I understand what you are proposing. Copy the files from `udica` to the `selinuxd` image as part of the image build process and then use `selinuxd` as the init container that primes the selinux drop-directory?

cmd/security-profiles-operator/main.go
jhrozek: Yeah, I think I'll just remove it from now. In fact, I think I might even remove the EnableSeccomp flag and make only SELinux optional until we figure out a nicer way to instantiate the DS based on options.

deploy/base/config.yaml
jhrozek: Will do.

test/e2e_test.go
jhrozek: Will do.

deploy/base/kustomization.yaml
JAORMX: in the selinuxd Dockerfile:
```
microdnf install -y udica
```
That would basically install the base policies that `udica` comes with.
Then we could use that same image as the init container that primes the drop-directory.

jhrozek: oh, right. Packages still exist :-)

test/e2e_test.go
jhrozek: Done

deploy/base/config.yaml
jhrozek: Done

cmd/security-profiles-operator/main.go
jhrozek: Removed

deploy/base/operator.yaml
jhrozek: Removed

deploy/base/kustomization.yaml
jhrozek: Done in selinuxd.
Here's a follow-up question: I think we can ditch the second init container entirely and change the ownership of `/etc/selinux.d` in the init container. Then we could copy the files into the directory when selinuxd starts with a lifecycle handler. It is maybe a bit hacky to have the logic split over two places, but at least it would allow us to keep the DS template more minimal and better manageable.

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -82,89 +88,212 @@ var Manifest = &appsv1.DaemonSet{
...
-						Limits: v1.ResourceList{
-							v1.ResourceMemory:           resource.MustParse("64Mi"),
-							v1.ResourceCPU:              resource.MustParse("250m"),
-							v1.ResourceEphemeralStorage: resource.MustParse("50Mi"),
+						SecurityContext: &v1.SecurityContext{
+							AllowPrivilegeEscalation: &falsely,
+							ReadOnlyRootFilesystem:   &truly,
+							Capabilities: &v1.Capabilities{
+								Drop: []v1.Capability{"ALL"},
+								Add:  []v1.Capability{"CHOWN", "FOWNER", "FSETID", "DAC_OVERRIDE"},
```
JAORMX, line 160: Why was the DAC_OVERRIDE capability needed again?

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -82,89 +88,212 @@ var Manifest = &appsv1.DaemonSet{
...
+										FieldPath: "spec.nodeName",
+									},
 								},
 							},
 						},
 					},
-				}},
+					{
+						Name:  "selinuxd",
+						Image: "quay.io/jhrozek/selinuxd",
```
JAORMX, line 239: let's switch to `jaosorior/selinuxd` as that's what's currently being built automatically when we push to the selinuxd repo. We should now add a github issue about moving that to an appropriate org. (First we gotta figure what that org is...)

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -82,89 +88,212 @@ var Manifest = &appsv1.DaemonSet{
...
+								MountPath: "/etc/selinux",
+							},
+							{
+								Name:      "host-varlibselinux-volume",
+								MountPath: "/var/lib/selinux",
+							},
+						},
+						SecurityContext: &v1.SecurityContext{
+							// TODO(jhrozek): If we wanted RO fs, we would need to mount
+							// the selinuxd db elsewhere than /var/run/selinuxd.db
```
JAORMX, line 271: Let's open up a Github issue for this. We could either make that directory configurable, or just expect a directory inside of `/var/run`.

deploy/base/config.yaml
```diff
@@ -7,3 +7,4 @@ metadata:
     security-profiles-operator/config: ""
 data:
   SPOdImagePullPolicy: Always
+  EnableSelinux: "false"
```
saschagrunert, line 10: Can we directly use a bool here?
```suggestion
  EnableSelinux: false
```

saschagrunert (approved):

deploy/base/config.yaml
jhrozek: Haha, yes, I'll see about fixing this

internal/pkg/controllers/spod/bindata/spod.go
jhrozek: Oops, sorry. This was a leftover from when I was testing some custom image of my own.

internal/pkg/controllers/spod/bindata/spod.go
jhrozek: I..might have copied the capabilities of the other init container blindly. I'll test what is really needed. But yeah, DAC_OVERRIDE might not be since we're not accessing anything the usual UNIX permissions wouldn't allow us to access.

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
```diff
@@ -161,63 +166,96 @@ func (r *ReconcileSP) removeFinalizer(sp *spov1alpha1.SelinuxPolicy) (reconcile.
...
+		return res, err
+	}
+
+	l.Info("Checking if policy is removed", "policyName", sp.Name)
+	installed, err := r.isPolicyInstalled(sp)
+	if err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "Looking up policy status")
+	}
+
+	if installed {
```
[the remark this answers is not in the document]
jhrozek: Yes, I planned to iterate on this PR by taking the newer features of selinuxd into account in subsequent PRs.

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
```diff
@@ -241,3 +279,35 @@ func (r *ReconcileSP) wrapPolicy(cr *spov1alpha1.SelinuxPolicy) string {
...
+
+	httpc := http.Client{
+		Transport: &http.Transport{
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return net.Dial("unix", bindata.SelinuxdSocketPath)
+			},
+		},
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "GET", "http://unix/policies/", nil)
```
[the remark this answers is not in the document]
jhrozek: I don't think that feature was available in selinuxd at that point. But it seems easy enough to add, I'll see about adding it to this PR.

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
```diff
@@ -161,63 +166,96 @@ func (r *ReconcileSP) removeFinalizer(sp *spov1alpha1.SelinuxPolicy) (reconcile.
...
-		if kerrors.IsNotFound(getErr) {
-			// this is good. Nothing to do.
-			return reconcile.Result{}, nil
-		}
+func (r *ReconcileSP) reconcilePolicyFile(sp *spov1alpha1.SelinuxPolicy, l logr.Logger) error {
+	policyPath := path.Join(bindata.SelinuxDropDirectory, GetPolicyName(sp.Name, sp.Namespace)+".cil")
+	policyContent := []byte(r.wrapPolicy(sp))
+
+	l.Info("Writing to policy file", "policyPath", policyPath)
+	err := ioutil.WriteFile(policyPath, policyContent, 0600)
```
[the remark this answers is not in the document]
jhrozek: It does trigger an inotify. I'll add a check if the contents are already written before writing the file (I'll add a simple check for now, just comparing the contents)

internal/pkg/controllers/spod/bindata/spod.go
jhrozek: I'll do that

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
```diff
@@ -161,63 +166,96 @@ func (r *ReconcileSP) removeFinalizer(sp *spov1alpha1.SelinuxPolicy) (reconcile.
...
-	cm := r.newConfigMapForPolicy(sp)
-
-	// Check if this cm already exists
-	foundCM := &corev1.ConfigMap{}
-	err := r.client.Get(context.TODO(), types.NamespacedName{Name: cm.Name, Namespace: cm.Namespace}, foundCM)
-	if err != nil && kerrors.IsNotFound(err) {
-		l.Info("Creating a new ConfigMap", "ConfigMap.Namespace", cm.Namespace, "ConfigMap.Name", cm.Name)
-		if err = r.client.Create(context.TODO(), cm); err != nil {
-			return reconcile.Result{}, IgnoreAlreadyExists(err)
+func (r *ReconcileSP) reconcilePolicy(sp *spov1alpha1.SelinuxPolicy, l logr.Logger) (reconcile.Result, error) {
```
naveensrinivasan, line 169: The reconciliation isn't reporting any k8s events. k8s events help with troubleshooting for the operators compared to logs. It also helps e2e tests to validate the condition.

Could we add events for the invalid and valid conditions?

JAORMX: This PR is already quite big. Mind if we file a Github issue and do that in a subsequent PR?

saschagrunert:
Unfortunately this now needs a rebase after the merge of https://github.com/kubernetes-sigs/security-profiles-operator/pull/179

deploy/base/kustomization.yaml
jhrozek: As discussed on Slack, let's do this if/when we need to make the DS manageable better. I added a comment in the meantime.

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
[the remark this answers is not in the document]
jhrozek: So I tried removing the write on every reconcile but found a race in selinuxd instead. There is a PR for selinuxd to address it now, but so far I submit this PR with the write on every reconcile and would take the selinuxd fixed into effect once that PR is merged.

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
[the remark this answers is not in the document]
jhrozek: Done

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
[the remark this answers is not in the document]
jhrozek: Done, but we still need to extend the status subresource and add a test. So far I added a TODO item into the PR comment and will file issues for the follow-up work.

internal/pkg/controllers/spod/bindata/spod.go
jhrozek: Dropped.

internal/pkg/controllers/spod/bindata/spod.go
jhrozek: Fixed.

deploy/base/config.yaml
jhrozek: Hmm, I must be misunderstanding something. I'm getting errors trying to use a straight-up boolean in a configmap, seems like CMs only allow string values?

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
jhrozek: I added a todo note to the opening comment and I'll open all the tickets at the same time

test/suite_test.go
```diff
@@ -149,10 +158,13 @@ func (e *kinde2e) SetupTest() {
 	// Build and load the test image
 	e.logf("Building operator container image")
 	e.run("make", "image", "IMAGE="+e.testImage)
-	e.logf("Loading container image into nodes")
+	e.logf("Loading container images into nodes")
 	e.run(
 		e.kindPath, "load", "docker-image", "--name="+e.clusterName, e.testImage,
 	)
+	e.run(
+		e.kindPath, "load", "docker-image", "--name="+e.clusterName, "quay.io/jaosorior/selinuxd"
```
JAORMX, line 166: you need a trailing `,`

JAORMX (changes requested):

test/suite_test.go
jhrozek: I like to live dangerously and submit patches to CI without even running make.

JAORMX (approved):
/approve

will wait for @saschagrunert to add lgtm

internal/pkg/controllers/selinuxpolicy/selinuxpolicy_controller.go
naveensrinivasan: That works me. Thanks 👍

saschagrunert (approved):
/lgtm

PR #219, opened 2021-01-22
Update golangci-lint and run tests in parallel

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now enable additional linters with the bump of golangci-lint. The
tests are now using the t.Parallel() method to let them executed
simultaneously.

None
None

```release-note
None
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,8 +1,10 @@
 ---
 run:
   build-tags:
-    - netgo
     - e2e
+    - netgo
+    - osusergo
+    - seccomp
   concurrency: 6
   deadline: 5m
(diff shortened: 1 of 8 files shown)


JAORMX:
Now that the tests are being parallelized, should we enable the `-race` flag in both the build and the test invocation?

saschagrunert:
> Now that the tests are being parallelized, should we enable the `-race` flag in both the build and the test invocation?

Good point, I enabled it for the tests. 

I think we should keep it disabled for the (production) build, because I'm not sure about the performance implications when enabling the race detector.

Issue #199, opened 2021-01-15
Create distinct cluster-bound and namespace-bound resources


JAORMX:
## Context

Currently, the resources (SeccompProfile and SelinuxPolicy) are bound according to the deployment of the operator. For this, we have two distinct [kustomize overlays](https://github.com/kubernetes-sigs/security-profiles-operator/tree/master/deploy/overlays) called `cluster` and `namespaced`, which set the RBAC and Deployment mode, amongst other things.

## What would you like to be added:

The proposal is to have two types per resource:

* A "cluster" scoped resource which would be accessible from all namespaces
* A "namespaced" resource which would only be accessible from the namespace it was created.

e.g.

We would have:

* `ClusterSeccompProfile` and `ClusterSelnuxPolicy` as cluster-scoped resources
* `SeccompProfile` and `SelinuxPolicy` as namespace-scoped resources

This is similar to how other kube resources are scoped. e.g. `ClusterRole` and `Role`.

### Less maintenance

This would allow us to have a more unified deployment model (no need for two overlays) where we could start auto-generating our  RBAC rules with kubebuilder (less maintenance burden).

### Unifies use-cases

This would also enable the use-cases of both overlays in one operator, as opposed to having to deploy the two overlays.

### Shared policies

Finally, it would also allow deployers to create cluster-wide policies for common scenarios. e.g. A cluster-wide SeccompProfile that's usable for nginx-based web-services. Validation of permissions can be done via a webhook.

saschagrunert:
I really like this idea and I'm in favor of adding it. I think we may have to refactor some code parts to be re-usable across both (cluster and namespaced) types. :+1:

saschagrunert:
Who wants to work on the seccomp bits? @cmurphy @pjbgf I can also take it but would like to give you precedence :innocent:

JAORMX:
@saschagrunert I can work on this.

saschagrunert:
@JAORMX do we wanna bring this into v0.3.0, too?
/remove-lifecycle stale

JAORMX:
@saschagrunert I don't think this is needed for v0.3.0

JAORMX:
I think we can close this, we have no concrete use-cases really.

PR #228, opened 2021-01-25
Feature - use the downward api to get the namespace

naveensrinivasan:
#### What type of PR is this?
The use downward API to get the pod namespace https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/

/kind cleanup

The namespace was hardcoded before this. The operator could be deployed in a non-standard namespace, and this will not break if it is so.

None

The code changes have unit-tests. 

NONE

```release-note
Utilized the downard API to get the Pod namespace for configmap controller.

```

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -16,6 +16,13 @@ limitations under the License.
 
 package config
 
+import (
+	"errors"
+	"os"
+
+	"k8s.io/release/pkg/util"
+)
+
 const (
 	// OperatorName is the name when referring to the operator.
 	OperatorName = "security-profiles-operator"
@@ -47,15 +54,25 @@ const (
 	SPOdImagePullPolicy = "SPOdImagePullPolicy"
 )
 
+var ErrPodNamespaceEnvNotFound = errors.New("the env variable MY_POD_NAMESPACE hasn't been set")
+
 // GetOperatorNamespace gets the namespace that the operator is currently running on.
-func GetOperatorNamespace() string {
-	// TODO(jaosorior): Get a method to return the current operator
-	// namespace.
-	//
-	// operatorNs, err := k8sutil.GetOperatorNamespace()
-	// if err != nil {
-	// 	return "security-profiles-operator"
-	// }
-	// return operatorNs
-	return "security-profiles-operator"
+func GetOperatorNamespace() (string, error) {
+	// This is MY_POD_NAMESPACE should have been set by the downward API to identify
+	// the namespace which this controller is running from
+	MyPodNamespace := util.EnvDefault("MY_POD_NAMESPACE", "")
+	if MyPodNamespace == "" {
+		return "", ErrPodNamespaceEnvNotFound
+	}
+	return MyPodNamespace, nil
+}
+
+// GetEnvDefault returns the value of the given environment variable or a
+// default value if the given environment variable is not set.
+func GetEnvDefault(variable, defaultVal string) string {
+	envVar, exists := os.LookupEnv(variable)
+	if !exists {
+		return defaultVal
+	}
+	return envVar
 }
--- a/internal/pkg/controllers/spod/configmap_controller.go
+++ b/internal/pkg/controllers/spod/configmap_controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
@@ -61,9 +62,15 @@ func (r *ReconcileSPOd) Reconcile(request reconcile.Request) (reconcile.Result, 
 		}
 		return reconcile.Result{}, fmt.Errorf("error getting spod configuration: %w", err)
 	}
+
+	namespace, err := config.GetOperatorNamespace()
+	if err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "getting operator namespace")
+	}
+
 	spodKey := types.NamespacedName{
 		Name:      r.baseSPOd.GetName(),
-		Namespace: config.GetOperatorNamespace(),
+		Namespace: namespace,
 	}
 	foundSPOd := &appsv1.DaemonSet{}
 	if getErr := r.client.Get(ctx, spodKey, foundSPOd); getErr != nil {
--- a/internal/pkg/controllers/spod/setup.go
+++ b/internal/pkg/controllers/spod/setup.go
@@ -18,10 +18,12 @@ package spod
 
 import (
 	"context"
+	"sync"
 
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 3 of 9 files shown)


saschagrunert (commented):
/ok-to-test

@JAORMX does this change influence the way how https://github.com/kubernetes-sigs/security-profiles-operator/issues/199 should be handled?

JAORMX:
I don't think this is the way to go, as it would restrict the operator to only work on one namespace. While we want this to work on one namespace, many, or on the whole cluster. In the PR the env variable`MY_POD_NAMESPACE` was used and that's fine, we just need to make sure that similar functionality is available. I don't think that's possible by using the downward API though.

So... in operator-sdk (which uses controller-runtime as well), the typical environment variable is called `WATCH_NAMESPACE`. From here on I'll refer to the env variable with this name... In that implementation, in a similar fashion as this PR, one is able to take it into use to make the operator listen on a single namespace: https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#using-environment-variables-for-namespace

There is, however, a mode to listen on multiple-namespaces. This consists of replacing the default cache for a `MultiNamespacedCacheBuilder`: https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#configuring-cluster-scoped-operators-with-multinamespacedcachebuilder . This requires a custom `WATCH_NAMESPACE` that doesn't include a single namespace, but instead is a comma-delimited string with the target namespaces... Some operators implement this from an annotation that some package managers are able to set. In the case of the `operator-sdk`/`Operator Lifecycle Manager`, the annotation is called `olm.targetNamespaces`.

Finally, if we'd like the operator to listen to all namespaces, one can simply set `WATCH_NAMESPACE` to an empty string.

On the operator's package definition, one then is able to set [relevant flags](https://sdk.operatorframework.io/docs/building-operators/golang/operator-scope/#watching-resources-in-a-single-namespace-1) which will set the relevant annotation, and thus this would configure the operator differently.

So, while we definitely want the operator to work on any namespace, as this PR enables, I think we'd also want to enable the other use-cases. Thus, I think we shouldn't use the downward API for this, and try to use an annotation instead.

What do you folks think?

internal/pkg/controllers/spod/configmap_controller.go
```diff
@@ -61,9 +62,15 @@ func (r *ReconcileSPOd) Reconcile(request reconcile.Request) (reconcile.Result,
 		}
 		return reconcile.Result{}, fmt.Errorf("error getting spod configuration: %w", err)
 	}
+
+	namespace, err := config.GetOperatorNamespace()
+	if err != nil {
+		return reconcile.Result{}, errors.Unwrap(err)
```
cmurphy, line 68: Why use `.Unwrap` here? That will lose the error:

```
$ cat err.go 
package main

import (
        "errors"
        "fmt"
)

func main() {
        var ErrPodNamespaceEnvNotFound = errors.New("the env variable MY_POD_NAMESPACE hasn't been set")
        unwrapped := errors.Unwrap(ErrPodNamespaceEnvNotFound)
        fmt.Printf("%T\n", unwrapped)
}                                                                                                                                                                                                                                                     
 $ go run err.go
<nil>
```

cmurphy (commented):
@JAORMX I think you're right that we want this to work either within a namespace or across the cluster, and that's currently accomplished by deploying the operator into the `security-profiles-operator` namespace while configuring it to either watch one other namespace (configured by `RESTRICT_TO_NAMESPACE`) or to watch the whole cluster. If I am reading this PR correctly this isn't removing the ability to watch the whole cluster, it's just changing how the deployment's namespace is looked up. So I think this is a step in the right direction, while the multi-namespaced cache builder could be an additional enhancement to the separate `RESTRICT_TO_NAMESPACE` feature?

internal/pkg/controllers/spod/configmap_controller.go
naveensrinivasan: I did that because of `wrapcheck` error from golang-lint

image

naveensrinivasan: Probably revisit whether `wrapcheck` makes sense.

cmurphy: That check is asking for the opposite, use `errors.Wrap` to add more context to the error

naveensrinivasan: @cmurphy Thanks, my ignorance. I will address this.

JAORMX (approved):
Thanks for the responses folks! Let's then cover the extra use-cases I mentioned in my comment in subsequent PRs.

internal/pkg/config/config.go
```diff
@@ -47,15 +52,25 @@ const (
...
-	// operatorNs, err := k8sutil.GetOperatorNamespace()
-	// if err != nil {
-	// 	return "security-profiles-operator"
-	// }
-	// return operatorNs
-	return "security-profiles-operator"
+func GetOperatorNamespace() (string, error) {
+	// This is MY_POD_NAMESPACE should have been set by the downward API to identify
+	// the namespace which this controller is running from
+	MyPodNamespace := GetEnvDefault("MY_POD_NAMESPACE", "")
```
saschagrunert, line 61: How about:
```suggestion
	MyPodNamespace := util.EnvDefault("MY_POD_NAMESPACE", "")
```

And importing util from `k8s.io/release/pkg/util`

internal/pkg/controllers/spod/configmap_controller.go
```diff
@@ -61,9 +61,15 @@ func (r *ReconcileSPOd) Reconcile(request reconcile.Request) (reconcile.Result,
 		}
 		return reconcile.Result{}, fmt.Errorf("error getting spod configuration: %w", err)
 	}
+
+	namespace, err := config.GetOperatorNamespace()
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf(" error getting operator namespace: %w", err)
```
saschagrunert, line 67:
```suggestion
		return reconcile.Result{}, errors.Wrap(err, "getting operator namespace")
```

internal/pkg/controllers/spod/setup.go
```diff
@@ -43,8 +44,15 @@ type DaemonTunables struct {
 	WatchNamespace      string
 }
 
+var operatorNamespace string
+
 // Setup adds a controller that reconciles the SPOd DaemonSet.
 func Setup(ctx context.Context, mgr ctrl.Manager, dt *DaemonTunables, l logr.Logger) error {
+	var err error
+	if operatorNamespace, err = config.GetOperatorNamespace(); err != nil {
+		return fmt.Errorf("error in getting operator namespace: %w", err)
```
saschagrunert, line 53:
```suggestion
		return errors.Wrap(err, "getting operator namespace")
```

internal/pkg/controllers/spod/setup.go
```diff
@@ -43,8 +44,15 @@ type DaemonTunables struct {
 	WatchNamespace      string
 }
 
+var operatorNamespace string
+
 // Setup adds a controller that reconciles the SPOd DaemonSet.
 func Setup(ctx context.Context, mgr ctrl.Manager, dt *DaemonTunables, l logr.Logger) error {
+	var err error
```
saschagrunert, line 51:
```suggestion
func Setup(ctx context.Context, mgr ctrl.Manager, dt *DaemonTunables, l logr.Logger) (err error) {
```

internal/pkg/controllers/spod/setup.go
```diff
@@ -43,8 +44,15 @@ type DaemonTunables struct {
 	WatchNamespace      string
 }
 
+var operatorNamespace string
```
saschagrunert, line 47: Do we want to be on the safe side of life and protect this global var with a mutex?

saschagrunert:
@naveensrinivasan can you please rebase to get the e2e fix in?

cmurphy:
The latest rebase re-added the `webhook-server-cert` to the operator deployment https://github.com/kubernetes-sigs/security-profiles-operator/pull/228/files#diff-ee1324908347cf8841925458a9937fb58cd6593d4002af28b74a0071bb12dbf4R52 which had been removed in https://github.com/kubernetes-sigs/security-profiles-operator/pull/237 to try to avoid the flakiness that cert-manager was introducing. You'll need to fix the commit to contain only your new code and not re-introduce removed code.

naveensrinivasan:
> The latest rebase re-added the `webhook-server-cert` to the operator deployment https://github.com/kubernetes-sigs/security-profiles-operator/pull/228/files#diff-ee1324908347cf8841925458a9937fb58cd6593d4002af28b74a0071bb12dbf4R52 which had been removed in #237 to try to avoid the flakiness that cert-manager was introducing. You'll need to fix the commit to contain only your new code and not re-introduce removed code.

Thanks! Updated the code.

JAORMX:
This can be closed, as https://github.com/kubernetes-sigs/security-profiles-operator/pull/349 merged

PR #233, opened 2021-01-25
make: Deploy cert-manager before deploying the operator in OpenShift

jhrozek:
There exists the 'deploy-openshift-dev' Makefile command that deploys the 
operator in an OpenShift cluster. But since the cert-manager dependency was
added, the target didn't work as OpenShift doesn't come with cert-manager
by default. This patch adds the dependency and waits for the cert-manager
pods to come up.

/kind bug

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -207,5 +207,5 @@ do-deploy-openshift-dev:
 	git checkout deploy/operator.yaml
 
 # Deploy for development into OpenShift
-.PHONY: dev-deploy-openshift
+.PHONY: deploy-openshift-dev
 deploy-openshift-dev: push-openshift-dev do-deploy-openshift-dev
--- a/Makefile
+++ b/Makefile
@@ -197,6 +197,9 @@ push-openshift-dev: set-openshift-image-params openshift-user image
 
 .PHONY: do-deploy-openshift-dev
 do-deploy-openshift-dev:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -197,6 +197,9 @@ push-openshift-dev: set-openshift-image-params openshift-user image
 
 .PHONY: do-deploy-openshift-dev
 do-deploy-openshift-dev:
+	@echo "Deploying cert-manager"
+	oc apply -f https://github.com/jetstack/cert-manager/releases/download/v1.1.0/cert-manager.yaml
```
jhrozek, line 201: This installs cert-manager from upstream, which is also what the e2e tests do, that's why I kept the same thing here. But it turns out that OCP ships cert-manager-operator in the openshift-marketplace:
```
 oc describe packagemanifests cert-manager-operator -nopenshift-marketplace
```
It might be possible to install that operator instead.

JAORMX: This is fine for development though. Ultimately, we could even use OCP's own cert-rotation controllers... and when we package this for OLM, the OLM has its own way of providing certs too. So, let's keep this as-is and change later if needed.

JAORMX (approved):
/lgtm

saschagrunert (approved):
/lgtm

Issue #221, opened 2021-01-22
selinux: kustomize templates should enable selinuxd by default


jhrozek:
#### What would you like to be added:
The kustomize templates should switch `EnableSelinux` to true by default when deploying on OpenShift.

#### Why is this needed:
To cover the set of defaults better tailored to OpenShift

PR #234, opened 2021-01-25
make: Set EnableSelinux=true through kustomize for OpenShift

jhrozek:
When deploying on OpenShift, we are typically interested in deploying
the SELinux controller as well. Since we already have a
OpenShift-specific Makefile target that uses a kustomize overlay,
let's just extend that overlay to change the default values of the
'config' ConfigMap.

Closes: #221

/kind cleanup

```release-note

```


JAORMX (approved):

saschagrunert (approved):
/lgtm

Issue #220, opened 2021-01-22
e2e tests are broken due to failing cert-manager certificate creation


saschagrunert:
The e2e tests are currently not able to deploy the operator, because cert-manager seems to fail to create the certificate in kind.

The operator deployment gets stuck in `ContainerCreating` state:

```
> k describe pod security-profiles-operator-844b669589-fxvvm
Name:           security-profiles-operator-844b669589-fxvvm
Namespace:      security-profiles-operator
Priority:       0
Node:           spo-e2e-1611320144-control-plane/172.18.0.2
Start Time:     Fri, 22 Jan 2021 14:06:49 +0100
Labels:         app=security-profiles-operator
                name=security-profiles-operator
                pod-template-hash=844b669589
Annotations:    seccomp.security.alpha.kubernetes.io/pod: runtime/default
Status:         Pending
IP:
IPs:            <none>
Controlled By:  ReplicaSet/security-profiles-operator-844b669589
Containers:
  security-profiles-operator:
    Container ID:
    Image:         security-profiles-operator:latest
    Image ID:
    Port:          9443/TCP
    Host Port:     0/TCP
    Args:
      manager
    State:          Waiting
      Reason:       ContainerCreating
    Ready:          False
    Restart Count:  0
    Environment:
      RELATED_IMAGE_OPERATOR:          security-profiles-operator:latest
      RELATED_IMAGE_NON_ROOT_ENABLER:  bash:5.0
    Mounts:
      /tmp/k8s-webhook-server/serving-certs from cert (ro)
      /var/run/secrets/kubernetes.io/serviceaccount from security-profiles-operator-token-p4wvt (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             False
  ContainersReady   False
  PodScheduled      True
Volumes:
  cert:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  webhook-server-cert
    Optional:    false
  security-profiles-operator-token-p4wvt:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  security-profiles-operator-token-p4wvt
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  node-role.kubernetes.io/master=
Tolerations:     node-role.kubernetes.io/control-plane:NoSchedule op=Exists
                 node-role.kubernetes.io/master:NoSchedule op=Exists
                 node.kubernetes.io/not-ready:NoExecute op=Exists
[...]

PR #237, opened 2021-01-26
Split out ProfileBinding webhook

cmurphy:
#### What type of PR is this?

/kind bug
/kind flake

The ProfileBinding CRD isn't required for the minimum functionality of
the operator, and it requires installing cert-manager additional
dependency which may not be desireable to all users. Furthermore,
cert-manager is causing intermittent problems in the e2e tests that need
to be worked out. This change splits the webhook into its own command
and its own deployment, and skips deploying and testing it in the e2e
tests. The webhook e2e test can still be run manually by setting the
environment variable E2E_RUN_EXPERIMENTAL=Y.

Relates to #220 

#### Does this PR have test?

N/A

```release-note
NONE
```


saschagrunert (approved):
/lgtm

Do you think it would make sense to make the webhook deployment configurable in a follow-up? We could deploy the webhook based on a config in the same way we deploy the operator daemonset, right? We probably have to spli-up the deployment in any case, like:

1. User deploys a configuration before the single all-in-one manifest
2. Then the user can deploy the operator which uses the configuration.

cmurphy:
> Do you think it would make sense to make the webhook deployment configurable in a follow-up?

Had to think about it for a bit but yes I think it would be good to follow this pattern. I filed https://github.com/kubernetes-sigs/security-profiles-operator/issues/270 to track that.

Issue #222, opened 2021-01-22
selinux: the selinuxpolicy controller should not trigger needless writes to the policy file


jhrozek:
#### What would you like to be added:
Reschedule the selinux controller reconcile loop until selinuxd's `/ready` endpoint returns true, indicating that selinuxd is ready to accept requests.

#### Why is this needed:
There was a race in selinuxd that caused the daemon to only start watching for selinuxpolicy files after the startup which is quite expensive and takes a long time. This was uncovered when I tried to remove the code in the reconciler that always writes the policy files on every reconcile loop. If we want to only write once in order to not trigger more inotify watches for selinuxd and make the policy state flap, we need to know when selinuxd is ready to accept requests which was implemented in https://github.com/JAORMX/selinuxd/pull/25

PR #239, opened 2021-01-27
selinux: Wait for the /ready endpoint before writing the policies, write policies only if they differ

jhrozek:
It was pointed out during the initial review of the selinuxd integration
that the selinuxpolicy controller performs too many needless writes on
each reconcile while it's waiting for selinuxd to do its job and install
a policy.

Just naively perform a single write uncovered a race in selinuxd where
the initial inotify messages might have gotten lost before selinuxd was
done processing the initial shipped policies. To mitigate that, selinuxd
has now a /ready handler that exposes whether it is ready to process
policies.

This patch avoids reconciling any policies until the /ready handler
replies True. Making use of this, a policy is only written if it doesn't
exist yet or if the contents differ.

/kind cleanup
/release-note-none

Closes: #222

```release-note

```


JAORMX (approved):
Yes! This is a lot nicer. We could move the logic of detecting if the policy is the same to selinuxd... but that could be done later. 

/lgtm

saschagrunert (approved):
/lgtm

PR #243, opened 2021-01-29
Feature - Speedup docker builds

naveensrinivasan:
#### What type of PR is this?
/kind cleanup

Decreased the docker builds duration by using cache

None

#### Does this PR have a test?
N/A

NONE

```release-note
 Decreased docker builds duration by using cache
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,9 @@ RUN adduser \
     --uid "${UID}" \
     "${USER}"
 
-ADD . /work
+COPY . /work
+
+FROM build as make
 RUN make
 
 FROM scratch
@@ -41,6 +43,6 @@ LABEL name="Security Profiles Operator" \
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=build /etc/passwd /etc/passwd
 COPY --from=build /etc/group /etc/group
-COPY --from=build /work/build/security-profiles-operator /
+COPY --from=make /work/build/security-profiles-operator /
 
 ENTRYPOINT ["/security-profiles-operator"]


saschagrunert (approved):
/lgtm

naveensrinivasan:
@saschagrunert  How do I know what version of docker is running in prow? Can it be upgraded?

naveensrinivasan:
Decided not to use the new `cache` syntax because prow does not support the latest version of Docker.

saschagrunert:
> @saschagrunert How do I know what version of docker is running in prow? Can it be upgraded?

The prow instance is running inside of the Google Cloud which means we cannot update docker that way.

Dockerfile
```diff
@@ -28,7 +28,12 @@ RUN adduser \
     --uid "${UID}" \
     "${USER}"
 
+# Speed up build by leveraging docker layer caching
+COPY go.mod go.sum ./
+RUN go mod download
```
saschagrunert, line 33: Do we really need this? We removed it because of the vendoring.

JAORMX: yeah, this shouldn't be needed any more.

Dockerfile
```diff
@@ -28,7 +28,9 @@ RUN adduser \
     --uid "${UID}" \
     "${USER}"
 
-ADD . /work
+COPY . /work
```
saschagrunert, line 31: What's the reason to use `COPY` instead of `ADD` here?

saschagrunert (commented):
So if I get it correctly, then we add an additional layer to speedup the build? If that is the case can you please outline that fact in the commit message/body? :upside_down_face:

Dockerfile
naveensrinivasan: hadolint had these warnings and that is the reason for the change https://github.com/hadolint/hadolint/wiki/DL3020

naveensrinivasan:
> So if I get it correctly, then we add an additional layer to speedup the build? If that is the case can you please outline that fact in the commit message/body? 🙃

I can update the message

saschagrunert (approved):
/lgtm

PR #245, opened 2021-01-31
Cleanup - .dockerignore files

naveensrinivasan:
#### What type of PR is this?
/kind cleanup

The docker build was getting started for most of the files that weren't being used in the container. This will prevent that.

None

#### Does this PR have test?
N/A

NONE

NONE

```release-note
NONE
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,12 @@
-build
+# Ignore everything
+*
+
+# Allow files and directories
+!/go.mod
+!/go.sum
+!/Makefile
+!/tools.go
+!/api/**
+!/cmd/**
+!/internal/**
+!/vendor/** 


.dockerignore
```diff
@@ -1 +1,11 @@
...
+*
+
+# Allow files and directories
+!/go.mod
+!/go.sum
+!/Makefile
+!/api/**
+!/cmd/**
+!/internal/**
+!/vendor/**
```
JAORMX, line 11: uhm... what about Dockerfile.ubi and its dockerignore?

or README and tools.go?

naveensrinivasan: Why do we need README, Dockerfile.ubi, and it's dockerignore? Changes to these files should not trigger docker build.

I understand on `tools.go`

Let me know what I am missing?

JAORMX:
oops! I don't know why I understood `.gitignore` instead of `.dockerignore`. That makes sense.

/lgtm

naveensrinivasan:
> oops! I don't know why I understood `.gitignore` instead of `.dockerignore`. That makes sense.
> 
> /lgtm

Thanks, I have updated it with `tools.go`

JAORMX (approved):

saschagrunert (approved):
/lgtm

PR #247, opened 2021-02-01
Add seccomp profile recording feature

saschagrunert:
#### What type of PR is this?

/kind feature

This is the initial implementation of the seccomp profile recording
feature. Right now it only works with CRI-O newer than v1.20.0 as well
as the deployed webhook. Another requirement is a working [OCI seccomp
BPF hook][0], whereas CRI-O has to be configured to watch the hooks'
directory. For example with a drop-in configuration like this:

```
> cat /etc/crio/crio.conf.d/03-hooks.conf
[crio.runtime]
hooks_dir = [
    "/path/to/oci/hook/dir",
]
```

The hook itself consists of a JSON which contains the execution
instructions as well as the path to the hook binary.

To be able to let the webhook do anything useful, we now deploy an
additional CRD called `ProfileRecording`. This CRD only contains a
selector, which will match any label for the application to be recorded.
If we create the workload, the webhook will take care of adding the
correct annotation to the pod. The hook will then record the seccomp
profile into a temporary JSON file. If the workload has been finished,
the new recorder controller will take the JSON file and reconcile it
into a `SeccompProfile` CRD object, which then gets synchronized on
disk.

The actual implementation is pretty rough and not well tested.
Documentation, test enhancements and stability improvements will follow
in later patches.

[0]: https://github.com/containers/oci-seccomp-bpf-hook

Refers to https://github.com/kubernetes-sigs/security-profiles-operator/issues/46

#### Does this PR have test?

Not right now.

demo

```release-note
Add seccomp profile recording support via the OCI seccomp BPF hook
```


saschagrunert:
Please take a look, this PR is already huge, but I have a few follow-ups in mind about:

* Documentation
* Unit testing
* Restructuring the project (the `profile` controller should be prefixed with `seccomp-`, this also affects other resources)
* Manually testing the corner cases

cc @cmurphy @pjbgf @hasheddan @JAORMX

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+	p, ok := obj.(*corev1.Pod)
+
+	if !ok {
+		return false
+	}
+
+	return p.Annotations[config.SeccompProfileRecordAnnotationKey] != ""
+}
+
+func (r *RecorderReconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
```
naveensrinivasan, line 136: How about adding events to the reconcile loop when there are failures with a warning? For example `unable to parse annotation`

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	if pod.Status.Phase == corev1.PodRunning || pod.Status.Phase == corev1.PodSucceeded {
+		if r.podsToWatch[req.NamespacedName.String()] != "" {
+			// We're tracking this pod already
+			return reconcile.Result{}, nil
+		}
+		outputFile, err := parseAnnotation(pod.Annotations[config.SeccompProfileRecordAnnotationKey])
+		if err != nil {
+			logger.Info("Unable to parse annoation. Ingoring", "error", err)
```
naveensrinivasan, line 162: Typo `Ingoring` `annoation`

naveensrinivasan, line 162: Comment on why we are ignoring would be helpful.

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+			return reconcile.Result{}, nil
+		}
+		if !strings.HasPrefix(outputFile, config.SeccompProfileRecordOutputPath) {
+			logger.Info("Ingoring Profile outside standard output path")
```
naveensrinivasan, line 166: Typo `Ingoring`

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+		logger.Info("RecorderReconciler", "Pod", name, "Profile", profile)
+	}
+	return reconcile.Result{}, nil
+}
+
+func (r *RecorderReconciler) collectProfile(ctx context.Context, name types.NamespacedName) error {
+	profilePath := r.podsToWatch[name.String()]
+	if profilePath == "" {
+		// ignoring this error for now
+		r.log.Info("Can not collect seccomp profile. Profile path empty")
```
naveensrinivasan, line 187: Typo `Can not`

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+		outputFile, err := parseAnnotation(pod.Annotations[config.SeccompProfileRecordAnnotationKey])
+		if err != nil {
+			logger.Info("Unable to parse annoation. Ingoring", "error", err)
+			return reconcile.Result{}, nil
+		}
+		if !strings.HasPrefix(outputFile, config.SeccompProfileRecordOutputPath) {
+			logger.Info("Ingoring Profile outside standard output path")
+		}
+		r.podsToWatch[req.NamespacedName.String()] = outputFile
+		r.record.Event(pod, event.Normal(reasonProfileRecording, "Recording seccomp profile"))
```
naveensrinivasan, line 169: `r.record` could be `nil`. How about checking for `nil`?

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	if pod.Status.Phase == corev1.PodSucceeded {
+		if err := r.collectProfile(ctx, req.NamespacedName); err != nil {
+			return reconcile.Result{}, nil
+		}
+	}
+
+	for name, profile := range r.podsToWatch {
+		logger.Info("RecorderReconciler", "Pod", name, "Profile", profile)
```
naveensrinivasan, line 178: Is there a specific reason to just `log` within the loop. If there is a need, probably a comment would help.

pjbgf (changes requested):
Great starting point @saschagrunert :+1: , just added a few nits.

A potential change for a follow-up PR could be to consider checking whether the workload is running with `allowPrivilegeEscalation` - assuming that it has as much of an impact in CRI-O as it has in Docker.

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: Fixed :+1:

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: Fixed :+1:

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: Fixed :+1:

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: No this was a debugging leftover. I removed that loop completely.

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: Yep, added an event for that special use case.

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: Added a comment above :+1:

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: In which path would that be the case? I thought we set the record via `event.NewAPIRecorder` in `Setup()` and then use it without any modification?

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+			nodeAddress = addr.Address
+			break
+		}
+	}
+
+	if nodeAddress == "" {
+		return errors.New("Unable to get node's internal Address")
+	}
+
+	reconiler := RecorderReconciler{
```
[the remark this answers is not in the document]
saschagrunert: Oops, fixed as suggested :+1:

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (r *RecorderReconciler) isPodOnLocalNode(obj runtime.Object) bool {
+	p, ok := obj.(*corev1.Pod)
+
+	if !ok {
+		return false
+	}
+
+	return p.Status.HostIP == r.nodeAddress
```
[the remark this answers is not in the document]
saschagrunert: Yes, I changed the possible node addresses to `[]string` and check for each now.

saschagrunert:
> A potential change for a follow-up PR could be to consider checking whether the workload is running with `allowPrivilegeEscalation` - assuming that it has as much of an impact in CRI-O as it has in Docker.

That's an interesting point. I'm not sure if `allowPrivilegeEscalation` has any implications because the recording process (the hook, which runs as root) is not coupled to the actual workload. It attaches the `raw_syscalls:sys_enter` tracepoint to the process and waits for it to exit. That's also one reason why we cannot cleanup the JSON files after the recording, because the operator runs with less permissions than the hook.

deploy/base/webhook/certificate.yaml
```diff
@@ -13,8 +13,8 @@ metadata:
   namespace: security-profiles-operator
 spec:
   dnsNames:
-  - profilebinding-webhook-service.security-profiles-operator.svc
-  - profilebinding-webhook-service.security-profiles-operator.svc.cluster.local
+  - profile-webhook-service.security-profiles-operator.svc
+  - profile-webhook-service.security-profiles-operator.svc.cluster.local
```
JAORMX, line 17: nit: what about just "webhook-service"? We'll probably add more webhook capabilities in the future. (At least I'm already working on one extra addition).

deploy/base/webhook/config.yaml
```diff
@@ -9,11 +9,40 @@ webhooks:
 - clientConfig:
     caBundle: Cg==
     service:
-      name: profilebinding-webhook-service
+      name: profile-webhook-service
```
JAORMX, line 12: what about just `webhook-service`?

cmd/security-profiles-operator/main.go
```diff
@@ -180,6 +183,9 @@ func runManager(ctx *cli.Context) error {
 	if err := seccompprofilev1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
 		return errors.Wrap(err, "add seccompprofile API to scheme")
 	}
+	if err := profilerecording1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
+		return errors.Wrap(err, "add profilerecording API to scheme")
+	}
```
JAORMX, line 188: Is this needed in the manager?

internal/pkg/controllers/profilerecorder/profilerecorder.go
```diff
@@ -0,0 +1,256 @@
...
+				return reconcile.Result{}, nil
+			}
+		} else {
+			// Returning an error means we will be requeued implicitly.
+			logger.Error(err, "Error reading pod")
+			return reconcile.Result{}, errors.Wrap(err, errGetPod)
+		}
+	}
+
+	if pod.Status.Phase == corev1.PodRunning || pod.Status.Phase == corev1.PodSucceeded {
```
JAORMX, line 162: if the pod already succeeded why would we start recording at this point? Wouldn't we already loose all the container's events then?

JAORMX (changes requested):
Looks good at first glance! Just to be clear, this doesn't address setting the bpf hook on the node, right? Do envision the operator setting that up at some point?

deploy/base/webhook/certificate.yaml
saschagrunert: Fixed as suggested :+1:

deploy/base/webhook/config.yaml
saschagrunert: Fixed as suggested :+1:

cmd/security-profiles-operator/main.go
saschagrunert: No I don't think so, removed it.

internal/pkg/controllers/profilerecorder/profilerecorder.go
saschagrunert: Ah yeah I think we can remove that code path, too.

saschagrunert:
> Looks good at first glance! Just to be clear, this doesn't address setting the bpf hook on the node, right? Do envision the operator setting that up at some point?

Yes, right now the hook is a hard dependency. I'm not convinced that we should take over that part, since it also means that we have to deploy a binary and a config on operator startup. The runtime (CRI-O in my case) has to be configured to look into the right directory for the hook.

JAORMX:
/lgtm

Any other issues can be fixed in further PRs. Thanks for this! I really dig how this is shaping up!

Issue #244, opened 2021-01-30
seccomp: audit logs enrichment


pjbgf:
#### What would you like to be added:

Enrich seccomp logs with Kubernetes data (i.e. `namespace`, `pod` and `container`) and potentially outcome (e.g. `blocked`/`allowed`).

#### Why is this needed:

In a cluster and at scale, the built-in audit logs for seccomp are hard to track back to specific pods/containers.
The existing logs:

`[252130.031583] audit: type=1326 audit(1611996299.149:466250): auid=4294967295 uid=0 gid=0 ses=4294967295 pid=615549 comm="sh" exe="/bin/busybox" sig=0 arch=c000003e syscall=1 compat=0 ip=0x7f61a81c5923 code=0x7ffc0000`

Initial suggested enriched logs format:
`audit(1611996299.149:466250) type=seccomp node=kube-worker1 pid=20923 ns=default pod=my-pod c=container1 exe=/init syscall=epoll_pwait`

saschagrunert:
Yeah this looks like a great enhancement. How would you implement it from a technical perspective? Would it work on every Linux machine?

pjbgf:
Yes, I got it already working for Ubuntu and CRI-O. Need to test it a bit better but I believe it would not be a problem to support other distros (CentOS and RedHat) and container runtimes.

One question I have is whether it is worth having that in-tree (including the generation of a new docker image), if so keen on creating that PR today.

@JAORMX does SELinux also generate audit logs on "complain-mode"?

JAORMX:
@pjbgf it does generate audit logs, both in "enforcing" and "permissive" mode (which I guess is similar to "complain-mode").

cmurphy:
/close

This was done in https://github.com/kubernetes-sigs/security-profiles-operator/pull/279 https://github.com/kubernetes-sigs/security-profiles-operator/pull/251 https://github.com/kubernetes-sigs/security-profiles-operator/pull/339

PR #251, opened 2021-02-01
Add log enrichment feature for seccomp

pjbgf:
#### What type of PR is this?
/kind feature

Enriches seccomp audit logs with Kubernetes information (namespace, pod and container names), allowing SPO users to easily spot what workloads are missing specific syscalls on their profiles.

To illustrate its use, I removed a few syscalls from the operator's profile and this is what I got:
image

Relates to #244 

#### Does this PR have test?
Not yet

- A new RBAC role was created to isolate log enrichment requirements as this may be split from the other components in a follow-up PR.
- The best approach I could find to map PID to Pods was using `HostPID` and reading the `cgroup` file. This approach is container runtime specific and currently only supports CRI-O.
- `HostPID` is only enabled when the log enricher is enabled
- Any suggestion for a better name to replace log enricher?

```release-note
Add seccomp audit log enrichment feature
```

### DIFF
--- a/deploy/base/config.yaml
+++ b/deploy/base/config.yaml
@@ -8,3 +8,4 @@ metadata:
     security-profiles-operator/config: ""
 data:
   EnableSelinux: "false"
+  EnableLogEnricher: "false"
--- a/deploy/base/operator.yaml
+++ b/deploy/base/operator.yaml
@@ -31,6 +31,8 @@ spec:
               value: non-root-enabler
             - name: RELATED_IMAGE_SELINUXD
               value: quay.io/jaosorior/selinuxd
+            - name: RELATED_IMAGE_LOG_ENRICHER
+              value: paulinhu/kube-audit-log-enricher
       tolerations:
         - key: "node-role.kubernetes.io/master"
           operator: "Exists"
--- a/internal/pkg/controllers/spod/bindata/spod.go
+++ b/internal/pkg/controllers/spod/bindata/spod.go
@@ -30,6 +30,7 @@ var (
 	truly                           = true
 	userRoot                  int64 = 0
 	userRootless              int64 = 2000
+	hostCharDev                     = v1.HostPathCharDev
 	hostPathDirectory               = v1.HostPathDirectory
 	hostPathDirectoryOrCreate       = v1.HostPathDirectoryOrCreate
 )
@@ -39,6 +40,7 @@ const (
 	SelinuxDropDirectory = "/etc/selinux.d"
 	SelinuxdSocketDir    = "/var/run/selinuxd"
 	SelinuxdSocketPath   = SelinuxdSocketDir + "/selinuxd.sock"
+	kmsgPath             = "/dev/kmsg"
 )
 
 var Manifest = &appsv1.DaemonSet{
@@ -301,6 +303,52 @@ var Manifest = &appsv1.DaemonSet{
 						//	},
 						// },
 					},
+					{
+						Name:            "log-enricher",
+						Image:           "paulinhu/kube-audit-log-enricher",
+						ImagePullPolicy: v1.PullAlways,
+						VolumeMounts: []v1.VolumeMount{
+							{
+								Name:      "host-devkmsg-volume",
+								MountPath: kmsgPath,
+								ReadOnly:  true,
+							},
+						},
+						SecurityContext: &v1.SecurityContext{
+							// /dev/kmsg is being used as kernel messages source in order to
+							// support the widest range of distros. However, access to
+							// devices requires the container to run as privileged.
+							Privileged:             &truly,
+							ReadOnlyRootFilesystem: &truly,
+							SELinuxOptions: &v1.SELinuxOptions{
+								// TODO(pjbgf): Use a more restricted selinux type
+								Type: "spc_t",
+							},
+						},
+						Resources: v1.ResourceRequirements{
+							Requests: v1.ResourceList{
+								v1.ResourceMemory:           resource.MustParse("64Mi"),
+								v1.ResourceCPU:              resource.MustParse("50m"),
+								v1.ResourceEphemeralStorage: resource.MustParse("10Mi"),
+							},
+							Limits: v1.ResourceList{
+								v1.ResourceMemory:           resource.MustParse("128Mi"),
+								v1.ResourceCPU:              resource.MustParse("150m"),
+								v1.ResourceEphemeralStorage: resource.MustParse("20Mi"),
+							},
+						},
+						Env: []v1.EnvVar{
+							{
+								Name: "NODE_NAME",
+								ValueFrom: &v1.EnvVarSource{
+									FieldRef: &v1.ObjectFieldSelector{
+										APIVersion: "v1",
+										FieldPath:  "spec.nodeName",
+									},
+								},
+							},
+						},
+					},
 				},
 				Volumes: []v1.Volume{
 					// /var/lib is used as symlinks cannot be created across
@@ -385,6 +433,15 @@ var Manifest = &appsv1.DaemonSet{
 							HostPath: &v1.HostPathVolumeSource{
 								Path: config.ProfileRecordingOutputPath,
 								Type: &hostPathDirectoryOrCreate,
+							},
+						},
+					},
+					{
+						Name: "host-devkmsg-volume",
+						VolumeSource: v1.VolumeSource{
+							HostPath: &v1.HostPathVolumeSource{
+								Path: kmsgPath,
+								Type: &hostCharDev,
 							},
 						},
 					},
--- a/internal/pkg/controllers/spod/configmap_controller.go
+++ b/internal/pkg/controllers/spod/configmap_controller.go
@@ -160,6 +160,17 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 			"--with-selinux=true")
 	}
 
+	enableLogEnricher, err := strconv.ParseBool(cfg.Data[config.SPOcEnableLogEnricher])
+	if err == nil && enableLogEnricher {
+		newSPOd.Spec.Template.Spec.Containers = append(
+			newSPOd.Spec.Template.Spec.Containers,
+			r.baseSPOd.Spec.Template.Spec.Containers[2])
+
+		// HostPID is only required for the log-enricher
+		// and is used to access cgroup files to map Process IDs to Pod IDs
+		newSPOd.Spec.Template.Spec.HostPID = true
+	}
+
 	for i := range templateSpec.Containers {
 		templateSpec.Containers[i].ImagePullPolicy = pullPolicy
 	}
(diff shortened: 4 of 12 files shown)


pjbgf:
I will update the deployment model document with a section for "optional components" I think this may be a good candidate for that given its privileges.

I could not find a lesser privileged approach for the PID->POD mapping, so had to enable `HostPID`, but really keen on hearing ideas on better way around it. @saschagrunert @cmurphy @JAORMX

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -375,6 +423,15 @@ var Manifest = &appsv1.DaemonSet{
 							},
 						},
 					},
+					{
+						Name: "host-varsyslog-volume",
+						VolumeSource: v1.VolumeSource{
+							HostPath: &v1.HostPathVolumeSource{
+								Path: "/var/log/syslog",
```
saschagrunert, line 430:
```suggestion
								Path: config.SyslogDropDirectory,
```

deploy/base/operator.yaml
```diff
@@ -33,6 +33,8 @@ spec:
               value: non-root-enabler
             - name: RELATED_IMAGE_SELINUXD
               value: quay.io/jaosorior/selinuxd
+            - name: RELATED_IMAGE_LOG_ENRICHER
+              value: paulinhu/kube-audit-log-enricher
```
saschagrunert, line 37: Wondering if we want to put the sources from the dependency project (https://github.com/pjbgf/kube-audit-log-enricher) into the operator. Any thoughts on that? :)

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -39,6 +40,7 @@ const (
 	SelinuxDropDirectory = "/etc/selinux.d"
 	SelinuxdSocketDir    = "/var/run/selinuxd"
 	SelinuxdSocketPath   = SelinuxdSocketDir + "/selinuxd.sock"
+	SyslogDropDirectory  = "/var/log/syslog"
```
saschagrunert, line 43: So it looks like that we have a direct dependency to `syslogd` (and/or auditd?). Not all distributions ship it per default and stick to journald. I think we can work with this limitation, I'm just wondering how to make it accessible to the broadest possible mass.

internal/pkg/controllers/spod/configmap_controller.go
```diff
@@ -110,6 +110,17 @@ func (r *ReconcileSPOd) getConfiguredSPOd(cfg *corev1.ConfigMap) *appsv1.DaemonS
...
+	enableLogEnricher, err := strconv.ParseBool(cfg.Data[config.SPOcEnableLogEnricher])
+	if err == nil && enableLogEnricher {
+		newSPOd.Spec.Template.Spec.Containers = append(
+			newSPOd.Spec.Template.Spec.Containers,
+			r.baseSPOd.Spec.Template.Spec.Containers[2])
+	}
+
+	// HostPID is only required for the log-enricher
+	// and is used to access cgroup files to map Process IDs to Pod IDs
+	newSPOd.Spec.Template.Spec.HostPID = (err == nil && enableLogEnricher)
```
saschagrunert, line 122: We could reduce the attack surface by putting it into a different daemonset, which probably would not make things easier to maintain.

saschagrunert (commented):
I like the idea of scraping the logs for useful information within the operator. :+1: Sure, the privileges of the operator daemonset have to be increased for this, which might be not possible avoid.

deploy/base/operator.yaml
JAORMX: Let's make this a topic for next project call. We need to make a similar decision for selinuxd.

internal/pkg/controllers/spod/bindata/spod.go
JAORMX: `/var/log/syslog` isn't avialable in CoreOS. IIRC in some distros, syslogd isn't enabled either but these logs come from a journald plugin that persists the log there. We can keep this for now, but at some point we need to make this configurable or pluggable. I guess we'd need a journald driver for this to work on other distros.

pjbgf: I will make a few tests using `/dev/kmsg` instead. But would be interesting to know its support across distros.

internal/pkg/controllers/spod/configmap_controller.go
pjbgf: Yep, I was a bit torn on that. I am inclined to keep it for the time being - after all it is an optional component like selinuxd that currently runs privileged. And later review the overall attack surface and privileges. 

If we strive to have accurate security profiles per container we should be able to limit damage/risk of such privileges.

deploy/base/operator.yaml
pjbgf: Just added it to the monthly meeting - but happy to discuss it Thursday if we have the time.

internal/pkg/controllers/spod/bindata/spod.go
pjbgf: Thank you, made the change considering the new target `/dev/kmsg`.

deploy/base/webhook/rbac.yaml
```diff
@@ -75,3 +75,30 @@ roleRef:
...
+  name: spo-log-enricher-binding
+  namespace: security-profiles-operator
+subjects:
+- kind: ServiceAccount
+  name: security-profiles-operator
+  namespace: security-profiles-operator
+roleRef:
+  kind: ClusterRole
+  name: spo-log-enricher-role
+  apiGroup: rbac.authorization.k8s.io
```
cmurphy, line 104: Should this actually be in `deploy/base/rbac.yaml`?

pjbgf:
I have made a few changes in order to broaden support by using `/dev/kmsg` instead of `/var/log/syslog`. Unfortunately, this version requires the container to run as `privileged` in order to have access to said device.

I propose we go ahead with this approach (plus fixing any other feedback) and then on the follow-up PRs:
- Bring dependency code into the project
- Improve testing
- Decrease privilege requirements
- Create security profiles to water tight the feature
- Add documentation

internal/pkg/controllers/spod/configmap_controller.go
cmurphy: I think this line could be moved up into the above `if` block and just set the value to `true`

cmurphy (commented):
This will be so helpful!

internal/pkg/controllers/spod/configmap_controller.go
pjbgf: @cmurphy agreed, it does make it clearer.

deploy/base/webhook/rbac.yaml
pjbgf: Thank you for spotting that, it is now fixed.

deploy/base/config.yaml
```diff
@@ -9,3 +9,4 @@ metadata:
 data:
   SPOdImagePullPolicy: Always
   EnableSelinux: "false"
+  EnableLogEnricher: "false"
```
saschagrunert, line 12: What is the reason that we have to quote those booleans? (I think I asked this before at the change above)
```suggestion
  EnableLogEnricher: false
```

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -388,6 +436,15 @@ var Manifest = &appsv1.DaemonSet{
 							},
 						},
 					},
+					{
+						Name: "host-devkmsg-volume",
+						VolumeSource: v1.VolumeSource{
+							HostPath: &v1.HostPathVolumeSource{
+								Path: "/dev/kmsg",
```
saschagrunert, line 443:
```suggestion
								Path: KmsgDropDirectory,
```

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -39,6 +40,7 @@ const (
 	SelinuxDropDirectory = "/etc/selinux.d"
 	SelinuxdSocketDir    = "/var/run/selinuxd"
 	SelinuxdSocketPath   = SelinuxdSocketDir + "/selinuxd.sock"
+	KmsgDropDirectory    = "/dev/kmsg"
```
saschagrunert, line 43: This var does not have to be public, right?

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -301,6 +303,52 @@ var Manifest = &appsv1.DaemonSet{
...
+								Name:      "host-devkmsg-volume",
+								MountPath: KmsgDropDirectory,
+								ReadOnly:  true,
+							},
+						},
+						SecurityContext: &v1.SecurityContext{
+							// /dev/kmsg is being used as kernel messages source in order to
+							// support the widest range of distros. However, access to
+							// devices requires the container to run as privileged.
+							Privileged:             &truly,
```
saschagrunert, line 321: I could not find much, so can you explain the "why"? Which kind of privileged do we need beside having the device available within the container?

saschagrunert (commented):
Genereally I like that approach and it seems to work on most (all?) Linux distributions. :+1:

deploy/base/config.yaml
JAORMX: ConfigMaps don't support it. Has to be `map[string]string` :(

internal/pkg/controllers/spod/bindata/spod.go
pjbgf: The problem is that (TMK) Kubernetes does not support things such as `--device` that some container run-times have which allows for access to specific devices without the need to run as `privileged`. Therefore, in order to support access to devices you need to run `privileged`.

Trying to simply map the volume into the container and run as root without `privileged` results in `Operation not permitted`, which is fair to be honest.

I already have a few ideas on how to circumvent this - like we did with the non-root enabler. It's coming on follow-up PRs. :)

internal/pkg/controllers/spod/bindata/spod.go
pjbgf: fixed it and renamed the variable to make it more generic so we can use as both volume path and drop path.

internal/pkg/controllers/spod/bindata/spod.go
saschagrunert: Awesome, thank you for the wrap-up!

internal/pkg/controllers/spod/bindata/spod.go
pjbgf: right on, just updated that too, PTAL

deploy/base/operator.yaml
```diff
@@ -31,6 +31,10 @@ spec:
...
+      nodeSelector:
+        node-role.kubernetes.io/master: ""
```
saschagrunert, line 37: Let's remove this selector, some k8s envs do not add this label at all.

saschagrunert (approved):
Just one nit, happy to move forward with this PR indipendently if we wanna move the sources in to this repository or not

saschagrunert (approved):
/lgtm

PR #252, opened 2021-02-01
Fix intermittent issue with E2E tests

pjbgf:
#### What type of PR is this?
/kind bug

The `libseccomp-static` package only exists from alpine `v3.13` onwards, but the initial `golang:1.15-alpine` tag used `v3.12` and later on updated to `v3.13`. The base image `golang:1.15.7-alpine` uses that alpine branch.

Here's my locally cached `golang:1.15-alpine` and the new image:
image

After removing `golang:1.15-alpine` locally and trying again the result is different:
image

Therefore it looks like when the e2e works successfuly is because it has a fresher version of the `golang:1.15-alpine` tag. 
Yet another reason to love mutable tags. :smile: 

None

#### Does this PR have test?

N/A

image
image

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.15-alpine AS build
+FROM golang:1.15.7-alpine AS build
 WORKDIR /work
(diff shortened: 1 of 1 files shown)


hasheddan (approved):
/lgtm

PR #254, opened 2021-02-02
Automatically detect if 'podman' should be used.

JAORMX:
#### What type of PR is this?

<!--
Add one of the following kinds:
/kind cleanup

This makes it so that if podman is available on a running system, it'll
set the `CONTAINER_RUNTIME` variable appropraitely in the Makefile.

The podman binary is explicitly checked as the `podman` package
contains already a `docker` shim, so in both cases (docker or podman)
there will be a `docker` executable available. If the `podman` binary
isn't available, it falls back to docker as a default.

This just makes life a little easier if you're using podman (e.g. you no longer
need to always set the `CONTAINER_RUNTIME` variable.

None

#### Does this PR have test?

N/A.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ else
   LDFLAGS := -s -w -extldflags "-static" $(LDVARS)
 endif
 
-export CONTAINER_RUNTIME ?= docker
+export CONTAINER_RUNTIME ?= $(if $(shell which podman),podman,docker)
 
 ifeq ($(CONTAINER_RUNTIME), podman)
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):
/lgtm

saschagrunert:
/retest
:thinking:

naveensrinivasan:
The tests are flaky

JAORMX:
This one is ready to go. It was an actual bug in the code (the makefile format I used initially introduced an accidental space before the word "docker") now that's fixed.

PR #260, opened 2021-02-03
Expand on seccomp example profiles

pjbgf:
#### What type of PR is this?

/kind documentation

- Create a seccomp profile that is safer for production workloads. It is slightly more restrictive than the docker default and also auto allows essential calls to reduce log noise. This can also be used as template for users to test their seccomp profiles in production before setting in the enforcer mode.
- Change `targetWorkload` from seccomp profile examples to `generic`. A point of discussion here would be to ship them in by default in the future.
- Add description to each profile to clarify their meaning and highlight unsafe profiles.

None

#### Does this PR have test?

N/A

```release-note
Add complain-mode seccomp profile that is safer to run in production workloads
```

### DIFF
--- a/examples/seccompprofile.yaml
+++ b/examples/seccompprofile.yaml
@@ -2,23 +2,101 @@
 apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
 kind: SeccompProfile
 metadata:
-  name: profile-block
+  name: profile-block-all
+  annotations:
+    description: "Blocks all syscalls."
 spec:
   defaultAction: "SCMP_ACT_ERRNO"
-  targetWorkload: "example-profiles"
+  targetWorkload: "generic"
 ---
 apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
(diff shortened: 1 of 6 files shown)


examples/seccompprofile.yaml
```diff
@@ -2,23 +2,101 @@
 apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
 kind: SeccompProfile
 metadata:
-  name: profile-block
+  name: profile-block-all
```
saschagrunert, line 5: Might be possible that we have to adapt the test-cases for this change:
https://github.com/kubernetes-sigs/security-profiles-operator/blob/5580b9f684c8dc267ba78de5647f9330f4c629af/test/tc_default_profiles_test.go#L32

pjbgf: Yep, went across the project and updated other references to it in documentation/tests. PTAL

saschagrunert (approved):
/lgtm

PR #262, opened 2021-02-03
Remove nodeSelector from deployment and webhook

saschagrunert:
#### What type of PR is this?

/kind bug

The `nodeSelector` value `node-role.kubernetes.io/master` seems to be
only set by nodes which are created via kubeadm (and others). If the
node does not have the label at all, then the operator will not be
scheduled there. This can be reproduced by the usage of the
`hack/local-up-cluster.sh` script within k/k.

I think we should be safe removing the selector since we already have
the tolerations in place.

None
#### Does this PR have test?
None

None

```release-note
None
```


JAORMX:
/lgtm

but sounds like a capability we should add to the configuration.

PR #264, opened 2021-02-03
Also listen to its own Namespace if RESTRICT_TO_NAMESPACE is given

JAORMX:
#### What type of PR is this?

/kind bug
/kind flake
/kind regression

This makes the manager Deployment always listen on its own namespace, even if
the RESTRICT_TO_NAMESPACE environment variable is given. This ensures that it's
able to listen to its own configuration in the expected namespace.

This was problematic in the namespace-operator, as the Deployment wasn't able
to listen to its own config.

```release-note
NONE
```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
/lgtm

PR #268, opened 2021-02-03
Allow e2e test to receive custom image

JAORMX:

This allows folks to use a pre-built image for the e2e tests. Currently
only works in OpenShift, but the environment variable is there in case
this would be useful for the `kind` e2e test.

```release-note
NONE
```


jhrozek (approved):
/lgtm
tested by calling make push-openshift-dev and then setting the image. Nothing was built or pushed.

saschagrunert (approved):
/lgtm

PR #279, opened 2021-02-05
Add log-enricher code into spo image

pjbgf:
#### What type of PR is this?

/kind feature

Removes the dependency of an external container image in order to provide the log enrichment functionality.

Fixes #251 

#### Does this PR have test?

Yes, but will be expanded on follow-up PRs.

- Initial PR bringing the log-enrichment code over into the repo, follow-up PRs will deal with asserting and expanding support (`CRI-O`, `containerd`, `docker`, etc)
- Improving general testing
- Documentation

```release-note
NONE
```


JAORMX (approved):

jhrozek:
It's nice to see the code being merged into SPO but on a first glance, `make verify` is having issues.

pjbgf:
> It's nice to see the code being merged into SPO but on a first glance, `make verify` is having issues.

@jhrozek I completely missed the fact I haven't done lint tests on my other repo. :man_facepalming: it should be good for review now.

internal/pkg/enricher/main.go
```diff
@@ -0,0 +1,167 @@
...
+
+func tailDevice(device string, msgs chan string) {
+	file, err := os.Open(device)
+	if err != nil {
+		return
+	}
+	defer file.Close()
+
+	// seek to the end of device
+	_, err = file.Seek(0, os.SEEK_END)
```
jhrozek, line 103: `os.SEEK_END` appears to be deprecated: https://golang.org/pkg/os/#pkg-constants

internal/pkg/enricher/types.go
```diff
@@ -0,0 +1,427 @@
...
+type containerInfo struct {
+	PodName       string
+	ContainerName string
+	Namespace     string
+	ContainerID   string
+}
+
+// List of x86-x64 syscalls
+// Source: https://raw.githubusercontent.com/torvalds/linux/master/arch/x86/entry/syscalls/syscall_64.tbl
+var systemCalls = map[int]string{
```
jhrozek, line 43: How portable is this? Since we're using systemCalls as a lookup and don't fail if the syscall cannot be looked up, if there was a syscall not found in this table we would have printed an empty string. Would it be better in that case to print the number as string instead?

jhrozek (commented):
Please see one nit and one question inline, otherwise looks good!

internal/pkg/enricher/main.go
pjbgf: Great point, I replaced that with `io.SeekEnd` instead.

internal/pkg/enricher/types.go
pjbgf: I changed the output to contain both values (`scid` for syscall id and `scname` for syscall name) - that way we have specific fields for each. As a follow-up PR we could expand support to other architectures when translating the syscall name.

saschagrunert (approved):
/lgtm

PR #283, opened 2021-02-09
Add basic Vagrant test job

saschagrunert:
#### What type of PR is this?

/kind design

This basic test stub will be later on used for VM based e2e testing.

None
#### Does this PR have test?

None
/hold for testing

```release-note
None
```

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,21 @@
+name: e2e
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+jobs:
+  test:
+    runs-on: macos-10.15
+    timeout-minutes: 30
+    env:
+      RUN: ./hack/ci/run.sh
+    steps:
+      - uses: actions/checkout@v2
+      - name: Boot Virtual Machine
+        run: |
+          ln -sf hack/ci/Vagrantfile .
+          vagrant up
+      - name: Show environment information
+        run: |
+          $RUN kubectl get nodes -o wide
--- a/hack/ci/Vagrantfile
+++ b/hack/ci/Vagrantfile
@@ -1,0 +1,47 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Vagrant box for testing
+Vagrant.configure("2") do |config|
+  config.vm.box = "fedora/33-cloud-base"
+  memory = 4096
+  cpus = 4
+
+  config.vm.provider :virtualbox do |v|
+    v.memory = memory
+    v.cpus = cpus
+  end
+
+  config.vm.provider :libvirt do |v|
+    v.memory = memory
+    v.cpus = cpus
+  end
+
+  config.vm.provision "install-dependencies", type: "shell", run: "once" do |sh|
+    sh.inline = <<~SHELL
+    whoami
+      set -euxo pipefail
+      dnf install -y \
+        conntrack \
+        container-selinux \
+        golang-go \
+        iptables \
+        make \
+        openssl
+    SHELL
+  end
+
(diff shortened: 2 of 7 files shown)


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,21 @@
+name: e2e
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+jobs:
+  test:
+    runs-on: macos-10.15
```
jhrozek, line 9: So the PR clearly shows that the action works fine, but out of curiosity, what is the reasoning behind running the job on a mac?

hack/ci/Vagrantfile
```diff
@@ -0,0 +1,25 @@
...
+  memory = 2048
+  cpus = 2
+  config.vm.provider :virtualbox do |v|
+    v.memory = memory
+    v.cpus = cpus
+  end
+  config.vm.provider :libvirt do |v|
+    v.memory = memory
+    v.cpus = cpus
+  end
```
jhrozek, line 16: TIL you can use two config.vm providers and it would work depending on what provider you use (I tested the vagrantfile locally on Fedora with libvirt), nice.

jhrozek (commented):
Seems to work fine, I just let two minor questions in the PR.

.github/workflows/e2e.yml
saschagrunert: This machine supports nested virtualization and makes vagrant work via virtualbox.

saschagrunert:
> Seems to work fine, I just let two minor questions in the PR.

I think I can include a basic CRI-O and kubernetes setup within this PR.

JAORMX (approved):

saschagrunert:
Bootstrapping the cluster via `hack/local-up-cluster.sh` already takes 18min, we may need to use some pre-built binaries, maybe them directly from [yum.k8s.io](https://yum.k8s.io).

.github/workflows/e2e.yml
jhrozek: ack, thanks

hack/ci/install-cri-o.sh
```diff
@@ -0,0 +1,71 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+cd "$(mktemp -d)"
+
```
JAORMX, line 19: What about doing:

```
dnf module enable cri-o:1.20
dnf install cri-o
```

instead of compiling cri-o?

saschagrunert: We download the static binary bundle from CRI-O which already contains all runtime dependencies. This is even faster than installing it via `dnf` :)

saschagrunert:
Cluster bootstrap takes now ~5minutes with downloading the binaries. The output of the e2e GitHub action job seems also fine:

```
2021-02-10T09:27:01.4993610Z NAME        STATUS   ROLES    AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                    KERNEL-VERSION           CONTAINER-RUNTIME
2021-02-10T09:27:01.5051080Z 127.0.0.1   Ready    <none>   44s   v1.20.2   127.0.0.1     <none>        Fedora 33 (Cloud Edition)   5.8.15-301.fc33.x86_64   cri-o://1.20.0
```

I think we now can fill the job with test logic.

/hold cancel

PTAL @pjbgf @JAORMX @jhrozek

PR #286, opened 2021-02-10
Run e2e test in vagrant-based GH action e2e test

JAORMX:
#### What type of PR is this?

/kind feature

This adds the e2e test run as part of the e2e GH action.

It introduces a "vanilla" kuberentes e2e cluster type which assumes that
a kubernetes cluster is already running and takes it into use.

Note that this also installs the oci hook for dynamically creating seccomp
profiles.

None

#### Does this PR have test?

N/A.

```release-note
NONE
```

Co-Authored-By: Sascha Grunert <mail@saschagrunert.de>

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -73,6 +73,10 @@ type openShifte2e struct {
 	skipPushImages  bool
 }
 
+type vanilla struct {
+	e2e
+}
+
 // We're unable to use parallel tests because of our usage of testify/suite.
 // See https://github.com/stretchr/testify/issues/187
 //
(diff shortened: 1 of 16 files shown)


test/suite_test.go
```diff
@@ -309,6 +325,29 @@ func (e *openShifte2e) execNodeOCP(node string, args ...string) string {
...
+	e.run(
+		containerRuntime, "pull", "quay.io/jaosorior/selinuxd",
+	)
+}
+
+func (e *vanilla) TearDownTest() {
+}
+
+func (e *vanilla) execNodeVanilla(node string, args ...string) string {
+	return e.run(containerRuntime, append([]string{"exec", node}, args...)...)
```
saschagrunert, line 348: Assuming we're already on the node (as `root`), we probably just have to call `bash`?

JAORMX: Done

saschagrunert (approved):
/lgtm

PR #295, opened 2021-02-11
Remove clusterRole and clusterRoleBinding overrides from namespaced-operator

JAORMX:
#### What type of PR is this?

/kind cleanup

The main target of this type of deployment is to have the SPO to only
liten for policies in specific namespaces. Having these overrides only
adds extra complexity, and the operator itself is able to function
without these.

Let's remove this extra complexity!

None

#### Does this PR have test?

This is already covered by the namespace set of tests

```release-note
The namespaced-operator deployment now relies on a ClusterRole and a ClusterRoleBinding instead of the previous Role And RoleBinding objects. It now more closely resembles the cluster-operator deployment.
```

### DIFF
--- a/deploy/namespace-operator.yaml
+++ b/deploy/namespace-operator.yaml
@@ -310,12 +310,12 @@ metadata:
   namespace: security-profiles-operator
 ---
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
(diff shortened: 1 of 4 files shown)


saschagrunert (approved):
/lgtm

PR #296, opened 2021-02-11
Add multi architecture image support

### LINKED ISSUE #282: Multi-architecture support
The idea is to build a multi-architecture image to support a common set of golang architectures (linux-only for now).

Question: For which kind of architectures do we see a benefit if we decide to implement that feature?

saschagrunert:
#### What type of PR is this?

/kind feature

We now build a multi architecture container image in GCB for amd64 and arm64.

Fixes #282 
#### Does this PR have test?
None

None

```release-note
Added multi-architecture support to the container image (amd64 and arm64 for now)
```

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -3,17 +3,22 @@
 timeout: 1200s
 options:
   substitution_option: ALLOW_LOOSE
+  machineType: N1_HIGHCPU_8
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - name: gcr.io/k8s-testimages/gcb-docker-gcloud:v20201130-750d12f
+    entrypoint: bash
+    env:
+      - TAG=$_GIT_TAG
+      - HOME=/root  # for docker buildx
     args:
-      - build
-      - --tag=gcr.io/$PROJECT_ID/security-profiles-operator:$_GIT_TAG
-      - --tag=gcr.io/$PROJECT_ID/security-profiles-operator:latest
-      - .
+    - -c
+    - |
+      gcloud auth configure-docker && \
+      make image-cross
 substitutions:
   # _GIT_TAG will be filled with a git-based tag for the image, of the form
   # vYYYYMMDD-hash, and can be used as a substitution
   _GIT_TAG: '12345'
 images:
-  - 'gcr.io/$PROJECT_ID/security-profiles-operator:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/security-profiles-operator:latest'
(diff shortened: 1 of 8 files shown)


saschagrunert:
Did a manual test:
```

```
Logs: https://console.cloud.google.com/cloud-build/builds/967cc1f7-2ce1-4034-ad05-49af87bd49e4?project=448637284062

We now push 3 images:
- gcr.io/k8s-staging-sp-operator/security-profiles-operator (manifest list)
- gcr.io/k8s-staging-sp-operator/security-profiles-operator-amd64 (only amd64)
- gcr.io/k8s-staging-sp-operator/security-profiles-operator-arm64 (only arm64)

All three images have 3 tags: `latest`, a git ref and the `VERSION`:

saschagrunert:
@hasheddan can you give this a look? :upside_down_face:

cloudbuild.yaml
```diff
@@ -3,17 +3,22 @@
...
+      make image-cross
 substitutions:
   # _GIT_TAG will be filled with a git-based tag for the image, of the form
   # vYYYYMMDD-hash, and can be used as a substitution
   _GIT_TAG: '12345'
 images:
-  - 'gcr.io/$PROJECT_ID/security-profiles-operator:$_GIT_TAG'
-  - 'gcr.io/$PROJECT_ID/security-profiles-operator:latest'
+  - 'gcr.io/$PROJECT_ID/security-profiles-operator-amd64:$_GIT_TAG'
+  - 'gcr.io/$PROJECT_ID/security-profiles-operator-amd64:latest'
```
hasheddan, line 24: @saschagrunert I didn't quite follow this change, could you provide some more context?

saschagrunert: This is just a check after the cloudbuild.yaml has run. It verifies that those images exist in the registry, nothing more.

hasheddan: Gotcha, thanks! That makes sense, wasn't aware it didn't work with manifests :)

hasheddan (approved):
/lgtm

jhrozek (approved):
Looks good (not adding lgtm as another review was requested)

cloudbuild.yaml
jonjohnsonjr: > It verifies that those images exist in the registry, nothing more.

Doesn't this actually _push_ those images from the docker daemon to the registry? https://cloud.google.com/build/docs/build-config#images

In this case, if you've already pushed them, I guess the push is a no-op?

saschagrunert: That’s not doing the actual push. As I tested it, it was just checking that the images are now available (aka already pushed).

PR #302, opened 2021-02-15
WIP: Move non-root-enabler into own image

saschagrunert:
#### What type of PR is this?

/kind feature

Target of this change is to be independent from external images as well as providing a toolchain which allows us creating multiple images from distinguishable binaries.

None
#### Does this PR have test?
None

Things to be done:

- [ ] Check if e2e testing and image substitution still works
- [ ] Change hack/image-cross.sh to push the non-root-enabler image

```release-note
Changed `non-root-enabler` to be independent from `docker.io/bash:5`,
because the image is now shipped with the operator release.
```

### DIFF
--- a/Dockerfile.base
+++ b/Dockerfile.base
@@ -1,0 +1,21 @@
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


JAORMX:
Why not integrate the non-root-enabler as a subcommand of the SPO main binary? it would make it a lot easier to deploy as the operator would only need to pull one image.

saschagrunert:
> Why not integrate the non-root-enabler as a subcommand of the SPO main binary? it would make it a lot easier and faster to deploy as the operator would only need to pull one image.

We had this topic in the early stages of the project and @pjbgf raised security concerns when coupling the code this tightly together. I'm happy to re-evaluate this discussion.

saschagrunert:
I'll come-up with a new PR.

PR #304, opened 2021-02-15
Move test related GitHub actions into one workflow

saschagrunert:
#### What type of PR is this?

/kind cleanup

This cleans the test related GitHub actions by moving them together and
reducing some overhead.

None
#### Does this PR have test?
None

None

```release-note
None
```


saschagrunert:
/assign @JAORMX 
After that one I'd like to pre-build the container image and pass it as artifact to the e2e job. This should speed-up things, too.

naveensrinivasan:
The codeql had an option to run on cron. The `cron` gave an option to have the code scanned even if there were no code merge to the master. 

The reason being upstream `codeql` could have some fixes that would not be caught if there were not merges to the repoisotry.

I understand it is fast-moving now. But that also has to be considered especially it being a security scan.

PR #306, opened 2021-02-16
Move non-root-enabler into operator image

saschagrunert:
#### What type of PR is this?

/kind feature

We now move the non-root-enabler logic into an own internal
`nonrootenabler` package and re-use it within the operator binary.

This allows mocked unit testing as well as moving away from another
container image (docker.io/bash) dependency.

None
#### Does this PR have test?
Yes

None

```release-note
Removed `docker.io/bash:5` container image dependency for non-root-enabler logic.
```


internal/pkg/config/config.go
```diff
@@ -25,6 +25,12 @@ const (
 	// OperatorName is the name when referring to the operator.
 	OperatorName = "security-profiles-operator"
 
+	// OperatorRoot is the root directory of the operator.
+	OperatorRoot = "/var/lib/security-profiles-operator"
+
+	// UserRootless is the user which runs the operator.
+	UserRootless = 65535
```
JAORMX, line 32: why not make this int64 already? that way you don't need to cast it later on

internal/pkg/controllers/spod/bindata/spod.go
```diff
@@ -67,36 +66,16 @@ var Manifest = &appsv1.DaemonSet{
...
-						if [ ! -L ` + config.ProfilesRootPath + ` ]; then
-							/bin/ln -s ` + operatorRoot + ` ` + config.ProfilesRootPath + `
-						fi
-
-						/bin/chown -R 65535:65535 ` + operatorRoot + `
-						cp -f -v /opt/seccomp-profiles/* ` + config.KubeletSeccompRootPath + `
-					`},
+						Name:            "non-root-enabler",
+						Args:            []string{"non-root-enabler"},
+						ImagePullPolicy: v1.PullNever,
```
JAORMX, line 71: The daemon container defaults to `PullAlways`. Let's make both containers default to the same pull policy instead.

internal/pkg/config/config.go
saschagrunert: `os.Chown` accepts only `int`: https://golang.org/pkg/os/#Chown

This is used within nonrootenabler.go, so I would have a need for casting there.

internal/pkg/controllers/spod/bindata/spod.go
saschagrunert: Sure, fixing as suggested :+1:

internal/pkg/config/config.go
JAORMX: gotcha. That's fine then.

saschagrunert:
Had to rebase once again. I think the main issue was the missing `Chmod` in case the operator root already existed. This should be fixed now, too.

PR #308, opened 2021-02-16
Fix kustomize make target and file download

saschagrunert:
#### What type of PR is this?

/kind cleanup

The `$(BUILD_DIR)/kustomize` did not really work because it always tried
to re-download the binary. We now fix that behavior by making the target
`PHONY` and downloading only if the file does not already locally
exists.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,12 +117,15 @@ go-mod: ## Cleanup and verify go modules
 		$(GO) mod tidy && \
 		$(GO) mod verify
 
+.PHONY: $(BUILD_DIR)/kustomize
 $(BUILD_DIR)/kustomize: $(BUILD_DIR)
-	export \
-		VERSION=4.0.1 \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -117,12 +117,15 @@ go-mod: ## Cleanup and verify go modules
 		$(GO) mod tidy && \
 		$(GO) mod verify
 
+.PHONY: $(BUILD_DIR)/kustomize
```
JAORMX, line 120: But why? The Makefile target should only trigger if `$(BUILD_DIR)/kustomize` does not exist in the filesystem.

saschagrunert: Hm, hm…I'm not able to reproduce the issue any more. I was not sure what caused the target to fail locally, I'll have to revisit once the issue comes up again.

/close

PR #311, opened 2021-02-17
Stick to x86_64 nix binaries for image builds

saschagrunert:
#### What type of PR is this?

/kind failing-test

We cross-compile via nix, so there is no need to use the `$(uname -m)`,
which will cause the `docker buildx` builds to download the arm binaries
instead of the x86_64 ones. This will result in cache poisoning on
cachix and will make the CI fail due to a timeout on the master branch.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,19 +17,19 @@ WORKDIR /work
 
 RUN apt-get update && apt-get install -y wget xz-utils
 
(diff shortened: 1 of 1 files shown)


JAORMX (commented):
/lgtm

PR #313, opened 2021-02-17
Move images used for e2e testing away from dockerhub

JAORMX:
#### What type of PR is this?

/kind bug

This moves our test images away from using dockerhub to using less restrictive
registries.

The hello-world image was replaced with one that we host in the
security-profile-operator's quay.io org called test-hello-world (note the
"test-" prefix which indicates that the image is used for testing).

The busybox image was replaced with a fedora-minimal image (coming from
Fedora's registry).

#### Does this PR have test?

N/A.

```release-note
NONE
```

### DIFF
--- a/test/tc_base_profiles_test.go
+++ b/test/tc_base_profiles_test.go
@@ -56,7 +56,7 @@ metadata:
   name: hello
 spec:
   containers:
-  - image: hello-world:linux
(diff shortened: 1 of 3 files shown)


saschagrunert (approved):
/lgtm

JAORMX:
/hold

let's replace the nginx image too

PR #325, opened 2021-02-19
Separate RBAC for daemon and deployment

### LINKED ISSUE #227: RBAC Permissions for controllers to be created by kubebuilder using markers
The RBAC permission for the controllers to be automated by code generation using kubbuilder markers. 

##### Example
```
// +kubebuilder:rbac:groups=batch.tutorial.kubebuilder.io,resources=cronjobs,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=batch.tutorial.kubebuilder.io,resources=cronjobs/status,verbs=get;update;patch
// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
// +kubebuilder:rbac:groups=batch,resources=jobs/status,verbs=get
```
https://book.kubebuilder.io/reference/markers/rbac.html
#### Why is this needed:
This avoids the manual updating of the permissions for k8s objects in the generated `yaml`. 
Also, it avoids over-provisioning systems with excess permission.

Here is an example
```
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  labels:
    app: security-profiles-operator
  name: security-profiles-operator
rules:
- apiGroups:
  - ""
  resources:
  - pods
  - services
  - services/finalizers
  - endpoints
  - persistentvolumeclaims
  - events
  - configmaps
  - configmaps/finalizers
  - secrets
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - apps
  resources:
  - deployments
  - daemonsets
  - replicasets
  - statefulsets
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - apps
  resourceNames:
  - selinux-operator
  resources:
  - deployments/finalizers
  verbs:
  - update
- apiGroups:
  - ""
  resources:
  - pods
  verbs:
  - get
- apiGroups:
  - apps
  resources:
  - replicasets
  - deployments
  verbs:
  - get
- apiGroups:
  - security.openshift.io
  resourceNames:
  - privileged
  resources:
  - securitycontextconstraints
  verbs:
  - use
---
```

The above role can delete `events` which not required.

JAORMX:
#### What type of PR is this?

/kind bug
/kind cleanup

This uses different roles and service accounts for the manager
deployment and the spod. These roles are generated through kubebuilder
annotations, and thus the mechanism and controller-gen calls to do so
were added to the Makefile.

The code was divided in such a way that it's easy to do these
controller-gen invocations and generate the needed rbac rules for the
two separate components. And so, everything that runs on the daemon is
now under the `daemon` directory. The same was done for the manager.

The webhooks directory didn't need to be touched as it was already separated
appropriately.

Fixes #227

#### Does this PR have test?

N/A.

```release-note
The security-profiles-operator now ships with separate service accounts for the daemon and webhook
```


JAORMX:
I must be missing some RBAC permission still and I just can't figure out which...

saschagrunert (approved):

jhrozek (approved):
/lgtm
my tests were also successfull and I couldn't reproduce the issue I raised on slack earlier anymore

internal/pkg/daemon/enricher/container.go
```diff
@@ -31,6 +31,11 @@ const (
 	crioPrefix = "cri-o://"
 )
 
+// NOTE(jaosorior): Should this actually be namespace-scoped?
```
pjbgf, line 34: Unfortunately, this can't be namespace-scoped otherwise we won't be able to translate PID to container/pod across the cluster. Unless we find a better approach to it.

pjbgf (approved):
really  good stuff, being able to trace back the rbac requirements is the way to go! :+1: 

/lgtm

saschagrunert (approved):
/lgtm

PR #329, opened 2021-02-25
spod: Harden the selinuxd container

### LINKED ISSUE #224: selinux: selinuxd container hardening
Not sure if this is an RFE or a bug, feel free to change the issue.

- resource requests and limits for the selinuxd container
- the selinuxd container should use read-only rootfs

#### Why is this needed:
To harden the selinuxd container. Note that the RO FS is dependant on mounting the database selinuxd uses in a RW location and the database location is not configurable at the moment, so this is dependent on an selinux RFE as well.

jhrozek:
spod: Set resource Requests and Limits for the selinuxd container

The limits are quite high, but better than none. I set the initial requests
and limits by just observing the container with kubectl top. I, rounding
them up, using as requests and then using double as limits.

spod: ReadyOnlyRootFs for the selinux container

Recent releases of selinuxd allow to specify a custom DB dir path. We can
take this into use by putting the DB into an EmptyDir where we alrady have
the selinuxd socket. This allows us to run the selinuxd container with a
read-only FS.

/kind cleanup
Fixes #224

```release-note
NONE

```

```release-note

```

### DIFF
--- a/internal/pkg/controllers/spod/bindata/spod.go
+++ b/internal/pkg/controllers/spod/bindata/spod.go
@@ -239,6 +239,7 @@ var Manifest = &appsv1.DaemonSet{
 							{
 								Name:      "selinux-drop-dir",
 								MountPath: SelinuxDropDirectory,
+								ReadOnly:  true,
 							},
 							{
 								Name:      "selinuxd-private-volume",
--- a/internal/pkg/controllers/spod/bindata/spod.go
+++ b/internal/pkg/controllers/spod/bindata/spod.go
@@ -37,8 +37,9 @@ var (
 
 const (
 	SelinuxDropDirectory = "/etc/selinux.d"
-	SelinuxdSocketDir    = "/var/run/selinuxd"
-	SelinuxdSocketPath   = SelinuxdSocketDir + "/selinuxd.sock"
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

JAORMX:
```
 internal/pkg/controllers/spod/bindata/spod.go:42:2: ST1003: const SelinuxdDbPath should be SelinuxdDBPath (stylecheck)
	SelinuxdDbPath       = SelinuxdPrivateDir + "/selinuxd.db" 
```

This one seems like a legit issue.

jhrozek:
On Sat, Feb 27, 2021 at 10:05:55AM -0800, Juan Osorio Robles wrote:
> ```
>  internal/pkg/controllers/spod/bindata/spod.go:42:2: ST1003: const SelinuxdDbPath should be SelinuxdDBPath (stylecheck)
> 	SelinuxdDbPath       = SelinuxdPrivateDir + "/selinuxd.db" 
> ```
> 
> This one seems like a legit issue.

Oops, fixed.

saschagrunert (approved):

jhrozek:
Hmm, the webhook test appears flaky...
/retest

saschagrunert (approved):
/lgtm

PR #336, opened 2021-03-03
Add CRD for SPOD deployment configuration

### LINKED ISSUE #327: Make security-profiles-operator daemonset configurable

A mechanism to make the creation of the security-profiles-operator daemonset configurable. At this point in time, the entire Daemonset is [hardcoded](https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/internal/pkg/controllers/spod/bindata/spod.go#L45).

#### Why is this needed:

As an infrastructure SRE there are a couple of scenarios, where I'd like to change the configuration of the daemonset:

- change container image source
- change nodeselectors and tolerations
- change resource requests and limits
- securitycontexts
- ... you name it

#### User story covered

As an infrastructure SRE, Kirsti wants to make sure that common profiles are available to all users, so folks can deploy applications securely. [↗️](https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/doc/user-stories.md#as-an-infrastructure-sre-kirsti-wants-to-make-sure-that-common-profiles-are-available-to-all-users-so-folks-can-deploy-applications-securely)

JAORMX:
#### What type of PR is this?

/kind feature
/kind design
/kind api-change

This adds a CRD called `SecurityProfilesOperatorDaemon` (short is spod),
which will both configure the SPOD daemon and reflect its status in the
cluster.

The Custom Resource looks very similar to the ConfigMap we used to have, but it
has the advantage that the fields are actually validated by the CRD itself, as
well as two status fields:

* State: A single word description of the status of the spod instance

* Conditions: A standard set of conditions which indicate observations about
  the deployment. These help us interact with the deployment using kubectl.

This also enables the deployment of as many SPO daemons as needed. Where,
an instance of a SPOD object matches an instance of the DaemonSet. This is
because the DaemonSet now is created with the same name as the SPOD
object. Thus, one can simply create a new DaemonSet by creating a SPOD
object with a different name.

Fixes #327

#### Does this PR have test?

N/A.

```release-note
The DaemonSet configuration is now handled by a Custom Resource called
SecurityProfilesOperatorDaemon.
```


saschagrunert (approved):
LGTM, just a nitty nit.

JAORMX:
@saschagrunert took in your suggestion. Thanks!

pjbgf (approved):
/lgtm

PR #338, opened 2021-03-03
Add leases RBAC to webhook

saschagrunert:
#### What type of PR is this?

/kind regression

This adds necessary leader election RBAC permissions to the webhooks.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/internal/pkg/webhooks/binding/binding.go
+++ b/internal/pkg/webhooks/binding/binding.go
@@ -80,7 +80,9 @@ func newContainerMap(spec *corev1.PodSpec) containerMap {
 
 // Leader election
 // nolint:lll
-// +kubebuilder:rbac:groups=core,namespace="security-profiles-operator",resources=configmaps;events,verbs=get;list;watch;create;update;patch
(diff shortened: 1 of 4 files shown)


internal/pkg/webhooks/binding/binding.go
```diff
@@ -81,6 +81,7 @@ func newContainerMap(spec *corev1.PodSpec) containerMap {
 // Leader election
 // nolint:lll
 // +kubebuilder:rbac:groups=core,namespace="security-profiles-operator",resources=configmaps;events,verbs=get;list;watch;create;update;patch
```
saschagrunert, line 83: Hm, the webhooks stills complains that it cannot create events in other namespaces, like `default`. Maybe we have to make this one more general?

JAORMX: We could simply add
```
 // +kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;update;patch
```

And that would allow the operator to issue events in all namespaces.

saschagrunert: Alright, I also removed the `events` from the resources here.

JAORMX:
Failed booting the VM

PR #339, opened 2021-03-03
Log enricher running unprivileged and extends supported audit line types

pjbgf:
#### What type of PR is this?

/kind feature

- Extends log enrichment support to `type=AVC` (SELinux) and `type=SECCOMP` audit lines 
- Removes dependency on `/dev/kmsg` device.

Relates to: https://github.com/kubernetes-sigs/security-profiles-operator/issues/244

#### Does this PR have test?

Yes. But E2E tests will be added on follow-up PRs.

Different distros may log seccomp audit lines differently, I have tested this based on log lines from Ubuntu (`type=1326`) and Fedora (`type=SECCOMP`).

The `/dev/kmsg` ended up being a dead-end as it cannot be executed in a least privilege mode. A follow-up PR will cater for the automatic creation of a symlink `/var/log/spo.log` pointing to the supported kernel log in the given distribution. In the interim the symlink needs to be manually created.

```release-note
Log enricher now supports SELinux log lines and runs unprivileged.
```


saschagrunert (approved):
/lgtm

saschagrunert:
> The `/dev/kmsg` ended up being a dead-end as it cannot be executed in a least privilege mode. A follow-up PR will cater for the automatic creation of a symlink `/var/log/spo.log` pointing to the supported kernel log in the given distribution. In the interim the symlink needs to be manually created.

Hey @pjbgf, do you think you could follow up on that?

PR #340, opened 2021-03-04
Add auto generate profiles use case

pjbgf:
#### What type of PR is this?

/kind documentation

Add new use case for auto generating security profiles for specific applications.
We need this to flash out the motivation around this feature.

None

#### Does this PR have test?

N/A

I tried to convey on the use case that the generated profiles are a starting point which will need to be manually reviewed and refined over time. 
For seccomp and AppArmor, it seems unrealistic to be able to automatically generate a profile that will consider *all* the execution path permutations for a given application, although the ability of running user provided E2E tests during the profile session may improve the results. 

```release-note
NONE
```


saschagrunert (approved):
/lgtm

JAORMX (approved):
Thanks!

PR #342, opened 2021-03-04
Handle Updates of SPOD object

JAORMX:
#### What type of PR is this?

/kind cleanup
/kind feature

This PR ensures that updates to the SPOD object are handled appropriately.
Thus, when the SPOD object contains new values, such as new images, flags, or
tolerations, the DS that the SPOD object owns will get updated as well.

None

#### Does this PR have test?

Yes

The "ready" condition now flicks with "updating" when the SPOD object is indeed
updating.

```release-note
Updates to the SecurityProfilesOperatorDaemon object are now reflected in the
daemonset.
```

### DIFF
--- a/api/spod/v1alpha1/spod_types.go
+++ b/api/spod/v1alpha1/spod_types.go
@@ -39,13 +39,15 @@ type SPODSpec struct {
 type SPODState string
 
 const (
-	// The policy is pending installation.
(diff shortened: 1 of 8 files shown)


saschagrunert (approved):
/lgtm

13942289419 (changes requested):
1

PR #344, opened 2021-03-04
Add multi-container recording support

saschagrunert:
#### What type of PR is this?

/kind feature

This adds support for multi container recording via the OCI seccomp BPF
hook. To not confuse users, we now require the (unreleased) version of
CRI-O v1.21.0. We always suffix the recording by the container name,
which allows the user to distinguish the profiles in a sane way.

None
#### Does this PR have test?
Yes

None

```release-note
None
```


JAORMX:
This is awesome!

PR #345, opened 2021-03-04
Restore the default tolerations

jhrozek:
#### What type of PR is this?
/kind bug

The spod_controller unconditionally copies toleratins from the spod
object. By default, the spod object had no tolerations, which resulted
in the DS scheduling pods only on worker nodes.

This patch includes the same tolerations as spod.go in the default spod
object.

None

#### Does this PR have test?
No

This is a minimal patch. I think another alternative might be "if no
tolerations are set in spod", use the default tolerations. I even suspect
this might be what the admin wants? Opinions welcome.

No

```release-note

```

### DIFF
--- a/deploy/base/config.yaml
+++ b/deploy/base/config.yaml
@@ -7,3 +7,13 @@ metadata:
 spec:
   enableSelinux: false
   enableLogEnricher: false
+  tolerations:
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoSchedule"
+    - key: "node-role.kubernetes.io/control-plane"
+      operator: "Exists"
+      effect: "NoSchedule"
+    - key: "node.kubernetes.io/not-ready"
+      operator: "Exists"
+      effect: "NoExecute"
--- a/deploy/namespace-operator.yaml
+++ b/deploy/namespace-operator.yaml
@@ -768,6 +768,16 @@ metadata:
 spec:
   enableLogEnricher: false
   enableSelinux: false
+  tolerations:
+  - effect: NoSchedule
+    key: node-role.kubernetes.io/master
+    operator: Exists
+  - effect: NoSchedule
+    key: node-role.kubernetes.io/control-plane
+    operator: Exists
+  - effect: NoExecute
+    key: node.kubernetes.io/not-ready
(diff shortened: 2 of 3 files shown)


saschagrunert (approved):
/ok-to-test
/lgtm

jhrozek:
On Thu, Mar 04, 2021 at 07:53:00AM -0800, Sascha Grunert wrote:
> @saschagrunert approved this pull request.
> 
> /ok-to-test
> /lgtm

Thank you for the lgtm, but what do you think about using default
tolerations if the CRD doesn't specify any?

I'm honestly just confused about whether we need the default tolerations
in the spod.go definition. I would prefer a single source of truth.

jhrozek:
`make verify` kicks my ass every single time..

fixed..

saschagrunert (approved):
/lgtm

saschagrunert:
I hope those tolerations still work with `hack/local-up-cluster.sh`.

jhrozek:
> I hope those tolerations still work with `hack/local-up-cluster.sh`.

I think yes; it's the same tolerations we had been using before the CRD change (I did went and verify from an older checkout). Wouldn't CI notice if it wasn't the case?

deploy/base/config.yaml
```diff
@@ -7,3 +7,13 @@ metadata:
...
+  tolerations:
+    - key: "node-role.kubernetes.io/master"
+      operator: "Exists"
+      effect: "NoSchedule"
+    - key: "node-role.kubernetes.io/control-plane"
+      operator: "Exists"
+      effect: "NoSchedule"
+    - key: "node.kubernetes.io/not-ready"
+      operator: "Exists"
+      effect: "NoExecute"
```
JAORMX, line 19: let's remove the `not-ready` toleration. The rest of them are fine. If the kubelet is stating that it's not ready to schedule workloads, let's respect that. The SPO is not considered a critical part of Kube (yet :smiley: )

JAORMX (changes requested):

JAORMX:
These tolerations should be fine, even if they don't exist in a vanilla kube instance; they just indicate that this pod is fine to schedule even if the reciprocating taints are enabled in the nodes.

PR #346, opened 2021-03-04
Re-enable SELinux by default on OCP

jhrozek:
#### What type of PR is this?
/kind bug

Prior to using CRDs, we were using kustomize to change the enableSelinux
key of the config CM to true. This no longer works since we switched to
CRD as it appears that kustomize doesn't handle CRD:
    https://github.com/kubernetes-sigs/kustomize/blob/1d4b3fa36c8918ad884c7aaa6a055fa2d0fa9bf9/examples/transformerconfigs/crd/README.md

Let's re-enable SELinux by default on OpenShift by just using sed.
Long-term we want to switch to auto-detect the capabilities anyway.

None

#### Does this PR have test?
N/A

```release-note
NONE

```


saschagrunert (approved):
/lgtm

PR #348, opened 2021-03-04
spod: Lower the selinuxd CPU requests and limits

jhrozek:
#### What type of PR is this?
/kind bug

The CPU requests and limits were way too high for the selinuxd container,
which was causing the DS pods to not be scheduled on clusters with low
CPU core nodes.

When I was setting the limits earlier, I was stupid and set even the CPU
/request/ based on what I saw in top and set a higher limit..well, that's
wrong because a process can function well with much less (unlike memory).

This time I did better experimenting with the different limits --
setting anything more than 1000m of CPU doesn't really make sense as a
limit, because rarely there is more than one process executing (always
selinuxdctl, sometimes there's an added setfilecontext helper).

With different CPU limits the selinuxd initialization took, until the
READY signal approximately:
 - with 300m CPU: 9 minutes
 - with 500m CPU: 6 minutes
 - with 1000m CPU or more: about 3 minutes
So the times scale linearly, but there's really no point in setting a
full core request (the process runs fine with much less) and no point in
setting a double-core limit (the process never uses them).

With memory, we unfortunately need to stick to the high values, I tested
setting request to 256Mi but got OOMKilled. So maybe 384 might work as
the lowest request, but I'm afraid it might not with complex policies.

None

#### Does this PR have test?
N/A

None

```release-note
None

```

### DIFF
--- a/internal/pkg/manager/spod/bindata/spod.go
+++ b/internal/pkg/manager/spod/bindata/spod.go
@@ -272,12 +272,12 @@ var Manifest = &appsv1.DaemonSet{
 						Resources: v1.ResourceRequirements{
 							Requests: v1.ResourceList{
 								v1.ResourceMemory:           resource.MustParse("512Mi"),
-								v1.ResourceCPU:              resource.MustParse("1000m"),
+								v1.ResourceCPU:              resource.MustParse("100m"),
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

saschagrunert (approved):
/lgtm

PR #349, opened 2021-03-05
Use Downward API to get operator namespace

JAORMX:
#### What type of PR is this?

/kind cleanup

This uses the downward API to fetch the operator's namespace.

None

#### Does this PR have test?

N/A.

This was based on #228

Credit to @naveensrinivasan for starting this work. Thanks!

```release-note
NONE
```

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -17,8 +17,11 @@ limitations under the License.
 package config
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
+
+	"k8s.io/release/pkg/util"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


JAORMX:
Supersedes #228

saschagrunert (approved):
/lgtm

internal/pkg/config/config.go
```diff
@@ -61,15 +64,34 @@ const (
...
+}
+
+// GetEnvDefault returns the value of the given environment variable or a
+// default value if the given environment variable is not set.
+func GetEnvDefault(variable, defaultVal string) string {
+	envVar, exists := os.LookupEnv(variable)
+	if !exists {
+		return defaultVal
+	}
+	return envVar
```
saschagrunert, line 96: I think this is not needed/used, right?

saschagrunert (commented):
/hold 
for my question

internal/pkg/config/config.go
JAORMX: Good find! It's not. Let me remove it.

saschagrunert:
`fatal error: concurrent map read and map write`

/retest

saschagrunert (approved):
/hold cancel
/lgtm

PR #350, opened 2021-03-05
Remove targetWorkload field from seccomp profiles

### LINKED ISSUE #347: Consider removing the `targetWorkload` field for seccomp profiles
We could think about removing the `targetWorkload` field from the seccomp profile CRD. Do we really need it? During development it was mostly confusing to have another sub-path in the whole local path.

Let's evaluate what we would break when removing it.

saschagrunert:
#### What type of PR is this?

/kind cleanup

Most of us felt confused by the field and I think since we do not
support configmaps any more we can remove it.

Fixes #347 
#### Does this PR have test?

None
None

```release-note
Removed `targetWorkload` field from seccomp profile CRD
```


saschagrunert:
This should work now PTAL @JAORMX @cmurphy @pjbgf

PR #361, opened 2021-03-10
Per-node statuses

### LINKED ISSUE #343: SPO scalable status
Per-node status of SPO's policies like seccomp or SELinux.

#### Why is this needed:
At the moment, the status of SPO profiles like seccomp or SELinux is per-profile. This presents several challenges:
 - it is impossible to reflect per-node status such as a profile failed to install on a single node or a node
   does not support the selected security profile
 - simply adding per-node attributes like a map might not scale, given very high number of nodes, the object might
   become too big for etcd's 1MB limit (see also https://github.com/kubernetes/kubernetes/issues/73324)
 - because the per-profile status is written to by several sources (each pod in a DaemonSet), the status might 
   appear as "flapping" as different pods reach different states at their own pace.

#### User story covered
As an application SRE, Rajesh needs to be able to see the state of the installed profile(s) and verify that it has indeed been installed on the system.

#### Proposed solution overview
To address all the problems above, we should add a new status-like object that would represent profile
node status. This object (node status) would be owned by the profile object so that the node status is
garbage collected when the profile object is.

Design goals:
 - it must be possible to easily list all node statuses of one profile status
 - the profile status should display the overall profile status. This would be handy e.g. for column headers

#### Implementation details
The current status of the seccomp profile looks like this:
```go
type SeccompProfileStatus struct {
	Path            string   `json:"path,omitempty"`
	Status          string   `json:"status,omitempty"`
	ActiveWorkloads []string `json:"activeWorkloads,omitempty"`
	// The path that should be provided to the `securityContext.seccompProfile.localhostProfile`
	// field of a Pod or container spec
	LocalhostProfile string `json:"localhostProfile,omitempty"`
}
```

The part if the profile status and the new `SeccompProfileNodeStatus` structure might look as follows:
```go
type SeccompProfileStatus struct {
    Path            string   `json:"path,omitempty"`
    Status          string   `json:"status,omitempty"`
    ActiveWorkloads []string `json:"activeWorkloads,omitempty"`
    // The path that should be provided to the `securityContext.seccompProfile.localhostProfile`
    // field of a Pod or container spec
    LocalhostProfile string `json:"localhostProfile,omitempty"`
}

type SeccompProfileNodeStatus struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`
	
	NodeName        string   `json:"nodeName"`
	Status          string   `json:"status,omitempty"`
}
```

Since the reconcile loop runs in a pod which is part of a DS, it would first check if a node status for this node
exists and if not, create it with some initial state (`Initializing`?). The node status name would be set to
`profile+nodeName` and the node status would be labeled with the profile name. For readability, the node name
is included in the node status as well.

On status update, the controller would write the status for the node. Additionally, the profile status includes
an aggregated status as well, which would display the lowest common status of all node statuses - for example, 
if all node statuses are `Active` except for one in `Initializing`, the overall status would still be `Initializing`,
only when all statuses transition to `Active` does the aggregated status change to `Active` as well.

The SELinux controller could be amended along the same lines.

#### Open questions
 - Several status attributes are inherently node-specific (`Path`, or even `ActiveWorkloads`), but it's
   unlikely that the admin would care about their value per node as opposed to knowing the status of a policy
   per node. Is that assumption correct?
 - Is it realistic that the `LocalhostProfile` or `Path` would be different on different nodes?
 - Events: Some events issues by the controllers are more specific to the node (e.g. failed to write a policy
   file to disk), some are better linked to the overall policy object (the profile contains a syntax error).
   We can:
      - split the events and issue those that pertain to the node for the node status object
      - keep issuing them for the profile itself, but try to add the node name as a key-value pair
        to the event messages. This is my preference as the events might be related to each other and
        it's easier to look at one place

jhrozek:
This PR introduced a new API secpolnodestatus that aims to address the following problems:
- it was impossible to reflect per-node status such as a profile failed to install on a single node or a node does not support the selected security profile
- simply adding per-node attributes like a map might not scale, given very high number of nodes, the object might become too big for etcd's 1MB limit
- because the per-profile status is written to by several sources (each pod in a DaemonSet), the status might appear as "flapping" as different pods reach different states at their own pace.

The secpolnodestatus is created, managed and deleted together with finalizers through an API in a new module. Instead of updating the global state directly, the DS pods now call the API update method that updates the node status and if needed, also the global status.

When a policy is deleted, the object is marked as terminating, when the policy payload is removed, the node status object along with its finalizer is deleted.  Finally, when all finalizers are gone, so is the global policy object.

#### What type of PR is this?
/kind feature

Implements per-node statuses for better scalability and better
granularity of error messages.

Fixes #343

#### Does this PR have test?
So far it piggy-backs on existing tests, new tests are still to be added.

```release-note
None
```


jhrozek:
Today's version should be code-complete and passed my manual testing. I haven't gotten around to trying e2e tests, though. But this patch should be at least ready for code-review!

jhrozek:
I think we'll need the seccomp conditions to wait more easily for the profile to be truly ready...

jhrozek:
Marking as WIP again since I haven't been able to finish debugging the selinux test and the seccomp conditions.

...to be continued

JAORMX:
@jhrozek seems this needs a rebase :smile:

jhrozek:
The seccomp delete test is still failing for me and I'm out of time to fix this today, but I think the issue is in the test adjustments, not the code itself. Pushing today's progress and I'll continue tomorrow, hopefully.

JAORMX:
@jhrozek any idea why the delete profiles test is failing?

jhrozek:
> @jhrozek any idea why the delete profiles test is failing?

Yes, I think the test was inherently racy even before, it was 1) not waiting for the profiles to be installed after creating them and 2) assuming if a profile is being terminated because it's in use and then the pod is deleted then the profile is immediately deleted. I fixed 1) with waiting on a condition and 2) with waiting on profile deletion in the background with a goroutine. Now the test works for me quite reliably locally, let's see how it gues in CI.

JAORMX (approved):
This looks good to me. Waiting for more folks to review as this is a considerably big change.

test/tc_delete_profiles_test.go
```diff
@@ -140,14 +146,37 @@ spec:
...
 		podCleanup := e.writeAndCreate(fmt.Sprintf(testCase.podManifest, namespace), "delete-pod*.yaml")
 		defer podCleanup()
 		e.waitFor("condition=ready", "pod", deletePodName)
 		e.logf("Ensuring profile cannot be deleted while pod is active")
 		e.kubectl("delete", "seccompprofile", deleteProfileName, "--wait=0")
-		output := e.kubectl("get", "seccompprofile", deleteProfileName)
-		e.Contains(output, "Terminating")
+
+		e.logf("Waiting for profile to be marked as terminating but not deleted")
+		// TODO(jhrozek): deleting manifests as Ready=False, reason=Deleting, can we wait in a nicer way?
```
JAORMX, line 157: you'd need a specific condition for deleting :confused:

jhrozek:
This is a rather large patchset, so feel free to take your time for review. 

I'm working on another small patch that should address another race in the seccomp controller, but it's not critical to get it in with this PR. I'm also thinking about a follow-up enhancement -- a separate controller that would reconcile the node statuses with the knowledge of the DS, but that' s a separate effort and I'll file a separate ticket.

api/secpolnodestatus/v1alpha1/secpolnodestatus_types.go
```diff
@@ -0,0 +1,97 @@
...
+const (
+	// Identifies the policy by name so that the admin can list all node statuses for a certain policy.
+	StatusToPolLabel = "spo.x-k8s.io/policy-name"
+	// Identifies the node on which the policy is installed so that the admin can list policies per node.
+	StatusToNodeLabel = "spo.x-k8s.io/node-name"
+	// Allows the admin to filter out node statuses with a certain state (e.g. show me all that failed).
+	StatusStateLabel = "spo.x-k8s.io/policy-state"
+	// The kind of policy so that the admin can filter only e.g. all selinux policy statuses.
+	StatusKindLabel = "spo.x-k8s.io/policy-kind"
+)
```
saschagrunert, line 52: Is there a reason to use "policy" here over "profile"?

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,208 @@
...
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/util"
+)
+
+type StatusClient struct {
+	pol             client.Object
+	nodeName        string
+	finalizerString string
+}
+
+func NewForPolicy(pol client.Object) (*StatusClient, error) {
```
saschagrunert, line 41: Same here: Do we wanna use "Profile" over "Policy"?

saschagrunert (approved):
Yay, I really like this feature! :partying_face: Just a naming/convention nit, otherwise LGTM

api/secpolnodestatus/v1alpha1/secpolnodestatus_types.go
jhrozek: not really, fixed

internal/pkg/nodestatus/nodestatus.go
jhrozek: yes, fixed, thank you

test/tc_delete_profiles_test.go
jhrozek: The `Deleting` condition is used when setting one, e.g. `status.SetConditions(rcommonv1.Deleting())`, I was just stumped that I couldn't find an equivalent of `GetConditions(Deleting)`

JAORMX (approved):

saschagrunert (approved):
/lgtm

Issue #360, opened 2021-03-10
Allow recording of deployments


saschagrunert:
Right now we support recording pods and containers into separate profiles. If we use higher level k8s objects like Deployments with replicas, then the profiles will overwrite each other depending on which replica gets written last.

#### Idea
Separate profiles by not only adding the container name to the profile, but also the pod name.

#### Issue
The webhook does not know the final Pod name when being executed. Either `pod.Name` is set (for example when using Pod objects) or the `pod.GenerateName` is set but not the `Name`. This means we have no chance with the current implementation to find out the final pod name.

PR #363, opened 2021-03-11
Enable replication recording

saschagrunert:
#### What type of PR is this?

/kind feature

The operator now supports recording replicated pods, for example when
using Deployments.

Fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/360
#### Does this PR have test?

Yes
None

```release-note
Add possibility to record seccomp profiles from replicas
```


internal/pkg/webhooks/recording/recording.go
```diff
@@ -97,15 +100,16 @@ func (p *podSeccompRecorder) Handle(
 	items := profileRecordings.Items
 
 	for i := range items {
-		if items[i].Spec.Kind != "SeccompProfile" {
+		item := items[i]
+		if item.Spec.Kind != "SeccompProfile" {
 			log.Info(fmt.Sprintf(
-				"recording kind %s not supported", items[i].Spec.Kind,
+				"recording kind %s not supported", item.Spec.Kind,
```
JAORMX, line 106: We should probably issue an event here too.

JAORMX (commented):
small comment, but I'd be fine merging otherwise. Let me know if you wanna cover that here or in a subsequent PR.

internal/pkg/webhooks/recording/recording.go
saschagrunert: Yeah I can follow-up on that. My idea was to enhance the unit testing of the webhooks in general so we can also add more debugging information with in that rush.

JAORMX: sounds good.

JAORMX:
@saschagrunert seems like you gotta retrigger the GH action e2e test

saschagrunert:
> @saschagrunert seems like you gotta retrigger the GH action e2e test

I'm wondering why not all 3 profiles gotta be found during the test. Let me debug this a bit…

saschagrunert:
We're close to working tests…

saschagrunert:
Found a bug related to the finalizer, which blocks the e2e test. 
/hold

JAORMX (approved):
/lgtm

PR #364, opened 2021-03-11
Use sync.Map for pods to watch

saschagrunert:
#### What type of PR is this?

/kind cleanup

There is no guarantee for non-concurrent access of the state, so we
should lock that map accordingly.

None
#### Does this PR have test?
None

None

```release-note
None
```


JAORMX (approved):
/lgtm

JAORMX:
We do run unit tests with `-race`; this might have gotten caught with better unit test coverage. Might be worth adding more tests.

saschagrunert:
This works now
/assign @JAORMX

JAORMX (commented):
/lgtm

PR #366, opened 2021-03-12
Give baseprofile tests time for reconciliation

saschagrunert:
#### What type of PR is this?

/kind failing-test

The current e2e tests are failing even within prow. This PR is an approach to fix this.

None
#### Does this PR have test?
Yes

None

```release-note
None
```

### DIFF
--- a/test/tc_base_profiles_test.go
+++ b/test/tc_base_profiles_test.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"strings"
+	"time"
 )
 
 func (e *e2e) testCaseBaseProfile([]string) {
@@ -57,7 +58,6 @@ spec:
   containers:
   - image: quay.io/security-profiles-operator/test-hello-world:latest
     name: hello
-    resources: {}
   securityContext:
     seccompProfile:
       type: Localhost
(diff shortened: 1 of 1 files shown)


test/tc_base_profiles_test.go
```diff
@@ -78,6 +78,15 @@ spec:
 	e.kubectl("create", "-f", helloProfileFile.Name())
 	defer e.kubectl("delete", "-f", helloProfileFile.Name())
 
+	e.logf("Waiting for profile to be reconciled")
+	for i := 0; i < 20; i++ {
+		output := e.kubectl("get", "sp", "hello")
+		if e.Contains(output, "hello") {
+			break
+		}
+		time.Sleep(time.Second)
```
JAORMX, line 87: would be nice to add conditions to the SeccompProfile object, that way we could just do `kubectl wait --for condition=ready`

saschagrunert: Hm, right now we save the profile to disk after the CRD got created. This means we have a time gap which we did not cover right now. We could add the status `ready` to CRDs, where all profiles have been reconciled successfully.

To do that we would need some kind of server instance which collects that information from each spod.

JAORMX: @saschagrunert that's exactly the work that @jhrozek is doing in https://github.com/kubernetes-sigs/security-profiles-operator/pull/361 I guess then all we need is the conditions there.

saschagrunert: Yeah, sounds reasonable. :+1:

saschagrunert:
This is now ready for review.

PR #370, opened 2021-03-15
Do not throw "Profile saved on disk" event if profile not modified

saschagrunert:
#### What type of PR is this?

/kind cleanup

If a seccomp profile has not been modified locally then we do not throw
the event any more. This avoids spamming the event on large clusters.

None
#### Does this PR have test?
None

None

```release-note
Throw "profile saved to disk" event only if a profile modification happened on the node.
```


jhrozek (approved):
/lgtm

jhrozek (approved):
/lgtm

PR #378, opened 2021-03-19
Add `release-note-none` label to dependabot updates

saschagrunert:
#### What type of PR is this?

/kind cleanup

This saves one more comment before merging those PRs.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,4 +4,6 @@ updates:
   directory: "/"
   schedule:
     interval: daily
+  labels:
+    - "release-note-none"
   open-pull-requests-limit: 10


jhrozek (approved):
/lgtm
the e2e test is surely unrelated, but I'm going to see why it failed anyway

saschagrunert:
> the e2e test is surely unrelated, but I'm going to see why it failed anyway

Not sure if that's the fix but let's give https://github.com/kubernetes-sigs/security-profiles-operator/pull/379 a try.

jhrozek:
> > the e2e test is surely unrelated, but I'm going to see why it failed anyway
> 
> Not sure if that's the fix but let's give #379 a try.

I'm not sure if that would help, because the profile itself is marked as installed so all the nodes /should/ (unless we still have a race) be reconciled at that point. But it's a good PR on its own. What I'm going to submit is to dump the node states to make sure we still don't have a race there. I also wonder if the files are always fsync-ed to the disk, gotta look at how does go's stdlib work..

saschagrunert:
> > > the e2e test is surely unrelated, but I'm going to see why it failed anyway
> > 
> > 
> > Not sure if that's the fix but let's give #379 a try.
> 
> I'm not sure if that would help, because the profile itself is marked as installed so all the nodes /should/ (unless we still have a race) be reconciled at that point. But it's a good PR on its own. What I'm going to submit is to dump the node states to make sure we still don't have a race there. I also wonder if the files are always fsync-ed to the disk, gotta look at how does go's stdlib work..

Sounds good, thank you!

PR #383, opened 2021-03-19
seccomp: Don't delete a profile in use

jhrozek:
#### What type of PR is this?
/kind bug

the finalizer we set using the pod reconciler was never taken into use
and thus the profiles could have been removed while still in use.

None

#### Does this PR have test?
yes

I haven't tested this at all because it's friday evening.

```release-note
A bug where a profile could have been deleted while still in use by pods was fixed
```

### DIFF
--- a/internal/pkg/daemon/seccompprofile/seccompprofile.go
+++ b/internal/pkg/daemon/seccompprofile/seccompprofile.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pkg/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"sigs.k8s.io/security-profiles-operator/api/seccompprofile/v1alpha1"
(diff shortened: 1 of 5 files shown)


saschagrunert (changes requested):
Code LGTM, can you please fix the lint issues that the CI is happy?

jhrozek:
On Mon, Mar 22, 2021 at 01:22:52AM -0700, Sascha Grunert wrote:
> @saschagrunert requested changes on this pull request.
> 
> Code LGTM, can you please fix the lint issues that the CI is happy?

I dream of the day when I'll be able to submit a PR without the verify
target kicking my ass.

this version splits the delete handler into its own function, because
the linter was complaining (rightfully so) that there is too mant if-s
in the delete branch.

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -374,6 +349,48 @@ func (r *Reconciler) reconcileSeccompProfile(
...
+	}
+
+	if err := r.setBothStatuses(ctx, nodeStatus, sp, setTerminatingStatus); err != nil {
+		l.Error(err, "cannot update SeccompProfile status")
+		r.record.Event(sp, event.Warning(reasonCannotUpdateProfile, err))
+		return reconcile.Result{}, errors.Wrap(err, "updating status for deleted SeccompProfile")
+	}
+
+	if controllerutil.ContainsFinalizer(sp, hasActivePodsFinalizerString) {
+		l.Info("Profile still in use, requeue")
```
cmurphy, line 371: nit: could mention in the message that this is a deletion event, e.g. `"cannot delete profile in use by pod, requeuing"`

cmurphy (commented):
I think this is worth a release note since it fixes a user-facing bug

internal/pkg/daemon/seccompprofile/seccompprofile.go
jhrozek: Done. Sorry it took so long to get back to your review; I was away for several weeks.

jhrozek:
> I think this is worth a release note since it fixes a user-facing bug

OK, I amended the PR description, if there's more I need to do, please let me know.

saschagrunert (approved):
/lgtm

PR #387, opened 2021-03-23
Install webhooks in main operator deployment

### LINKED ISSUE #270: Deploy ProfileBinding/ProfileRecording webhook as a configurable addon option
Currently the `Deployment` that handles the `ProfileBinding` and `ProfileRecording` webhook as well as the controllers for those CRDs is applied manually via the `webhook.yaml` manifest. Conversely, the DaemonSet controllers are configured and launched by the operator. We should follow the same pattern with the webhook configuration and eliminate the `webhook.yaml` manifest.

This relates to the discussion of the deployment model in #212. See also #237 where the new manifest file was added.

cmurphy:
#### What type of PR is this?

/kind feature

The original reason for splitting the webhook out from the main
deployment (b432ee4) was that the cert-manager deployment was
unreliable. The problem was fixed (ec596ffe) but the webhook remained
separate so that installing cert-manager would be optional.

This change brings the webhook back into the main deployment.
Cert-manager is now required. The functionality provided by the webhook
- profile bindings and profile recordings - are core features of the
operator and should not be optional. Since this operator is still in an
alpha state, relying on a third-party tool should be an acceptable
tradeoff for providing the full breadth of functionality with a
consistent, minimal deployment experience. We can reevaluate how to
manage webhook certs further down the road as this operator gains more
user adoption.

Fixes #270

#### Does this PR have test?

n/a

The github issue this fixes is about making the webhook deployment configurable in code, but I found it more sensible to just combine them. The ability to configure the DaemonSet makes sense since not all features (selinux, seccomp, apparmor) will be applicable to and supported by the host system of every node. But it makes less sense to do the same with a Deployment, since it's just a web app.

```release-note
The separate webhook deployment, which enabled the `ProfileBinding` and `ProfileRecording` resources, has now been merged into the main operator deployment manifest.
```


cmurphy:
Working on getting the e2e tests to run locally so this may or may not be green yet...

saschagrunert:
> Working on getting the e2e tests to run locally so this may or may not be green yet...

The tests are flaky right now, @jhrozek was investigating the issue AFAIK

saschagrunert (approved):
Code LGTM, let's see if the test issue was just a flake.

@JAORMX are you fine with having cert-manager now a hard requirement?

cmurphy:
/retest

`fatal error: runtime: cannot allocate memory` seems unrelated

saschagrunert (approved):

saschagrunert:
@cmurphy unfortunately this needs a rebase :upside_down_face:

saschagrunert (approved):
/lgtm

saschagrunert (approved):
/lgtm

PR #388, opened 2021-03-24
Do not include YEAR in generated boilerplate

saschagrunert:
#### What type of PR is this?

/kind documentation

We do not have to include the YEAR in the boilerplate header. This means
we can use a custom header for generated golang code, which should not
change next year.

None
#### Does this PR have test?

None

```release-note
None
```


hasheddan (approved):
/lgtm

cmurphy:
Shouldn't [this file](https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/hack/boilerplate/boilerplate.go.txt) be removed/changed too?

Why is the copyright year not required anymore?

saschagrunert:
> Shouldn't [this file](https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/hack/boilerplate/boilerplate.go.txt) be removed/changed too?

Nope, we still wanna preserve the YEAR for manually edited files.

> Why is the copyright year not required anymore?

Right now the verify boilerplate script skips generated files:
https://github.com/kubernetes/repo-infra/blob/5597dcdfb69ef4f4c7d994c2176034b9d6aa4721/hack/verify_boilerplate.py#L116

I think we should fix that somehow that the script only skips the YEAR instead of skipping the whole test.

Issue #373, opened 2021-03-16
Reconcile per-node statuses using a separate status reconciler


jhrozek:
#### What happened:
Even though the per-node statuses decrease the amount of races and make the statuses more dependable, it is only /mostly/ the case. As proposed in PR #361, the reconcile loop in the DS pods primarily managers the per-node statuses and if all per-node statuses agree, the same reconcile loop also updates the global policy status.

This has two problems:
 - we still have concurrent access to a shared resource and thus we get conflicts. This is irritating, but not unsolvable
 - the current algorithm to update the global policy status is simplistic, just list all the other per-node statuses and if they all are the same, update the global status. But the problem occurs if the other per-node statuses /are not created yet at all/. This is especially a problem for the seccomp reconcile loop that is very fast and just dumps a file onto the filesystem.

An example of the race might be:
 - two nodes A and B, A is very fast and managers to reconcile a seccomp profile all the way to it being installed before the reconciler on B even picks up the reconcile loop
 - A goes to installing, the pod on A does a list and finds only node-status A, thinks all nodes are 'consistent' and sets the global status as installed
 - later on B picks up the reconciliation and installs the policy, but before that a pod can theoretically be deployed on B that tries to reference that policy

#### What you expected to happen:
Race-free SPO statuses. I propose we add a new controller that only synchronizes the statuses. It could reconcile on changes to nodeStatus, list all the pods in a DS that manages the policy and only if all pods in the DS actually have a state, act on the state change.

This might have the following benefits:
 - there is exactly one place that writes to the node statuses, which means fewer races
 - a separate reconcile loop feels like a better way to observe a status of a DS than a pod running within the DS itself
 - the DS pods would no longer need write access to the policy object statuses, only to the node statuses
 - the code could potentially be better shared between policy controllers

#### How to reproduce it (as minimally and precisely as possible):
Not very often with the node statuses, but occasionally reproduces in e2e tests with seccomp.

#### Anything else we need to know?:
This is not a fully fleshed out idea, comments or alternative solutions welcome.

jhrozek:
In case I need to disappear this week, I'm pushing my progress here: https://github.com/jhrozek/security-profiles-operator/tree/status_ctrl

JAORMX:
/assign

Taking over this work as @jhrozek is not available for some days.

PR #389, opened 2021-03-24
Introduce status controller

JAORMX:
#### What type of PR is this?

/kind bug
/kind cleanup
/kind flake

This introduces a status controller which aggregates the node statuses generated by the DaemonSet
instances. This addresses the issue of the instances trying to update a single status for the policies
which resulted in races.

Fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/373

#### Does this PR have test?

N/A

```release-note
A new node status controller now runs on the main operator Deployment.
To standardize on a common status model, the SelinuxPolicy `state` was renamed to `status`.
The controller manager now listens on the same namespaces as the DaemonSet does. And thus requires more RBAC permissions.
The SecurityProfilesOperatorDaemon Custom Resource is now Namespaced and not Cluster scoped.
```


JAORMX:
I'm just pushing this as a PR so it's easier to view the changes this introduces, from here work on completing this feature.

cmurphy:
Tried this out locally and there's still a race in the e2e tests:

```
time="2021-03-24T11:13:48-07:00" level=info msg="+ /usr/bin/kubectl -n default get seccompprofile profile-allow-unsafe -o json"
{
    "apiVersion": "security-profiles-operator.x-k8s.io/v1alpha1",
    "kind": "SeccompProfile",
    "metadata": {
        "annotations": {
            "description": "UNSAFE: Allows all syscalls. Similar to running as unconfined as it provides no enforcement."
        },
        "creationTimestamp": "2021-03-24T18:13:45Z",
        "finalizers": [
            "spo-e2e-1616609345-worker-delete",
            "spo-e2e-1616609345-control-plane-delete",
            "spo-e2e-1616609345-worker2-delete"
        ],
        "generation": 1,
        "managedFields": [
            {
                "apiVersion": "security-profiles-operator.x-k8s.io/v1alpha1",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:metadata": {
                        "f:annotations": {
                            ".": {},
                            "f:description": {}
                        }
                    },
                    "f:spec": {
                        ".": {},
                        "f:defaultAction": {}
                    }
                },
                "manager": "kubectl-create",
                "operation": "Update",
                "time": "2021-03-24T18:13:45Z"
            },
            {
                "apiVersion": "security-profiles-operator.x-k8s.io/v1alpha1",
                "fieldsType": "FieldsV1",
                "fieldsV1": {
                    "f:metadata": {
                        "f:finalizers": {
                            ".": {},
                            "v:\"spo-e2e-1616609345-control-plane-delete\"": {},
                            "v:\"spo-e2e-1616609345-worker-delete\"": {},
                            "v:\"spo-e2e-1616609345-worker2-delete\"": {}
                        }
                    },
                    "f:status": {
                        ".": {},
                        "f:conditions": {},
                        "f:localhostProfile": {},
                        "f:path": {},
                        "f:status": {}
                    }
                },
                "manager": "security-profiles-operator",
                "operation": "Update",
                "time": "2021-03-24T18:13:47Z"
            }
        ],
        "name": "profile-allow-unsafe",
        "namespace": "default",
        "resourceVersion": "1204",
        "uid": "d389e75a-1dd6-4f56-a23c-9cb7626acde3"
    },
    "spec": {
        "defaultAction": "SCMP_ACT_ALLOW"
    },
    "status": {
        "conditions": [
            {
                "lastTransitionTime": "2021-03-24T18:13:46Z",
                "reason": "Available",
                "status": "True",
                "type": "Ready"
            }
        ],
        "localhostProfile": "operator/default/profile-allow-unsafe.json",
        "path": "/var/lib/kubelet/seccomp/operator/default/profile-allow-unsafe.json",
        "status": "Pending"
    }
}
I0324 11:13:48.298778 1097838 suite_test.go:445]  "msg"="Verifying profile-allow-unsafe profile on node spo-e2e-1616609345-worker"  
time="2021-03-24T11:13:48-07:00" level=info msg="+ /usr/bin/docker exec spo-e2e-1616609345-worker cat /var/lib/kubelet/seccomp/operator/default/profile-allow-unsafe.json"
cat: /var/lib/kubelet/seccomp/operator/default/profile-allow-unsafe.json: No such file or directory
```

JAORMX:
@cmurphy this is not complete. I just posted it for the diff. As it is now, this still updates the status in the daemonset. I have a local branch where I moved that to the new controller but its still not ready. I'll keep at it tomorrow.

JAORMX:
Still testing this

jhrozek:
Shit, sorry, I wanted to push the changes I did *before* I left to my private branch and didn't realize you guys already took over (which is good) so the change overwrote the PR. Sorry. I'll go back to being away now.

JAORMX:
@jhrozek I'm currently working on this branch

JAORMX:
> Shit, sorry, I wanted to push the changes I did _before_ I left to my private branch and didn't realize you guys already took over (which is good) so the change overwrote the PR. Sorry. I'll go back to being away now.

No worries! Greetings to the family! :smile: talk to you when you're back

cmurphy:
The profilebinding test is failing because this change is causing the `SeccompProfile` creation to not be atomic with setting its status, and the profilebinding webhook uses `SeccompProfile.Status.LocalhostProfile` to inject the profile path into the Pod's `SecurityContext`:

https://github.com/kubernetes-sigs/security-profiles-operator/blob/66c5df92f2fa69ff8d4567b4235558a4485f31fe/internal/pkg/webhooks/binding/binding.go#L170

and it ends up injecting an empty string:

```
$ kubectl get pod hello -o yaml | grep -A 3 '^ *securityContext:$'
    securityContext:
      seccompProfile:
        localhostProfile: ""
        type: Localhost
```

That later cause the `PodReconciler` not to find the pod in its index and so it doesn't add it to its `ActiveWorkloads` list which is how the test failure surfaces. But that's less of a problem than the actual profile binding not working.

So for this to work either we'll have to move that `LocalhostProfile` metadata to somewhere out of status (not a bad idea, status is kind of a funny place to put it), or have the webhook do some kind of concurrency-safe check to ensure the SeccompProfile Status isn't out of sync with the Spec.

JAORMX:
/test pull-security-profiles-operator-test-e2e

Failed with:
```
 fatal error: runtime: cannot allocate memory 
```

cmurphy (changes requested):
I did an initial pass, left a comment in the tests that will help the e2e stability. Will need to take some more time to understand all the changes. All this cross-controller machinery really increased the complexity of the operator, going back to when we added the profile binding controller, which is a bit of a shame imo.

internal/pkg/manager/workloadannotator/workloadannotator.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	if !strings.HasPrefix(*sp.LocalhostProfile, "operator/") {
+		return false
+	}
+	if !strings.HasSuffix(*sp.LocalhostProfile, ".json") {
+		return false
+	}
+	const pathParts = 3
+	return len(strings.Split(*sp.LocalhostProfile, "/")) == pathParts
+}
```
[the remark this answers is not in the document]
saschagrunert: Hm, adding a new bool field to operator-tracked profiles would be an alternative.

internal/pkg/webhooks/binding/binding.go
```diff
@@ -35,11 +36,14 @@ import (
 
 	profilebindingv1alpha1 "sigs.k8s.io/security-profiles-operator/api/profilebinding/v1alpha1"
 	seccompprofilev1alpha1 "sigs.k8s.io/security-profiles-operator/api/seccompprofile/v1alpha1"
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/util"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/webhooks/utils"
 )
 
 const finalizer = "active-workload-lock"
 
+var ErrProfWithoutStatus = errors.New("Profile hasn't been initialized with status")
```
saschagrunert, line 45:
```suggestion
var ErrProfWithoutStatus = errors.New("profile hasn't been initialized with status")
```

internal/pkg/util/reconcileutils_test.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
saschagrunert, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

JAORMX:
> I did an initial pass, left a comment in the tests that will help the e2e stability. Will need to take some more time to understand all the changes. All this cross-controller machinery really increased the complexity of the operator, going back to when we added the profile binding controller, which is a bit of a shame imo.

Yeah, it's not too nice that the operator got more complicated... but I just didn't figure out a better way of doing this. Having everything in the DaemonSet ended up in a ton of races. Having the deployment do this ends up addressing that, even if reconciling is a little slower.

In the end, it kinda resembles the kube architecture more:

* The Kube API == the Deployment: Here is where most of the controller logic actually lives, and it aggregates the results to one consistent one.
* The kubelet == the DaemonSet: It merely listens for objects and makes sure "it happens". So, this just makes sure that the profile is installed and reports its own status. It is then up to the Deployment to gather the status into one cohesive one.

saschagrunert (approved):

saschagrunert:
/hold

for @cmurphy

internal/pkg/manager/workloadannotator/workloadannotator.go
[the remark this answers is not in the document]
JAORMX: The main issue was that by having the reconciler that received events for pods in the DaemonSet, each instance might receive the same event for the same pod, so it would try to update the same Seccomp profile; resulting in races. By moving that reconciler to the Deployment, that ensures that only the leader does this, and so it eliminates the race conditions on that side.

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=securityprofilenodestatuses,verbs=get;list;watch
+
+// Reconcile reconciles a NodeStatus.
+func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
+	logger := r.log.WithValues("nodeStatus", req.Name, "namespace", req.Namespace)
+
+	ctx, cancel := context.WithTimeout(context.Background(), reconcileTimeout)
+	defer cancel()
+
+	logger.Info("status change")
```
cmurphy, line 91: nit: could maybe add more info or omit this log

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// get the status to be reconciled
+	instance := &statusv1alpha1.SecurityProfileNodeStatus{}
+	if err := r.client.Get(ctx, req.NamespacedName, instance); err != nil {
+		// Expected to find a node profile, return an error and requeue
+		// TODO(jhrozek): ignore not found?
+		return reconcile.Result{}, errors.Wrap(util.IgnoreNotFound(err), "Cannot find the node status")
```
cmurphy, line 98: Not sure `IgnoreNotFound` is right here, since it will convert the error to nil and then errors.Wrap will be wrapping nil instead of an error. If it's returning an error either way then may as well keep the full error context.

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+
+	lprof := logger.WithValues(
+		"Profile.Name", prof.GetName(),
+		"Profile.Namespace", prof.GetNamespace(),
+		"Profile.Kind", prof.GetObjectKind().GroupVersionKind(),
+	)
+
+	// get all the other statuses
+	// TODO: Delete status... it's just plain wrong
```
cmurphy, line 113: Still TODO?

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+	profName := instance.Labels[statusv1alpha1.StatusToProfLabel]
+	if profName == "" {
+		return reconcile.Result{}, errors.New("Unlabeled node status")
+	}
+
+	// TODO: Delete status... it's just plain wrong
```
cmurphy, line 119: Still TODO?

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// get the DS
+	spodDS, err := r.getDS(ctx, config.GetOperatorNamespace(), lprof)
+	if err != nil {
+		return reconcile.Result{}, errors.Wrap(err, "cannot get the DS")
+	}
+
+	if !daemonSetIsReady(spodDS) || daemonSetIsUpdating(spodDS) {
+		// If the DS is not ready or updating, don't bother updating the status
+		// TODO(jhrozek): is this needed? It sounds safer..
```
cmurphy, line 148: Still TODO? I agree it probably makes sense to requeue in this case

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+		return nil, errors.Wrap(err, "cannot create DS list label")
+	}
+	dsSelect.Add(*spodDSFilter)
+	dsListOpts := client.ListOptions{
+		LabelSelector: dsSelect,
+		Namespace:     namespace,
+	}
+
+	spodDSList := appsv1.DaemonSetList{}
+	if err := r.client.List(ctx, &spodDSList, &dsListOpts); err != nil {
```
cmurphy, line 185: Don't we know its name? Why do a `List` instead of a `Get`?

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,286 @@
...
+func daemonSetIsReady(ds *appsv1.DaemonSet) bool {
+	return ds.Status.DesiredNumberScheduled > 0 && ds.Status.DesiredNumberScheduled == ds.Status.NumberAvailable
+}
+
+func daemonSetIsUpdating(ds *appsv1.DaemonSet) bool {
+	return ds.Status.UpdatedNumberScheduled > 0 &&
+		(ds.Status.UpdatedNumberScheduled < ds.Status.DesiredNumberScheduled || ds.Status.NumberUnavailable > 0)
+}
+
+func listStatusesForPolicy(
```
cmurphy, line 273: nit: `listStatusesForProfile`

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -36,9 +37,24 @@ type StatusClient struct {
 	pol             client.Object
 	nodeName        string
 	finalizerString string
+	client          client.Client
 }
 
-func NewForProfile(pol client.Object) (*StatusClient, error) {
+func ProfilePredicate(obj runtime.Object, wantKind string) bool {
```
cmurphy, line 43: I don't see this used anywhere, why is it needed?

cmurphy (commented):
Sorry for the delay, overall I'm really happy with this change! Glad to see this becoming more stable and resilient. Left some minor comments and questions. The final fixup commit could potentially be rolled into one of the main commits for the sake of clean history.

> In the end, it kinda resembles the kube architecture more:
> 
> The Kube API == the Deployment: Here is where most of the controller logic actually lives, and it aggregates the results to one consistent one.
> The kubelet == the DaemonSet: It merely listens for objects and makes sure "it happens". So, this just makes sure that the profile is installed and reports its own status. It is then up to the Deployment to gather the status into one cohesive one.

That makes a lot of sense when you frame it that way. It makes me wonder if there's a more elegant way to handle 94895b1e85dfe6245429365fa8932a2af0c7afc2 though. I don't have any ideas at the moment.

internal/pkg/manager/nodestatus/nodestatus.go
JAORMX: Mind if I update this in anther PR?

JAORMX:
The  cla/linuxfoundation check seems to have gotten stuck

internal/pkg/manager/nodestatus/nodestatus.go
cmurphy: Sure :+1:

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -0,0 +1,282 @@
...
+	defer cancel()
+
+	logger := r.log.WithValues("nodeStatus", req.Name, "namespace", req.Namespace)
+	logger.Info("Reconciling node status")
+
+	// get the status to be reconciled
+	instance := &statusv1alpha1.SecurityProfileNodeStatus{}
+	if err := r.client.Get(ctx, req.NamespacedName, instance); err != nil {
+		// Expected to find a node profile, return an error and requeue
+		// TODO(jhrozek): ignore not found?
```
cmurphy, line 96: Still TODO?

cmurphy (approved):
One nitpick but could be addressed later

PR #396, opened 2021-04-08
Change SelinuxPolicy to SelinuxProfile

JAORMX:
#### What type of PR is this?

/kind cleanup
/kind api-change

SelinuxPolicy was always a little different when it shouldn't be. In order to
standardize on naming and make this a more cohesive operator, this changes the
name of the aforementioned object to SelinuxProfile. Thus matching
`SeccompProfile` which was supported from the start.

None

#### Does this PR have test?

N/A.

```release-note
SelinuxPolicy has been removed and is now SelinuxProfile.
```


internal/pkg/daemon/selinuxprofile/setup.go
```diff
@@ -38,14 +38,14 @@ func Setup(ctx context.Context, mgr ctrl.Manager, l logr.Logger) error {
 	if err != nil {
 		return errors.Wrap(err, "creating policy wrapper template")
```
saschagrunert, line 39:
```suggestion
		return errors.Wrap(err, "creating profile wrapper template")
```
also renaming the above `policyWrapper` to `profileWrapper`?

internal/pkg/daemon/selinuxprofile/utils.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package selinuxpolicy
+package selinuxprofile
 
 // NOTE(jaosorior): We can probably use a library for generating names instead
 // of using this.
```
saschagrunert, line 20: What about changing `GetPolicyK8sName` within this file, too?

saschagrunert (commented):
Do we wanna change `examples/selinuxpolicy.yaml`, too?

JAORMX:
@saschagrunert thanks for the review! Addressed the comments.

saschagrunert (approved):

PR #402, opened 2021-04-15
Add profile binding unit tests

saschagrunert:
#### What type of PR is this?

/kind bug

Adding unit test cases to the binding webhook.

This also fixes a bug where the binding error did not get returned
correctly which would cause the webhook to crash.

None
#### Does this PR have test?
Yes

None

```release-note
None
```


internal/pkg/webhooks/binding/binding.go
```diff
@@ -173,7 +173,7 @@ func (p *podSeccompBinder) getSeccompProfile(
 	err = util.Retry(
 		func() (retryErr error) {
 			seccompProfile, retryErr = p.GetSeccompProfile(ctx, key)
-			if err != nil {
+			if retryErr != nil {
```
saschagrunert, line 176: Whoops, that's a bug.

PR #419, opened 2021-04-27
Bump VM based e2e test to k8s v1.21.0

saschagrunert:
#### What type of PR is this?

/kind cleanup

Bump Kubernetes to the latest release for the vagrant based e2e tests.
#### Does this PR have test?
None

None

```release-note
None
```


JAORMX (approved):
we should probably have a cron job that does this automatically.

saschagrunert:
> we should probably have a cron job that does this automatically.

Most of the versions we use (kustomize, linter, k8s) are scattered around the project and it's likely that we miss them. I can check if we can use something like a single dependencies file for that.

Issue #276, opened 2021-02-04
Introduce metrics for the operator


saschagrunert:
We should think about a metrics endpoint which provides aggregated data. For example we could provide the amount of errored policies, which is more useful for SRE's than debugging events.

saschagrunert:
Follow-ups of #422:
- decide which metrics we would like to implement
- provide documentation around the feature
- add metric related e2e tests
- integrate controller-runtime metris: https://book-v1.book.kubebuilder.io/beyond_basics/controller_metrics.html (currently served on :8080 per default)
- secure the metrics via tls and authentication (maybe via an auth proxy or directly within the implementation)

jhrozek:
@saschagrunert We're interested in adding some metrics for the SELinux part. Would you prefer a separate ticket or just as part of this one. Of course we're going to do the work :-) I'm just asking about the issue.

saschagrunert:
> @saschagrunert We're interested in adding some metrics for the SELinux part. Would you prefer a separate ticket or just as part of this one. Of course we're going to do the work :-) I'm just asking about the issue.

Sure, let's do it as part of this issue, too. :+1:

saschagrunert:
This is done for now.

PR #422, opened 2021-04-27
Add initial metrics support

saschagrunert:
#### What type of PR is this?

/kind feature

This adds the necessary code infrastructure to allow serving metrics
within the operator daemonset (spod). The metrics are for now served
unencrypted on port :8081. An additional service `metrics` can be used
for scraping from prometheus (by using the auto-discover plugin
`kubernetes_sd_configs` or a `ServiceMonitor` together with the
prometheus-operator).

This patch features a sample metric counter
`security_profiles_operator_seccomp_profile`, which can be distinguished
by the labels `update` and `delete`. For example, if we create 4 new
profiles and delete them afterwards, then the output should look like:

```
\# HELP security_profiles_operator_seccomp_profile Counter about seccomp profile operations.
\# TYPE security_profiles_operator_seccomp_profile counter
security_profiles_operator_seccomp_profile{operation="delete"} 4
security_profiles_operator_seccomp_profile{operation="update"} 4
```

Things to be done after this patch:

- decide which metrics we would like to implement
- provide documentation around the feature
- add metric related e2e tests

Refers to #276 
#### Does this PR have test?
Yes

```release-note
Added `metrics` service endpoint to the operator namespace, which now serves the `security_profiles_operator_seccomp_profile` metric.
```


JAORMX (approved):
yes! This looks like a great start!

PR #424, opened 2021-04-28
Add RBAC proxy to metrics

saschagrunert:
#### What type of PR is this?

/kind feature

We now use the kube-rbac-proxy to provide two metrics endpoints. Those
get exposed through the `metrics-spod` and `metrics-controller-runtime`
services, serving an authorized self-signed TLS connection.

To be able to retrieve the metrics, it is now required to deploy a
`ClusterRole` and `ClusterRoleBinding` for a `ServiceAccount`. For
example:

```yaml
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: spo-metrics-client
rules:
- nonResourceURLs:
  - /metrics
  verbs:
  - get
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: spo-metrics-client
roleRef:
  kind: ClusterRole
  name: spo-metrics-client
subjects:
- kind: ServiceAccount
  name: default
```

The above example is also part of the default deployment, and can be
tested via `kubectl run` within the `security-profiles-operator`
namespace by running:

```
> kubectl run --rm -it pod --image=alpine -- sh -c \
    'apk update && apk add bash curl &&
     curl -vsk -H "Authorization: Bearer `cat /var/run/secrets/kubernetes.io/serviceaccount/token`"
     https://metrics-spod/metrics'
…
*   Trying 10.0.0.222:443...
* Connected to metrics-spod (10.0.0.222) port 443 (#0)
…
> authorization: Bearer xxx
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
< HTTP/2 200
< content-type: text/plain; version=0.0.4; charset=utf-8
< date: Wed, 28 Apr 2021 12:00:46 GMT
…
security_profiles_operator_seccomp_profile{operation="update"} 1
```

Documentation and e2e testing will follow in further patches.

Refers to #422 
#### Does this PR have test?

None
None

```release-note
Deploying kube-rbac-proxy sidecar in SPOD for exposing metrics via the new `metrics-spod` and `metrics-controller-runtime` services.
```


internal/pkg/manager/spod/bindata/spod.go
```diff
@@ -352,6 +357,72 @@ semodule -i /opt/spo-profiles/selinuxd.cil
...
+								v1.ResourceEphemeralStorage: resource.MustParse("20Mi"),
+							},
+						},
+						SecurityContext: &v1.SecurityContext{
+							AllowPrivilegeEscalation: &falsely,
+							RunAsUser:                &userRootless,
+							SELinuxOptions: &v1.SELinuxOptions{
+								// TODO(pjbgf): Use a more restricted selinux type
+								Type: "spc_t",
+							},
```
JAORMX, line 387: This probably doesn't need `spc_t`. Let's leave that empty so it uses the default `container_t`

JAORMX (changes requested):
the rbac-proxy isn't using TLS, is it?

saschagrunert:
> the rbac-proxy isn't using TLS, is it?

It is using a self-signed one right now and we also can provide an external one.

internal/pkg/manager/spod/bindata/spod.go
saschagrunert: Removed as suggested :heavy_check_mark:

deploy/namespace-operator.yaml
```diff
@@ -1533,14 +1546,59 @@ kind: Service
...
+    targetPort: 8442
+  selector:
+    app: security-profiles-operator
+    name: spod
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: security-profiles-operator
```
JAORMX, line 1564: Should this point to the label for the SPOD? Seems like the spod has a different label.

deploy/base/service.yaml
```diff
@@ -14,11 +14,25 @@ spec:
...
 spec:
   ports:
-    - port: 80
-      targetPort: 8081
+  - name: https
+    port: 443
+    targetPort: 8442
   selector:
-    app: spod
+    name: spod
```
JAORMX, line 25: should this point to the manager's label? It would be `name: security-profiles-operator`

JAORMX (changes requested):

deploy/base/service.yaml
saschagrunert: No, the metrics endpoints are served by the spod only.

JAORMX (commented):
/lgtm

PR #429, opened 2021-04-29
Merge metrics endpoints for using different paths

saschagrunert:
#### What type of PR is this?

/kind feature
/kind design

We now use the additional handler from the controller runtime to serve
the metrics under a new path `/metrics-spod`. This way we can ensure
having only one service and one kube-rbac-proxy instance.

Refers to #276 
#### Does this PR have test?

None

```release-note
None
```

### DIFF
--- a/deploy/base/metrics_client.yaml
+++ b/deploy/base/metrics_client.yaml
@@ -6,6 +6,7 @@ metadata:
 rules:
 - nonResourceURLs:
   - /metrics
+  - /metrics-spod
   verbs:
   - get
 ---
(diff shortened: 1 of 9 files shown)


deploy/base/metrics_client.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 rules:
 - nonResourceURLs:
   - /metrics
```
cmurphy, line 8: Is this still needed?

saschagrunert: Yep, we need both paths to be accessible.

PR #430, opened 2021-04-30
Add liveness/startup probe to SPOD

saschagrunert:
#### What type of PR is this?

/kind feature

We now add a liveness probe for giving the SPOD more time on startup to
become ready. Every controller should at least run one `Reconile` to be
marked as ready.

We now also add a new interface `Controller`
(`internal/pkg/controller/controller.go`), which has to be fulfilled by
every SPOD controller. This allows to simplify the implementation in
`main`.

None
#### Does this PR have test?
None

None

```release-note
Added liveness and startup probe to operator daemon set to streamline the operator stratup.
```

### DIFF
--- a/internal/pkg/controller/controller.go
+++ b/internal/pkg/controller/controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"net/http"
 
 	"github.com/go-logr/logr"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -36,4 +37,7 @@ type Controller interface {
 
 	// Setup is the initialization of the controller.
 	Setup(context.Context, ctrl.Manager, logr.Logger, *metrics.Metrics) error
+
+	// Healthz is the liveness probe endpoint of the controller.
+	Healthz(*http.Request) error
 }
--- a/internal/pkg/daemon/profilerecorder/profilerecorder.go
+++ b/internal/pkg/daemon/profilerecorder/profilerecorder.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
@@ -138,6 +139,11 @@ func (r *RecorderReconciler) Setup(
 		)).
 		For(&corev1.Pod{}).
 		Complete(r)
+}
+
+// Healthz is the liveness probe endpoint of the controller.
+func (r *RecorderReconciler) Healthz(*http.Request) error {
+	return nil
 }
 
 func (r *RecorderReconciler) isPodOnLocalNode(obj runtime.Object) bool {
--- a/internal/pkg/daemon/seccompprofile/seccompprofile.go
+++ b/internal/pkg/daemon/seccompprofile/seccompprofile.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path"
 	"time"
@@ -38,6 +39,7 @@ import (
 
 	"sigs.k8s.io/security-profiles-operator/api/seccompprofile/v1alpha1"
 	statusv1alpha1 "sigs.k8s.io/security-profiles-operator/api/secprofnodestatus/v1alpha1"
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/atomic"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/controller"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/daemon/metrics"
@@ -86,6 +88,7 @@ type Reconciler struct {
 	record  event.Recorder
 	save    saver
 	metrics *metrics.Metrics
+	ready   atomic.Bool
 }
 
 // Name returns the name of the controller.
@@ -118,6 +121,14 @@ func (r *Reconciler) Setup(
 		Complete(r)
 }
 
+// Healthz is the liveness probe endpoint of the controller.
+func (r *Reconciler) Healthz(*http.Request) error {
+	if !r.ready.Get() {
+		return errors.New("not ready")
+	}
+	return nil
+}
+
 // Security Profiles Operator RBAC permissions to manage SeccompProfile
 // nolint:lll
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=seccompprofiles,verbs=get;list;watch;create;update;patch
@@ -135,6 +146,11 @@ func (r *Reconciler) Setup(
 
 // Reconcile reconciles a SeccompProfile.
 func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
+	// Mark the controller as ready if the first reconcile has been finished
+	if !r.ready.Get() {
+		defer func() { r.ready.Set(true) }()
+	}
+
 	logger := r.log.WithValues("profile", req.Name, "namespace", req.Namespace)
 
 	ctx, cancel := context.WithTimeout(context.Background(), reconcileTimeout)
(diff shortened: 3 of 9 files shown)


internal/pkg/daemon/profilerecorder/profilerecorder.go
```diff
@@ -152,6 +184,9 @@ func (r *RecorderReconciler) isPodWithTraceAnnotation(obj runtime.Object) bool {
 
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
 func (r *RecorderReconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
+	// Mark the controller as ready if the first reconcile has been finished
+	defer func() { r.ready = true }()
```
JAORMX, line 188: Do reconcile looks run serially? If they're concurrent, this could lead to a race condition

saschagrunert: Yeah, how about that: https://github.com/kubernetes-sigs/security-profiles-operator/blob/53651fe8313dcb84324f6e7c07ecd79ac40f9499/internal/pkg/atomic/bool.go

jhrozek:
First, I really like the new shared Controller interface, that's a really nice refactor.

About the probes, maybe a silly question, but I'm very good at asking those -- aren't the Reconcile loops reactive and invoked only when the watcher hits something? Meaning, would the loop go through at least once if there are no selinux/seccomp profiles installed by default or if all are removed by the admin?

saschagrunert:
> First, I really like the new shared Controller interface, that's a really nice refactor.

Thank you! :pray: 

> About the probes, maybe a silly question, but I'm very good at asking those -- aren't the Reconcile loops reactive and invoked only when the watcher hits something? Meaning, would the loop go through at least once if there are no selinux/seccomp profiles installed by default or if all are removed by the admin?

Indeed the reconcile will not be hit if no objects to react on are available. That's exactly the reason why I cannot tell right now when the recording reconciler is ready. For seccomp and selinux profiles I just assume that we will definitely ship some default ones, which can be then used as indicator for readiness.

The other way around, do you have anything in mind how we could indicate readiness in a better way?

jhrozek:
On Mon, May 03, 2021 at 01:05:34AM -0700, Sascha Grunert wrote:
> > First, I really like the new shared Controller interface, that's a really nice refactor.
> 
> Thank you! :pray: 
> 
> > About the probes, maybe a silly question, but I'm very good at asking those -- aren't the Reconcile loops reactive and invoked only when the watcher hits something? Meaning, would the loop go through at least once if there are no selinux/seccomp profiles installed by default or if all are removed by the admin?
> 
> Indeed the reconcile will not be hit if no objects to react on are available. That's exactly the reason why I cannot tell right now when the recording reconciler is ready. For seccomp and selinux profiles I just assume that we will definitely ship some default ones, which can be then used as indicator for readiness.

OK. I was wondering a bit whether to create some default objects
straight up in the code, but I guess creating them with the manifests is
as good.

> 
> The other way around, do you have anything in mind how we could indicate readiness in a better way?

For readiness no, not really. I was wondering a bit about thinkgs like
trying to contact selinuxd for the liveness probe in the selinux
controller, but that's handled in the controller anyway.

Thanks for the replies.

internal/pkg/daemon/selinuxprofile/selinuxpolicy_controller.go
```diff
@@ -96,6 +103,21 @@ type ReconcileSP struct {
 	record         event.Recorder
 }
 
+// Name returns the name of the controller.
+func (r *ReconcileSP) Name() string {
+	return "seccomp-spod"
```
jhrozek, line 108: Nit: I guess this should say selinux-spod?

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -94,6 +88,45 @@ type Reconciler struct {
 	record  event.Recorder
 	save    saver
 	metrics *metrics.Metrics
+	ready   atomic.Bool
+}
+
+// Name returns the name of the controller.
+func (r *Reconciler) Name() string {
+	return "selinux-spod"
```
jhrozek, line 96: And this one should say seccomp-spod?

internal/pkg/daemon/selinuxprofile/selinuxpolicy_controller.go
saschagrunert: Yep, fixed as suggested :+1:

internal/pkg/daemon/seccompprofile/seccompprofile.go
saschagrunert: Yep, fixed as suggested :+1:

saschagrunert:
Let's see if I can get the e2e tests happy.

internal/pkg/controller/controller.go
```diff
@@ -0,0 +1,43 @@
...
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/daemon/metrics"
+)
+
+// Controller is the interface every controller should fulfill.
+type Controller interface {
+	// Name returns the name of the controller.
+	Name() string
+
+	// SchemaBuilder returns the registered schema of the controller.
+	SchemaBuilder() *scheme.Builder
```
cmurphy, line 36: (nit) the field is called `SchemeBuilder` so wouldn't it be better to call this `SchemeBuilder()` (a->e)?

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -113,6 +146,11 @@ type Reconciler struct {
 
 // Reconcile reconciles a SeccompProfile.
 func (r *Reconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
+	// Mark the controller as ready if the first reconcile has been finished
```
cmurphy, line 149: So the controller is considered fully ready even if only one daemon has finished?

internal/pkg/controller/controller.go
saschagrunert: Yep, fixed as suggested :+1:

internal/pkg/daemon/seccompprofile/seccompprofile.go
saschagrunert: Right now, the daemonset pod is considered to be ready if the initial seccomp profile reconcile has been finished. We should find better readiness metrics for selinux and the recorder controller somehow, but I was not sure which ones we could use.

cmurphy: I was more curious about the pods running on different nodes. If the daemonset is running 1 pod on 10 nodes, and only half of them complete a Reconcile loop and the other half fail for some reason, is the controller considered ready? Or does it not matter for this?

saschagrunert: So, for now the daemonset only runs 1 pod for each node, otherwise we would risk reconciliation races per node. I think this will also not change in the near future if I'm not missing anything. We also only apply the readiness probes to the SPOD (daemonset instances) and not the manager, which would be a bit more complicated from my point of view.

saschagrunert:
Decided to carve-out the controller logic into a separate PR (https://github.com/kubernetes-sigs/security-profiles-operator/pull/437) and will rebase this one later.

internal/pkg/daemon/seccompprofile/seccompprofile.go
cmurphy: Right, I definitely didn't mean we should run more than one pod per node. I think I understand now, thanks.

saschagrunert:
Rebased on top of the latest master.

saschagrunert:
@JAORMX @jhrozek please take a look.

JAORMX (approved):
/lgtm

This is looking great! Since this only addresses the Seccomp controller, we should track as issues the controllers that are missing so they get addressed. We're hoping to onboard someone to the SPO soon and those would be some great introductory tasks.

saschagrunert:
> …we should track as issues the controllers that are missing so they get addressed.

Sounds good, opened #442 and #443

PR #440, opened 2021-05-07
Set resource version for SPOD update

saschagrunert:
#### What type of PR is this?

/kind bug

Every update would fail because the resource version won't be set. This patch fixes this corner case.

None
#### Does this PR have test?
None

None

```release-note
Fixed SPOD update bug for default profiles where resource version is not set.
```

### DIFF
--- a/internal/pkg/manager/spod/spod_controller.go
+++ b/internal/pkg/manager/spod/spod_controller.go
@@ -281,6 +281,9 @@ func (r *ReconcileSPOd) handleUpdate(
 		foundProfile := &sccmpv1alpha1.SeccompProfile{}
 		var err error
 		if err = r.client.Get(ctx, pKey, foundProfile); err == nil {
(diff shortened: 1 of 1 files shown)


internal/pkg/manager/spod/spod_controller.go
```diff
@@ -281,6 +281,9 @@ func (r *ReconcileSPOd) handleUpdate(
 		foundProfile := &sccmpv1alpha1.SeccompProfile{}
 		var err error
 		if err = r.client.Get(ctx, pKey, foundProfile); err == nil {
```
hasheddan, line 283: Could consider using an `Applicator` :) https://github.com/crossplane/crossplane-runtime/blob/53874539b31013368170eb1090d2f083745d5051/pkg/resource/api.go#L168

saschagrunert:
Whoop this already has been fixed :shrug:

PR #458, opened 2021-05-19
Add default ServiceMonitor if the CRD is available

saschagrunert:
#### What type of PR is this?

/kind feature

We now deploy a default ServiceMonitor object for the operator if the
CRD is available within the cluster. This allows automatic metrics
discovery via the prometheus operator.

Refers to #276 
#### Does this PR have test?

None
None

```release-note
Added automatic ServiceMonitor deployment if the CRD is available within the cluster.
```


internal/pkg/manager/spod/spod_controller.go
```diff
@@ -275,6 +280,22 @@ func (r *ReconcileSPOd) handleCreate(
...
 	}
 
+	hasServiceMonitorCRD, err := r.hasServiceMonitorCRD(ctx)
+	if err != nil {
+		return errors.Wrap(err, "check for ServiceMonitor CRD")
+	}
+
+	if hasServiceMonitorCRD {
+		r.log.Info("Deploying operator service monitor")
+		if err := r.client.Create(
```
JAORMX, line 290: you could just do the call to `Create` and if the error indicates that the object is not registered or is on an unknown type, then you could just ignore it.

e.g.

```
if runtime.IsNotRegisteredError(err) || meta.IsNoMatchError(err) {
```

where the functions come from:

```
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/api/meta"
```

JAORMX (changes requested):

internal/pkg/manager/spod/spod_controller.go
saschagrunert: Ohhh, that’s a good workaround. I’ll change it.

Issue #467, opened 2021-06-02
Run log enricher as part of the daemon


saschagrunert:
This allows us re-using the metrics endpoint from the daemon and create metrics for the log enrichment feature. Right now the enricher runs in a dedicated container:

https://github.com/kubernetes-sigs/security-profiles-operator/blob/12d2a6a26d06695506c81ebe5122a6d49ede4082/internal/pkg/manager/spod/bindata/spod.go#L357-L399

JAORMX:
@jhrozek @saschagrunert this would also be useful for selinux profile recording, as we could gather those audit logs and pass them to [udica](https://github.com/containers/udica) to add any missing rules to the generated profile.

saschagrunert:
Obsolete now

PR #468, opened 2021-06-04
WIP: Move log enricher into daemon

saschagrunert:
#### What type of PR is this?

/kind feature

This moves the log enricher into the daemon as well as bootstraps a symlink to `/dev/kmsg` during the non-root-enabler. Drawback is that the daemonset now needs HostPid permissions which may be a blocker.

Fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/467

#### Does this PR have test?

None

None

```release-note
TBD
```

### DIFF
--- a/internal/pkg/manager/spod/bindata/spod.go
+++ b/internal/pkg/manager/spod/bindata/spod.go
@@ -34,7 +34,7 @@ var (
 	truly                           = true
 	userRoot                  int64 = 0
 	userRootless                    = int64(config.UserRootless)
-	hostPathFile                    = v1.HostPathFile
+	hostPathCharDev                 = v1.HostPathCharDev
 	hostPathDirectory               = v1.HostPathDirectory
 	hostPathDirectoryOrCreate       = v1.HostPathDirectoryOrCreate
 	metricsPort               int32 = 9443
@@ -47,7 +47,6 @@ const (
 	SelinuxdPrivateDir   = "/var/run/selinuxd"
 	SelinuxdSocketPath   = SelinuxdPrivateDir + "/selinuxd.sock"
 	SelinuxdDBPath       = SelinuxdPrivateDir + "/selinuxd.db"
-	varLogSpoPath        = "/var/log/spo.log"
 	MetricsImage         = "quay.io/brancz/kube-rbac-proxy:v0.9.0"
 )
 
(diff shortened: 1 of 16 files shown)


saschagrunert:
It still does not work because of the following error from the daemon:
```
E0604 11:15:38.285628 1857417 enricher.go:52] log-enricher "msg"="tail audit log" "error"="read audit log buffer: read /var/log/spo/spo.log: invalid argument"
E0604 11:15:38.285643 1857417 deleg.go:144] setup "msg"="running log enricher" "error"="tail failed: read audit log buffer: read /var/log/spo/spo.log: invalid argument"
```

internal/pkg/manager/spod/bindata/spod.go
```diff
@@ -200,6 +208,7 @@ semodule -i /opt/spo-profiles/selinuxd.cil
 						},
 					},
 				},
+				HostPID: true, // required for the log-enricher
```
saschagrunert, line 211: This is uncool

internal/pkg/manager/spod/bindata/spod.go
```diff
@@ -87,6 +86,10 @@ var Manifest = &appsv1.DaemonSet{
 								Name:      "host-varlib-volume",
 								MountPath: "/var/lib",
 							},
+							{
+								Name:      "host-varlog-volume",
+								MountPath: "/var/log",
```
JAORMX, line 91: Should we specifically mount `/var/log/spo`?

internal/pkg/manager/spod/bindata/spod.go
JAORMX: Why is this so?

saschagrunert: The enricher is using the process ID to retrieve the container ID:
https://github.com/kubernetes-sigs/security-profiles-operator/blob/12d2a6a26d06695506c81ebe5122a6d49ede4082/internal/pkg/daemon/enricher/main.go#L73

JAORMX: Why don't we keep it on its own container then to reduce the attack vector? The  only that container gets access to the host PIDs while the daemon doesn't. We can enable communication with the daemon via a Unix domain socket or something of the sorts.

saschagrunert: Like this https://github.com/kubernetes-sigs/security-profiles-operator/pull/469?

JAORMX: @saschagrunert you're a magician! :mage_man: that looks quite promising!

saschagrunert:
I'll come-up with a new proposal
/close

PR #479, opened 2021-06-09
Fix log enricher read and automatically add `/dev/kmsg` device 

saschagrunert:
#### What type of PR is this?

/kind feature

When using CRI-O, we can utilize the annotation
`io.kubernetes.cri-o.Devices` to correctly mount the device into the
container. This means we do not rely on a symlink any more and can use
the device directly.

The enricher read has been fixed because one read always returns a full
line which makes the offset calculation obsolete.

None
#### Does this PR have test?
None

None

```release-note
Automatically mount /dev/kmsg for log enricher usage if running with CRI-O and an allowed `io.kubernetes.cri-o.Devices` annotation.
```

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -67,6 +67,9 @@ const (
 
 	// EnricherLogFile is the path to the kernel messages log file used for the enricher.
 	EnricherLogFile = "/var/log/spo.log"
+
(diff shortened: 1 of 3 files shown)


saschagrunert:
Rebased, let's see if the tests pass now otherwise it's a bug in the PR.

PR #487, opened 2021-06-18
Switch to auditd logs for log enricher

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should rely on the auditd logs for the log enricher since the
audit messages will not appear in `/dev/kmsg` if auditd is running.

We can assume that most distributions ship with auditd enabled, and if
the file does not exist we simply handle the log enricher as no-op.

None
#### Does this PR have test?
None

@pjbgf I have no idea how we could avoid the `privileged` container. A symlink will not do the trick because the audit logs are usually really restrictive.

```release-note
Log enricher now requires running auditd (`/var/log/audit/audit.log`)
```


JAORMX:
@saschagrunert the only way to avoid a privileged container is to run it with a custom selinux/apparmor profile. Then we could add permissions to read the audit logs. The problem is the bootstrapping of those profiles.

saschagrunert:
Hm, yeah probably. In any case I think we have to rely on auditd logs for the enrichment feature, so this PR is ready for review.

JAORMX (approved):
/lgtm

PR #491, opened 2021-06-21
Retry log-enricher if container ID is empty

saschagrunert:
#### What type of PR is this?

/kind feature

During the creation of a container, it may be possible that the
container ID is still empty while looking up for log enrichment.

We now retry the whole process to be able to make the enricher more
robust.

None
#### Does this PR have test?
None

None

```release-note
Add retry functionality to log enricher if container ID is still empty during pod creation.
```


JAORMX:
/lgtm

Nice!

JAORMX (approved):

PR #504, opened 2021-06-24
Fix golang 1.16 usage

saschagrunert:
#### What type of PR is this?

/kind failing-test

We have to explicitly modify the path to set the correct golang version.
The package can be removed from the install script as well.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/hack/ci/Vagrantfile
+++ b/hack/ci/Vagrantfile
@@ -28,12 +28,11 @@ Vagrant.configure("2") do |config|
       echo 'export PATH=$PATH:/usr/local/go/bin' >> /etc/profile
       echo 'export GOPATH="$HOME/go"' >> /etc/profile
       echo 'export GOBIN="$GOPATH/bin"' >> /etc/profile
-      echo '. /etc/profile' >> $HOME/.bashrc
 
       dnf install -y \
         conntrack \
         container-selinux \
-        golang-go \
+        git \
         iptables \
         jq \
         make \
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -54,7 +54,6 @@ func (e *e2e) TestSecurityProfilesOperator() {
 
 	// Deploy the operator
 	e.deployOperator(manifest)
-	defer e.run("git", "checkout", manifest)
 
 	// Retrieve the inputs for the test cases
 	nodes := e.getWorkerNodes()
@@ -133,6 +132,7 @@ func (e *e2e) TestSecurityProfilesOperator() {
 
 	// Clean up cluster-wide deployment to prepare for namespace deployment
 	e.cleanupOperator(manifest)
+	e.run("git", "checkout", manifest)
 
 	e.logf("testing namespace operator")
 
@@ -145,7 +145,6 @@ func (e *e2e) TestSecurityProfilesOperator() {
 		"sed", "-i", fmt.Sprintf("s/NS_REPLACE/%s/", testNamespace),
 		namespaceManifest,
 	)
-	defer e.run("git", "checkout", namespaceManifest)
 	// All following operations such as create pod will be in the test namespace
 	e.kubectl("config", "set-context", "--current", "--namespace", testNamespace)
 	e.deployOperator(namespaceManifest)
@@ -156,6 +155,7 @@ func (e *e2e) TestSecurityProfilesOperator() {
 			tc.fn(nodes)
 		})
 	}
+	e.run("git", "checkout", namespaceManifest)
 }
 
 func (e *e2e) deployCertManager() {
--- a/hack/ci/env.sh
+++ b/hack/ci/env.sh
@@ -16,7 +16,9 @@ local_ip() {
     ip route get 1.2.3.4 | cut -d ' ' -f7 | tr -d '[:space:]'
 }
 
-GOPATH=$(go env GOPATH)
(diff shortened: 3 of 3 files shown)


test/e2e_test.go
```diff
@@ -54,7 +54,6 @@ func (e *e2e) TestSecurityProfilesOperator() {
 
 	// Deploy the operator
 	e.deployOperator(manifest)
-	defer e.run("git", "checkout", manifest)
```
JAORMX, line 57: why was this removed?

saschagrunert: I have no idea, the defer seems to change something when building it with go 1.16. This caused the error:

```
   suite_test.go:393: 
        	Error Trace:	suite_test.go:393
        	            				e2e_test.go:158
        	Error:      	Expected nil, but got: &exec.Error{Name:"git", Err:(*errors.errorString)(0xc0000723e0)}
        	Test:       	TestSuite/TestSecurityProfilesOperator
    suite.go:63: test panicked: runtime error: invalid memory address or nil pointer dereference
        goroutine 38 [running]:
        runtime/debug.Stack(0xc0000af4b8, 0x1464580, 0x2083bf0)
        	/usr/local/go/src/runtime/debug/stack.go:24 +0x9f
        github.com/stretchr/testify/suite.failOnPanic(0xc00018a780)
        	/vagrant/vendor/github.com/stretchr/testify/suite/suite.go:63 +0x5b
        panic(0x1464580, 0x2083bf0)
        	/usr/local/go/src/runtime/panic.go:965 +0x1b9
        sigs.k8s.io/release-utils/command.(*Stream).OutputTrimNL(...)
        	/vagrant/vendor/sigs.k8s.io/release-utils/command/command.go:350
        sigs.k8s.io/security-profiles-operator/test_test.(*e2e).run(0xc00045a980, 0x15d8539, 0x3, 0xc0000af7c0, 0x2, 0x2, 0x25, 0x24)
        	/vagrant/test/suite_test.go:394 +0xc1
        sigs.k8s.io/security-profiles-operator/test_test.(*e2e).TestSecurityProfilesOperator(0xc00045a980)
        	/vagrant/test/e2e_test.go:158 +0xde5
        reflect.Value.call(0xc00050ff20, 0xc000186c68, 0x13, 0x15d86ff, 0x4, 0xc00006de30, 0x1, 0x1, 0xc00006dcf8, 0x40d9ca, ...)
        	/usr/local/go/src/reflect/value.go:476 +0x8e7
        reflect.Value.Call(0xc00050ff20, 0xc000186c68, 0x13, 0xc00006de30, 0x1, 0x1, 0x1a3a26c, 0x24, 0x439)
        	/usr/local/go/src/reflect/value.go:337 +0xb9
        github.com/stretchr/testify/suite.Run.func1(0xc00018a780)
        	/vagrant/vendor/github.com/stretchr/testify/suite/suite.go:158 +0x379
        testing.tRunner(0xc00018a780, 0xc000128090)
        	/usr/local/go/src/testing/testing.go:1194 +0xef
        created by testing.(*T).Run
        	/usr/local/go/src/testing/testing.go:1239 +0x2b3
```

saschagrunert:
@JAORMX do you think we can get this in to stabilize the CI (at least a bit)?

test/e2e_test.go
JAORMX: Right, now that I think about it, the `defer` didn't make sense here as it'll get run at the end of the whole function. We really want to do a `git checkout` at the end of the cluster-scoped tests. Subsequently, we want to do the other git checkout at the end of the namespace scoped tests. I think what's resulting here is that the `git checkout` doesn't happen, which means that we don't reset the manifest. Are we running only cluster-scoped tests as a result? Seems to me like the `sed` calls wouldn't work if we're not replacing the right values.

saschagrunert: Sounds good, I changed to checkout within `cleanupOperator()` and added a new call for the namespaced one. Let's see if that works.

test/e2e_test.go
```diff
@@ -156,6 +154,7 @@ func (e *e2e) TestSecurityProfilesOperator() {
 			tc.fn(nodes)
 		})
 	}
+	e.cleanupOperator(manifest)
```
JAORMX, line 157: I think it should be `namespaceManifest` instead of `manifest` here.

saschagrunert: Yep, fixed :+1:

JAORMX:
Let's see how this goes in CI. I'll tag it once it's done.

hack/ci/Vagrantfile
```diff
@@ -28,12 +28,11 @@ Vagrant.configure("2") do |config|
       echo 'export PATH=$PATH:/usr/local/go/bin' >> /etc/profile
       echo 'export GOPATH="$HOME/go"' >> /etc/profile
       echo 'export GOBIN="$GOPATH/bin"' >> /etc/profile
-      echo '. /etc/profile' >> $HOME/.bashrc
 
       dnf install -y \
         conntrack \
         container-selinux \
-        golang-go \
+        git \
```
saschagrunert, line 35: Maybe this is the issue :)

saschagrunert:
@JAORMX tests are green now

JAORMX (approved):
/lgtm

yay!

PR #509, opened 2021-06-29
Cache pods and container IDs in log enricher

saschagrunert:
#### What type of PR is this?

/kind feature

We now use a cache and a default timeout of 6 hours to cache the
container info as well as the PIDs. This should speedup the enricher in
general.

None

#### Does this PR have test?

None

None

```release-note
Added container ID caching to log enricher for performance reasons.
```


internal/pkg/daemon/enricher/container.go
```diff
@@ -77,7 +88,7 @@ func (e *Enricher) getNodeContainers(logger logr.Logger, nodeName string) (map[s
 						continue
 					}
 
-					containers[rawContainerID] = containerInfo{
+					containers[rawContainerID] = &containerInfo{
```
JAORMX, line 91: why don't we update the cache already here? This will cache all the container infos that have been found already, wouldn't it?

internal/pkg/daemon/enricher/enricher.go
```diff
@@ -27,23 +30,39 @@ import (
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
 )
 
+const defaultCacheTimeout time.Duration = 6 * time.Hour
```
JAORMX, line 33: Given the ephemeral nature of containers, maybe we should choose a lower cache timeout. e.g. only a couple of hours maybe? or even less, what do you think? We should document our reasoning in comments here.

JAORMX (changes requested):
Good stuff! I actually didn't know about this caching library, but it looks quite neat! Let's at least add some comments about our cache TTL value reasoning.

internal/pkg/daemon/enricher/container.go
saschagrunert: Yes I think this makes sense, changed as suggested :+1:

internal/pkg/daemon/enricher/enricher.go
saschagrunert: Yeah I think we should reduce it, I've chosen 1 hour. I also added a comment that it's nothing more than a rough guess :man_shrugging:

JAORMX: Right, at least now people won't wonder why a value was chosen. Maybe one day we'll have enough usage and data for the operator to choose a value based on data :)

JAORMX (approved):
/lgtm

Issue #510, opened 2021-06-29
Recording profiles with the log enricher


saschagrunert:
#### What would you like to be added:

It should be possible to record profiles by using the log enricher, while preserving a good usability.

#### Why is this needed:

Right now the operator supports recording profiles by using the `ProfileRecording` CRD as well as the OCI seccomp BPF hook. This requires CRI-O as well as the hook to be pre-installed and configured on every node running the recording.

However, the log enricher supports CRI-O and containerd as well as relies only on auditd. Adding recording support to the enricher may conflict with the existing recording feature and should seamless integrate into the operator. What we could do, is:

- Add a new field to the [`ProfileRecordingSpec`](https://github.com/kubernetes-sigs/security-profiles-operator/blob/d4ee3f9a931763072f1a7a89bc3ad448b5911ccf/api/profilerecording/v1alpha1/profilerecording_types.go#L30) to indicate that we either want to record by using the hook or by using the logs
- Based on the new field, a Recording CRD would have to ship with an [audit log profile](https://github.com/kubernetes-sigs/security-profiles-operator/blob/d4ee3f9a931763072f1a7a89bc3ad448b5911ccf/examples/seccompprofile.yaml#L11-L18)
- Reuse the existing recording webhook to add the audit log profile to the target pod (error if there is a profile already added by the user)
- Let the enricher take the recording profile into account to stream the syscalls to the GRPC server
- Let the GRPC server collect the syscalls per container/pod until the recording gets deleted
- Reconcile a SeccompProfile CRD on recording deletion

#### User story covered

> As an infrastructure SRE, Kirsti wants to be able to easily correlate audit log entries related to security profiles to specific containers/pods. ([link](https://github.com/kubernetes-sigs/security-profiles-operator/blob/master/doc/user-stories.md#as-an-infrastructure-sre-kirsti-wants-to-be-able-to-easily-correlate-audit-log-entries-related-to-security-profiles-to-specific-containerspods))

jhrozek:
I think this would work well. I was thinking recently about the recording use-case for SELinux and I had a similar flow in mind.
Some comments inline -- they mostly focus on the SELinux case, so hopefully I'm not hijacking the ticket too much.

> #### What would you like to be added:
> 
> It should be possible to record profiles by using the log enricher, while preserving a good usability.
> #### Why is this needed:
> 
[...]

PR #513, opened 2021-06-30
Add support for log enricher profile recording

saschagrunert:
#### What type of PR is this?

/kind feature

Implements profile recording by using the log enricher.

It also fixes a couple of issues with the log enricher:

- We store the `containerID` here, which should be `rawContainerID`: https://github.com/kubernetes-sigs/security-profiles-operator/blob/cd71f464ca8606ff427bf8a2826c682d92053a28/internal/pkg/daemon/enricher/container.go#L100
- We re-init the local `containers` for every loop: https://github.com/kubernetes-sigs/security-profiles-operator/blob/cd71f464ca8606ff427bf8a2826c682d92053a28/internal/pkg/daemon/enricher/container.go#L66
- We can completely remove the `containers` map and only rely on the cache to stay safe: https://github.com/kubernetes-sigs/security-profiles-operator/blob/cd71f464ca8606ff427bf8a2826c682d92053a28/internal/pkg/daemon/enricher/container.go#L58

PoC of https://github.com/kubernetes-sigs/security-profiles-operator/issues/510

#### Does this PR have test?

None

Possible follow-upds:
- Update the docs to cover the new feature and `ProfileRecording` field
- Use GRPC streaming for the syscalls (open a stream on the beginning of the enricher and use that one for pushing the syscalls)

```release-note
Added support for recording profiles by using the log enricher.
```


saschagrunert:
I'll update the docs in a follow-up PR.

internal/pkg/daemon/enricher/container.go
```diff
@@ -111,12 +114,13 @@ func (e *Enricher) getContainerInfo(
...
-	if !found {
-		return nil, errors.Wrap(err, "no container info for container ID")
+	if info, err := e.infoCache.Get(
+		containerID,
+	); !errors.Is(err, ttlcache.ErrNotFound) {
+		return info.(*containerInfo), nil
 	}
 
-	return info, err
+	return nil, errors.New("found nothing")
```
JAORMX, line 123: `found nothing` isn't a very descriptive error message, we probably want to keep using something along the lines of `"no container info for container ID"`

saschagrunert: Fixed :+1:

JAORMX (approved):
/lgtm

PR #528, opened 2021-07-06
Rename spod `spod` to `default`

saschagrunert:
#### What type of PR is this?

/kind cleanup

It may be confusing for users to see so many "spod"s around, so we now
rename the standard configuration to `default` to make things more
clear and explicit.

None
#### Does this PR have test?
None

None

```release-note
Changed default daemon configuration to be "default" instead of "spod".
```

### DIFF
--- a/internal/pkg/manager/spod/spod_controller.go
+++ b/internal/pkg/manager/spod/spod_controller.go
@@ -376,7 +376,6 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 ) *appsv1.DaemonSet {
 	newSPOd := r.baseSPOd.DeepCopy()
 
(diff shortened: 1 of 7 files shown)


JAORMX (approved):

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -376,7 +376,6 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 ) *appsv1.DaemonSet {
 	newSPOd := r.baseSPOd.DeepCopy()
 
-	newSPOd.SetName(cfg.GetName())
```
saschagrunert, line 379: I think they're linked by their name on purpose.

PR #533, opened 2021-07-09
Preserve reason if tail fails

saschagrunert:
#### What type of PR is this?

/kind cleanup

We should still track the reason if the tail of the log file files, which is now the case.

None
#### Does this PR have test?
Yes

Prow e2e tests are broken until https://github.com/kubernetes/test-infra/pull/22823 got merged.

```release-note
None
```


JAORMX (approved):
/lgtm

PR #535, opened 2021-07-12
Check for correct enricher pod state in e2e test

saschagrunert:
#### What type of PR is this?

/kind failing-test

The test pod never completes which means we have to grep for `Running`
as well as abort correctly if the timeout got reached.

None
#### Does this PR have test?
Yes

None

```release-note
None
```


JAORMX (approved):
/lgtm

PR #544, opened 2021-07-20
Use libseccomp for syscall ID to name conversion

saschagrunert:
#### What type of PR is this?

/kind cleanup

The library already features a conversion which is also used by the OCI
seccomp BPF hook. This way we do not have to maintain our own syscall
types any more.

None
#### Does this PR have test?
None

None

```release-note
None
```


saschagrunert:
@jhrozek PTAL

internal/pkg/daemon/enricher/syscall_unsupported.go
```diff
@@ -0,0 +1,28 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package enricher
+
+import "errors"
+
+var errUnsupoprtedPlatform = errors.New("unsupoprted platform")
```
jhrozek, line 23: Nit: typo. Feel free to include the fix in another PR if you're working on one, this patch look good otherwise.

jhrozek (approved):
Looks good except for the typo in the error message. Not really worth waiting for another CI run, let's merge now and fix the typo later..
/lgtm

internal/pkg/daemon/enricher/syscall_unsupported.go
saschagrunert: Oops, I'll follow-up on that.

PR #546, opened 2021-07-21
Retry GRPC connection in enricher

saschagrunert:
#### What type of PR is this?

/kind cleanup

The connection will mostly fail the first time the enricher tries to
start. To avoid the annoying container crash we now retry a few times
before actually killing the container.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/internal/pkg/daemon/enricher/enricher.go
+++ b/internal/pkg/daemon/enricher/enricher.go
@@ -17,6 +17,7 @@ limitations under the License.
 package enricher
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/pkg/daemon/enricher/enricher.go
```diff
@@ -76,23 +79,37 @@ func (e *Enricher) Run() error {
...
+	if err := util.Retry(func() (err error) {
+		conn, cancel, err = e.impl.Dial()
+		if err != nil {
+			return errors.Wrap(err, "connecting to local GRPC server")
+		}
+		client := api.NewSecurityProfilesOperatorClient(conn)
+
+		metricsAuditIncClient, err = e.impl.MetricsAuditInc(client)
+		if err != nil {
+			return errors.Wrap(err, "create metrics audit client")
```
jhrozek, line 99: Shouldn't we close conn and call cancel also here..

internal/pkg/daemon/enricher/enricher.go
```diff
@@ -76,23 +79,37 @@ func (e *Enricher) Run() error {
...
+		}
+
+		recordSyscallClient, err = e.impl.RecordSyscall(client)
+		if err != nil {
```
jhrozek, line 103: ...and here to avoid leaking them when iterating?

jhrozek (commented):
Two questions inline, mostly looks good though

internal/pkg/daemon/enricher/enricher.go
saschagrunert: Uh yeah, fixed as suggested :+1:

jhrozek (approved):
/lgtm
thank you, this looks good

PR #549, opened 2021-07-22
Use -trimpath for builds

saschagrunert:
#### What type of PR is this?

/kind cleanup

The flag removes all file system paths from the resulting executable.
Instead of absolute file system paths, the recorded file names will
begin with either "go" (for the standard library), or a module
path@version (when using modules), or a plain import path (when using
GOPATH).

None
#### Does this PR have test?
None

None

```release-note
None
```


jhrozek (approved):
/lgtm

PR #550, opened 2021-07-22
Add resource requests/limits to manager and webhook

saschagrunert:
#### What type of PR is this?

/kind cleanup

Beside setting some sane defaults (tested locally), we now also remove
the CPU limits from the spod to avoid having negative side effects with
CPU throttling.

None
#### Does this PR have test?
None

None

```release-note
Removed CPU limits from SPOD and added resource requests/limits to manager and webhook.
```

### DIFF
--- a/deploy/base/manager_deployment.yaml
+++ b/deploy/base/manager_deployment.yaml
@@ -26,6 +26,12 @@ spec:
           securityContext:
             readOnlyRootFilesystem: true
             allowPrivilegeEscalation: false
+          resources:
+            requests:
+              memory: 32Mi
+              cpu: 250m
+            limits:
+              memory: 64Mi
           env:
             - name: RELATED_IMAGE_SELINUXD
               value: quay.io/jaosorior/selinuxd
--- a/deploy/base/webhook_deployment.yaml
+++ b/deploy/base/webhook_deployment.yaml
@@ -27,6 +27,12 @@ spec:
           securityContext:
             readOnlyRootFilesystem: true
             allowPrivilegeEscalation: false
+          resources:
+            requests:
+              memory: 32Mi
+              cpu: 250m
+            limits:
+              memory: 64Mi
           volumeMounts:
           - name: cert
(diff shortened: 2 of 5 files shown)


jhrozek (commented):
Can you elaborate on what issues were you seeing with the CPU throttling? Or what benefit do you see with always letting the workloads use as much CPU as they need?

saschagrunert:
> Can you elaborate on what issues were you seeing with the CPU throttling? Or what benefit do you see with always letting the workloads use as much CPU as they need?

I'm not seeing the issue on my machine or in our test setups. I know that customers had issues with settings CPU limits. The feature is probably a bit under-documented, but I think I found some blog posts around the main issue:

- https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718
- https://amixr.io/blog/what-wed-do-to-save-from-the-well-known-k8s-incident/
- https://erickhun.com/posts/kubernetes-faster-services-no-cpu-limits/

jhrozek:
> > Can you elaborate on what issues were you seeing with the CPU throttling? Or what benefit do you see with always letting the workloads use as much CPU as they need?
> 
> I'm not seeing the issue on my machine or in our test setups. I know that customers had issues with settings CPU limits. The feature is probably a bit under-documented, but I think I found some blog posts around the main issue:
> 
>     * https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718
> 
>     * https://amixr.io/blog/what-wed-do-to-save-from-the-well-known-k8s-incident/
> 
>     * https://erickhun.com/posts/kubernetes-faster-services-no-cpu-limits/

Thank you. I'll ack the patch. The reason I asked was that in OpenShift at least we've been asked to set CPU limits for everything to support a use-case of either small number of nodes or even a single node. But I guess this makes sense from the upstream perspective and we can always change the requests and limits, it's just a number to set or reset.

jhrozek (approved):
/lgtm

PR #595, opened 2021-09-21
CI: Try using ubuntu/focal64 box as a workaround for failing CI

jhrozek:
#### What type of PR is this?
/kind bug

our CI seems to be hitting

https://bugs.launchpad.net/cloud-images/+bug/1939580

which causes the shared folder to not be mounted and eventually
manifests as the hack/ci/e2e-ubuntu.sh not found, because it's not
copied over to the shared folder.

which is resolved by upgrading to a recent kernel (5.4.0.84). The
generic/ubuntu64 image we have been using appears to be using 5.4.0.81.

Unfortunately, the official ubuntu boxes only ship for virtualbox, which
means that we can't use them in general. But at least let's use them for
our CI, anyone running locally would probably not run into the issue
with the VB shared folders anyway.

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -70,7 +70,7 @@ dependencies:
       match: fedora
 
   - name: e2e-ubuntu
-    version: 2004
+    version: focal64
     refPaths:
     - path: hack/ci/Vagrantfile-ubuntu
       match: ubuntu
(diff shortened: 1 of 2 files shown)


jhrozek:
@saschagrunert I didn't find a way to provide different boxes for different providers. I tried `vm.box` under the per-provider section but that was just giving me an error locally with libvirt:
```
Bringing machine 'default' up with 'libvirt' provider...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

Libvirt Provider:
* The following settings shouldn't exist: box
```
Do you have another idea how to set a per-provider box?

jhrozek:
Superseded by #599

PR #633, opened 2021-11-03
Remove no_bpf build tag

saschagrunert:
#### What type of PR is this?

/kind cleanup

Target is to get rid of the overwritten go module and always build-in BPF support.

None

#### Does this PR have test?

None

None

```release-note
None
```

### DIFF
--- a/Dockerfile.ubi
+++ b/Dockerfile.ubi
@@ -16,9 +16,24 @@ FROM registry.fedoraproject.org/fedora-minimal:latest AS build
 USER root
 WORKDIR /work
 
-RUN microdnf install -y make golang-bin libseccomp-static glibc-static
+RUN microdnf install -y \
+        bzip2 \
+        curl-devel \
+        g++ \
+        glibc-static \
+        golang-bin \
+        libarchive-devel \
+        libmicrohttpd-devel \
+        libseccomp-static \
+        m4 \
+        make \
+        sqlite-devel \
+        tar \
+        zlib-static
 
 ADD . /work
+RUN hack/install-libelf.sh
+RUN hack/install-libbpf.sh
 RUN mkdir -p build
 
 RUN make
--- a/hack/install-libelf.sh
+++ b/hack/install-libelf.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+# Copyright 2021 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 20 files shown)


hack/install-libelf.sh
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+
+VERSION=0.185
+curl -sSfL --retry 5 --retry-delay 3 \
+    "https://sourceware.org/elfutils/ftp/$VERSION/elfutils-$VERSION.tar.bz2" -o- |
+    tar xfj -
+pushd "elfutils-$VERSION"
```
pjbgf, line 22: This package provide digital signature and sha512 checksum, it would be great to verify the integrity or provenance (preferred) of it before extracting.

https://sourceware.org/elfutils/ftp/0.185/elfutils-0.185.tar.bz2.sig
https://sourceware.org/elfutils/ftp/0.185/sha512.sum

Dockerfile.ubi
```diff
@@ -16,9 +16,24 @@ FROM registry.fedoraproject.org/fedora-minimal:latest AS build
...
+        libmicrohttpd-devel \
+        libseccomp-static \
+        m4 \
+        make \
+        sqlite-devel \
+        tar \
+        zlib-static
 
 ADD . /work
+RUN hack/install-libelf.sh
```
saschagrunert, line 35: @jhrozek Fedora does not ship libelf static any more since a couple of releases. I think we cannot avoid building it on our own here, right?

pjbgf (commented):
nit around remote package being used without further checks.

hack/install-libelf.sh
saschagrunert: Fixed :)

Dockerfile.ubi
jhrozek: That's right. I guess we could ask for it, but there is no guarantee the request would be granted. Moreover, thinking about SPO on OCP and therefore on RHCOS, the selection of build libraries would be even more restricted there. I'm also wondering about multi-architecture support. Not to spill too many internal things, but I've recently worked on porting some other operators to ppc64 and Z -- if the bpf support is not fully supported on those architectures, I was wondering that we would still either use another method there or just no recording. Either of these would mean disabling the built-in bpf recorder.

pjbgf:
nice one @saschagrunert 

/approve
/lgtm

PR #635, opened 2021-11-03
Rely on BPF PID tracking rather than mount namespaces

saschagrunert:
#### What type of PR is this?

/kind bug

We now only track the PIDs together with their cgroup/profile rather
than the OCI runtime itself via the mount namespace.

None
#### Does this PR have test?
None

Refers to https://github.com/kubernetes-sigs/security-profiles-operator/pull/618#discussion_r739703260

```release-note
None
```


internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -93,8 +60,17 @@ int sys_exit(struct trace_event_raw_sys_exit * args)
...
         bpf_ringbuf_submit(event, 0);
+
+        static const char init[MAX_SYSCALLS];
+        bpf_map_update_elem(&syscalls, &pid, &init, BPF_ANY);
+
+        u8 * const value = bpf_map_lookup_elem(&syscalls, &pid);
+        if (!value) {
+            // Should not happen, we updated the element straight ahead
+            return 0;
+        }
```
pjbgf, line 71: nit: it might be good to log this so we can debug as it beds in.

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -83,8 +47,11 @@ int sys_exit(struct trace_event_raw_sys_exit * args)
         bpf_map_update_elem(&comms, &pid, &comm, BPF_ANY);
     }
 
-    // Update the PIDs and send a new event for new PIDs
-    if (bpf_map_lookup_elem(&pids, &pid) == NULL) {
+    // Update the syscalls
+    u8 * const syscall_value = bpf_map_lookup_elem(&syscalls, &pid);
+    if (syscall_value) {
+        syscall_value[id] = 1;
```
pjbgf, line 53: nit: is `id` here the syscall Id? It may be worth renaming it for clarity.

pjbgf (commented):
two small nits but still LGTM

/approve
/lgtm

PR #638, opened 2021-11-04
Make cgroup tracking more robust

saschagrunert:
#### What type of PR is this?

/kind cleanup

Testing with different runtimes reveled that we cannot skip the
"crio-conmon-" cgroup because it may be temporarily the target cgroup to
track. To still filter out `conmon` as a process, we now use a list of
filtered comms during syscall accumulation.

None
#### Does this PR have test?
None

None

```release-note
None
```


saschagrunert:
@pjbgf PTAL

jhrozek (approved):
/lgtm

PR #649, opened 2021-11-09
Bump golangci to 1.43.0

jhrozek:
#### What type of PR is this?
/kind cleanup

Let's stay up2date

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```


saschagrunert (approved):

jhrozek:
@saschagrunert @pjbgf PTAL.

I'm not very happy myself with adding so many new skip files listed one by one and also not super happy about following master for the the boilerplate verify script. Makes me wonder if we should stay on an older version of golangci..

pjbgf:
> I'm not very happy myself with adding so many new skip files listed one by one and also not super happy about following master for the the boilerplate verify script. Makes me wonder if we should stay on an older version of golangci..

@jhrozek I am more concerned with following `master` for the boilerplate script. I would probably wait until a new version is cut so we can pin it.
I agree that the cost vs benefit of upgrading `golangci` does not look appealing - mainly for the point above.

jhrozek:
> > I'm not very happy myself with adding so many new skip files listed one by one and also not super happy about following master for the the boilerplate verify script. Makes me wonder if we should stay on an older version of golangci..
> 
> @jhrozek I am more concerned with following `master` for the boilerplate script. I would probably wait until a new version is cut so we can pin it. I agree that the cost vs benefit of upgrading `golangci` does not look appealing - mainly for the point above.

Oops, the patch was merged. I'm glad for the feedback, though and had a similar concern. Feel free to post a revert so that this gets properly discussed.

pjbgf:
@jhrozek that's fine. I think we can just wait for the `repo-infra` to cut a version and then we pin it back to it.

PR #651, opened 2021-11-09
Update btf and integrate into CI

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now support ARM for kernels not exposing the btf file. The generated
btf will be part of the CI runs as well.

None
#### Does this PR have test?
None

None

```release-note
None
```


hack/update-btf
```diff
@@ -1,18 +1,28 @@
...
 set -euo pipefail
 
+BTFHUB_COMMIT=807c5c41894143513100ac58fa07e72df221baa1
 TEMP_DIR=$(mktemp -u)
 ROOT_DIR=$(git rev-parse --show-toplevel)
 BPF_OBJECT="$ROOT_DIR/build/recorder.bpf.o"
 BTF_DIR="$ROOT_DIR/internal/pkg/daemon/bpfrecorder/btf"
 
 rm -rf "$BTF_DIR"
-git clone --depth=1 https://github.com/aquasecurity/btfhub "$TEMP_DIR"
```
pjbgf, line 10: Do you mind clarifying the removal of `--depth=1`? Not sure the benefit of removing it, and this may make such operation speedier.

saschagrunert: Yes the idea was to pin the BTF hub to a certain commit to not accidentally break the CI if it changes.

pjbgf: That make sense, thank you.

PR #656, opened 2021-11-10
Filter system mount namespace

saschagrunert:
#### What type of PR is this?

/kind cleanup

This allows us to filter processes like conmon completely automatically
because containers usually do not run in system mount namespace. This
also increases the performance when it's about to skip PIDs.

None
#### Does this PR have test?
None

None

```release-note
None
```


saschagrunert:
@pjbgf PTAL

PR #657, opened 2021-11-10
Make log verbosity usable

saschagrunert:
#### What type of PR is this?

/kind feature

We hack around klog to make the log verbosity actually usable either by
using the `-V` flag or the `SPO_VERBOSITY` environment variable.

None
#### Does this PR have test?
None

None

```release-note
Added CLI flag `-V` and environment variable parsing `SPO_VERBOSITY` to set the logging verbosity.
```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"context"
+	"flag"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


jhrozek (commented):
I think the code is good, but I also think it would be nice to touch up installation-usage.md. There is a troubleshooting section that might benefit from having this information.

saschagrunert:
> I think the code is good, but I also think it would be nice to touch up installation-usage.md. There is a troubleshooting section that might benefit from having this information.

How about adding a new feature in the spod config to enable debug logs? This way the operator can take care of adding the env to the daemon set.

I can follow up on that if we think it's a good idea.

jhrozek:
On Thu, Nov 11, 2021 at 04:12:26AM -0800, Sascha Grunert wrote:
> > I think the code is good, but I also think it would be nice to touch up installation-usage.md. There is a troubleshooting section that might benefit from having this information.
> 
> How about adding a new feature in the spod config to enable debug logs? This way the operator can take care of adding the env to the daemon set.
> 
> I can follow up on that if we think it's a good idea.

I like this!

saschagrunert:
> I like this!

Thanks, can we merge this one then? :upside_down_face:

jhrozek (approved):
/lgtm

PR #663, opened 2021-11-11
Early give-up on short-lived processes

saschagrunert:
#### What type of PR is this?

/kind bug

If we're not able to retrieve the container because the process is too
short lived, then we give up retrieving the syscalls from the profile
recorder. We also have to protect the BPF maps for concurrent access to
avoid crashes.

None
#### Does this PR have test?
None

None

```release-note
None
```


JAORMX (approved):
/lgtm

PR #671, opened 2021-11-15
Make seccomp profile architectures non pointers

saschagrunert:
#### What type of PR is this?

/kind api-change

This makes it easier for us to maintain the string slice rather than having pointers to strings.

None
#### Does this PR have test?
None

None

```release-note
Change seccomp profile type `Architectures` to `[]Arch` from `[]*Arch`
```


pjbgf (commented):
/lgtm

PR #675, opened 2021-11-16
Selinux profiles v1alpha2: A user-friendly and raw SelinuxProfile CRDs

JAORMX:
#### What type of PR is this?

/kind feature
/kind design
/kind api-change

This introduces a more user-friendly SELinux profile. The intent is to
reduce the knowledge that folks need to have about CIL itself, and
instead focus on a more kube-native representation.

This has the following advantages:

* Easier to validate
* Faster to parse
* Easier to read

Underneath, this will generate CIL files that are consumed by the daemon
and eventually installed on the host.

None

#### Does this PR have test?

Yes

Note that at the moment the SELinux CRDs share the metric, we need to decide how we'll handle this.
e.g. will we use the same metric with different labels or will we have different metrics for the different CRDs?

There currently is a very simple validation for SELinux profiles that merely checks that the inherit references
are allowed (already a better position than the raw profiles). However more validations could be added.
e.g. Do referenced labels in the profile exist? This can be done in a subsequent PR.

```release-note
a v1alpha2 version of the SelinuxProfile object has been introduced. This
removes the raw CIL from the object itself and instead adds a simple policy
language to ease the writing and parsing experience.

Alongside, a RawSelinuxProfile object was also introduced. This contains a wrapped
and raw representation of the policy. This was intended for folks to be able to take
their existing policies into use as soon as possible. However, on validations are done here.
```


JAORMX:
Pull-request updated, HEAD is now 50e97a85764cbe9580bc943f75b8c1988b76c2f0

JAORMX:
Pull-request updated, HEAD is now 741e6ef79ca2c42c9bd6cd51b6c9e7f8ca798583

jhrozek:
In general I think this would be a really nice usability improvement. Before I review the code in detail, one question:

Do you see us supporting the CIL based policy as well at the same time? I personally think we should, at least to support policies created by other tools than SPO or hand-written policies. This would be to support the case where someone has an existing policy and just wants to use SPO to manage it. I agree that we should promote this format and create our policies in it, but dropping CIL support altogether sounds like we would drop some existing use-cases as well.

JAORMX:
> In general I think this would be a really nice usability improvement. Before I review the code in detail, one question:
> 
> Do you see us supporting the CIL based policy as well at the same time? I personally think we should, at least to support policies created by other tools than SPO or hand-written policies. This would be to support the case where someone has an existing policy and just wants to use SPO to manage it. I agree that we should promote this format and create our policies in it, but dropping CIL support altogether sounds like we would drop some existing use-cases as well.

I don't think our current CRD supported many cases outside of a single process (container) as we were using it. We either way wrapped everything in a block.

I think it would be more useful to provide yet another CRD... something called `RawSelinuxPolicy` that would just contain the raw CIL and allow you to install them directly. That CIL would be unwrapped and allow extended use-cases. What do you think?

JAORMX:
Pull-request updated, HEAD is now 22ac0c21af9fc2b744bdf23306955b505aecf483

JAORMX:
Pull-request updated, HEAD is now f23793436a40e43c49470527f238b096a6a67c40

JAORMX:
Pull-request updated, HEAD is now 0db590a9465c7022b874c84fd29d62a0c23313d3

JAORMX:
Pull-request updated, HEAD is now 999c014a0245b80c882b7bf971aaaf229d14c821

JAORMX:
Pull-request updated, HEAD is now cc9c88aaf0e7a794382a7708264afb1c1a0bd800

JAORMX:
Pull-request updated, HEAD is now 246483b9f2945fb8fe38f2d45140019dc1ff425a

JAORMX:
Pull-request updated, HEAD is now 22c324228b779244e5e0baec10ea366d8886ad5a

JAORMX:
Pull-request updated, HEAD is now 546e64e04353ed45bf0b79bf8797b4d786adba7c

JAORMX:
Pull-request updated, HEAD is now 418bafff94dd6e2d45064d19c995eed4149e2f47

JAORMX:
Pull-request updated, HEAD is now e110d86ce79b73a25bb3779a8eefd1d8049e9c55

JAORMX:
Pull-request updated, HEAD is now 0f13f1c347e4a14f02f0b4829bfec957ec55c29e

JAORMX:
Pull-request updated, HEAD is now d75d34c20351069fd53ff8c30f28d1fb1619b765

JAORMX:
Pull-request updated, HEAD is now b8732a66ec21e3ea48b8218d11b661a26d87080b

JAORMX:
Pull-request updated, HEAD is now 8a32987f8f15f4fd1f1d4afb264879317282f7de

JAORMX:
Pull-request updated, HEAD is now f3bd06e372d4878e39ea2dcf0402bb4f24f695b3

JAORMX:
Pull-request updated, HEAD is now 517445ff197d1bffec68167d99976e25260dde10

jhrozek:
On Tue, Nov 16, 2021 at 10:58:45PM -0800, Juan Osorio Robles wrote:
> > In general I think this would be a really nice usability improvement. Before I review the code in detail, one question:
> > 
> > Do you see us supporting the CIL based policy as well at the same time? I personally think we should, at least to support policies created by other tools than SPO or hand-written policies. This would be to support the case where someone has an existing policy and just wants to use SPO to manage it. I agree that we should promote this format and create our policies in it, but dropping CIL support altogether sounds like we would drop some existing use-cases as well.
> 
> I don't think our current CRD supported many cases outside of a single process (container) as we were using it. We either way wrapped everything in a block.
> 
> I think it would be more useful to provide yet another CRD... something called `RawSelinuxPolicy` that would just contain the raw CIL and allow you to install them directly. That CIL would be unwrapped and allow extended use-cases. What do you think?

Either that or a field in the new CRD. I care less about how it's done
and more about it being possible to use policies recorded/created
outside SPO.

JAORMX:
Pull-request updated, HEAD is now 517445ff197d1bffec68167d99976e25260dde10

JAORMX:
Pull-request updated, HEAD is now 08e1712a4f41e8dac1ee944ecd66a901f4df9c80

JAORMX:
> On Tue, Nov 16, 2021 at 10:58:45PM -0800, Juan Osorio Robles wrote: > In general I think this would be a really nice usability improvement. Before I review the code in detail, one question: > > Do you see us supporting the CIL based policy as well at the same time? I personally think we should, at least to support policies created by other tools than SPO or hand-written policies. This would be to support the case where someone has an existing policy and just wants to use SPO to manage it. I agree that we should promote this format and create our policies in it, but dropping CIL support altogether sounds like we would drop some existing use-cases as well. I don't think our current CRD supported many cases outside of a single process (container) as we were using it. We either way wrapped everything in a block. I think it would be more useful to provide yet another CRD... something called `RawSelinuxPolicy` that would just contain the raw CIL and allow you to install them directly. That CIL would be unwrapped and allow extended use-cases. What do you think?
> Either that or a field in the new CRD. I care less about how it's done and more about it being possible to use policies recorded/created outside SPO.

Makes sense. I'll create a new CRD. This would actually allow folks to take Udica podicies directly as opposed to having to "unwrap" them as they need to do now.

JAORMX:
Pull-request updated, HEAD is now 4031985733ea59c56fa80bab3c762bdb7c089a84

JAORMX:
Pull-request updated, HEAD is now af0a5d1bdfd7bbe9873784c02462ac2c0b470d3b

JAORMX:
Pull-request updated, HEAD is now 5ffe3dd91d1c1821d77405b3e57dae96986a902a

JAORMX:
Pull-request updated, HEAD is now d0f25975a2d8bdc096e84dc6a7e2a307495ff544

JAORMX:
Pull-request updated, HEAD is now 37182788b2cbf83ef376683f820fa9419105b212

JAORMX:
Pull-request updated, HEAD is now 9de5ed6cef8d4a06fad18f45e0c57f26ceccac2c

JAORMX:
Pull-request updated, HEAD is now 5cf1af1292c3e02b4ed684d515a2327e59150147

JAORMX:
Pull-request updated, HEAD is now 0272d62e83f419012110be51ef17d1584342337b

JAORMX:
Pull-request updated, HEAD is now 9c132bfed92d367016ead880db1f3905ec4c533f

JAORMX:
Pull-request updated, HEAD is now 470d35e1d83239f8867db720fad199e56374ea04

JAORMX:
Pull-request updated, HEAD is now 31899f76b7078bffac35095d06eb0da6381ef9f0

JAORMX:
Pull-request updated, HEAD is now 2a05f8bf6d6a8a1a62cf800c62d6532692df6d1c

api/selinuxprofile/v1alpha2/selinuxprofile_types.go
```diff
@@ -14,22 +14,58 @@ See the License for the specific language governing permissions and
...
 // Ensure SelinuxProfile implements the StatusBaseUser interface.
 var _ profilebasev1alpha1.StatusBaseUser = &SelinuxProfile{}
 
+type PolicyRef struct {
+	// The Kind of the policy that this inherits from.
+	// Can be one of the policy objects: SelinuxPolicy or RawSelinuxPolicy.
+	// Or "System" if an already installed policy will be used.
+	// The allowed "System" policies are available in the
+	// SecurityProfilesOpertorDaemon instance.
+	// +kubebuilder:default="System"
```
jhrozek, line 40: Would it make sense to allow kubebuilder's enum feature to list the allowed kinds?

deploy/namespace-operator.yaml
```diff
@@ -675,6 +690,113 @@ status:
...
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.7.0
+  creationTimestamp: null
+  labels:
+    app: security-profiles-operator
+  name: rawselinuxprofiles.security-profiles-operator.x-k8s.io
+spec:
+  group: security-profiles-operator.x-k8s.io
+  names:
+    kind: RawSelinuxProfile
```
jhrozek, line 703: Can you also add an example of this CRD to `examples`?

internal/pkg/daemon/selinuxprofile/common_controller.go
```diff
@@ -87,56 +79,62 @@ const (
...
 
-// Security Profiles Operator RBAC permissions to manage SelinuxProfile
 // nolint:lll
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles/finalizers,verbs=delete;get;update;patch
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=rawselinuxprofiles,verbs=get;list;watch;create;update;patch
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=rawselinuxprofiles/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=rawselinuxprofiles/finalizers,verbs=delete;get;update;patch
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=clusterrawselinuxprofiles,verbs=get;list;watch;create;update;patch
```
jhrozek, line 124: These types no longer exist right?

internal/pkg/daemon/selinuxprofile/common_controller.go
```diff
@@ -87,56 +79,62 @@ const (
...
+func (r *ReconcileSelinux) Healthz(*http.Request) error {
 	return nil
 }
 
-// Security Profiles Operator RBAC permissions to manage SelinuxProfile
 // nolint:lll
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles/finalizers,verbs=delete;get;update;patch
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=rawselinuxprofiles,verbs=get;list;watch;create;update;patch
```
jhrozek, line 121: Do you know why are these changes not reflected in the RBAC manifests?

internal/pkg/translator/obj2cil.go
```diff
@@ -0,0 +1,142 @@
...
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/daemon/common"
+)
+
+func Object2CIL(cli client.Client, sp *selxv1alpha2.SelinuxProfile) (string, error) {
+	cilbuilder := strings.Builder{}
+	cilbuilder.WriteString(getCILStart(sp))
+	for _, inherit := range sp.Spec.Inherit {
+		usage, err := getInheritName(cli, inherit, sp.GetNamespace())
+		if err != nil {
+			return "", err
```
jhrozek, line 40: Should we wrap errors in case we refactor the code later?

internal/pkg/translator/obj2cil.go
```diff
@@ -0,0 +1,142 @@
...
+	ttypeFinal := ttype.String()
+	if ttype == selxv1alpha2.AllowSelf {
+		ttypeFinal = sp.GetPolicyUsage()
+	}
+	uniquePerms := sets.NewString(perms...).List()
+	sort.Strings(uniquePerms)
+	return fmt.Sprintf("(allow process %s ( %s ( %s )))\n", ttypeFinal, tclass, strings.Join(uniquePerms, " "))
+}
+
+func getCILEnd(sp *selxv1alpha2.SelinuxProfile) string {
```
jhrozek, line 140: The parameter sp is unused

internal/pkg/translator/obj2cil.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	for idx := range spod.Spec.SelinuxOpts.AllowedSystemProfiles {
+		prof := spod.Spec.SelinuxOpts.AllowedSystemProfiles[idx]
+		if prof == ancestorRef.Name {
+			return prof, nil
+		}
+	}
+	msg := fmt.Sprintf("system profile %s not in SecurityProfilesOperatorDaemon's allow list", ancestorRef.Name)
+	return "", fmt.Errorf("%s: %w", msg, err)
```
jhrozek, line 118: why not wrapf?

internal/pkg/translator/obj2cil.go
```diff
@@ -0,0 +1,142 @@
...
+	namespace string,
+) (string, error) {
+
+	var ancestor selxv1alpha2.SelinuxProfileObject
+	kind := ancestorRef.Kind
+	key := types.NamespacedName{Name: ancestorRef.Name, Namespace: namespace}
+	switch ancestorRef.Kind {
+	// We default to SelinuxPolicy if Kind is left empty
+	case "SelinuxPolicy":
+		kind = "SelinuxPolicy"
```
jhrozek, line 84: Confused why is this line needed, isn't kind already `SelinuxPolicy`?

internal/pkg/translator/obj2cil.go
```diff
@@ -0,0 +1,142 @@
...
+		ancestor = &selxv1alpha2.SelinuxProfile{}
+	case "RawSelinuxPolicy":
+		ancestor = &selxv1alpha2.RawSelinuxProfile{}
+	default:
+		return "", fmt.Errorf("Unknown inherit kind for entry: %s/%s", ancestorRef.Kind, ancestorRef.Name)
+	}
+	err := cli.Get(context.Background(), key, ancestor)
+	if err != nil && errors.IsNotFound(err) {
+		message := fmt.Sprintf("couldn't find inherit reference %s/%s", kind, ancestorRef.Name)
+		return "", fmt.Errorf("%s: %w", message, err)
```
jhrozek, line 94: why not wrapf?

jhrozek (commented):
Left a couple of comments, but it's shaping up nicely!

api/profilebase/v1alpha1/profilebase.go
```diff
@@ -24,6 +24,11 @@ import (
 	secprofnodestatusv1alpha1 "sigs.k8s.io/security-profiles-operator/api/secprofnodestatus/v1alpha1"
 )
 
+const (
+	// Denotes a profile coming from the system itself.
+	SystemProfileAnnotation = "spo.x-k8s.io/profile-from-system"
```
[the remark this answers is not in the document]
jhrozek: Sorry, this is from an earlier version of the PR, please ignore

JAORMX:
Pull-request updated, HEAD is now 686cca102f1c4c33bc5ba78980f229328116a90f

JAORMX:
Pull-request updated, HEAD is now ed9dd08256d83e13e3574aca3e0d13febd5cc6d0

internal/pkg/translator/obj2cil.go
JAORMX: uuuh... what's the difference between `Errorf` with `%w` and `Wrapf` ?

JAORMX:
Pull-request updated, HEAD is now 84a5b6872720253918013e1d761d1d4c787cc90a

JAORMX:
Pull-request updated, HEAD is now c6db034308ddadc5aaea74654bc3619ef0f29a6a

JAORMX:
Pull-request updated, HEAD is now f8ad7977912f95de689e95d7a4c8990487b6240d

internal/pkg/translator/obj2cil.go
jhrozek: Functionally, I think none, except maybe unwrapping?. I thought that Wrap was the preferred way these days and I seem to remember that one of the linters complained when I did not use Wrap.
It's really not the biggest deal.

pjbgf:
Retriggering the github actions now after btf change was merged.

pjbgf:
@JAORMX unfortunately this will require a rebase. Otherwise it lgtm for me.

JAORMX:
@pjbgf rebased

JAORMX:
/retest

The Seccomp metrics test is flaky but unrelated

internal/pkg/translator/obj2cil.go
```diff
@@ -0,0 +1,76 @@
...
+		cilbuilder.WriteString(getCILInheritline(inherit.GetPolicyUsage()))
+	}
+	for ttype, tclassMap := range sp.Spec.Allow {
+		for tclass, perms := range tclassMap {
+			cilbuilder.WriteString(getCILAllowLine(sp, ttype, tclass, perms))
+		}
+	}
+	cilbuilder.WriteString(getCILEnd())
+	return cilbuilder.String()
+}
```
pjbgf, line 49: Any scenarios in which the plain-text input (name, namespace, ref policy) could be used to impact the resulting policy? Do we need any validation here to enforce that?

pjbgf (commented):
Small nit around input validation on Object2Cil conversion. Apart from that looks good to be and happy to approve it.

Waiting for further reviews before tagging it with lgtm.

internal/pkg/translator/obj2cil.go
JAORMX: @pjbgf that's a great point! The name and namespace have a limited character set, so we're safe in the sense that no policy injection or manipulation could be done there.

Inherits coming from another SelinuxProfile are also safe because that would be the object's name. Even if one would introduce a malformed name here, the operator would try to fetch the object beforehand and fail the validation, thus not getting here at all.

System objects could be validated better, as it seems we would accept whatever we put in the SPOD object itself.

Related to types, classes and allowed actions, we could definitely validate those better. As a subsequent PR, I'd like to get the `Validate()` function to explicitly verify that the action is being allowed on an existing label, else it should fail.

JAORMX:
/retest

CRD ordering flake...

JAORMX:
/retest

CRD ordering flake...

internal/pkg/daemon/selinuxprofile/common_controller.go
```diff
@@ -215,14 +226,26 @@ func (r *ReconcileSP) reconcilePolicy(
 		return reconcile.Result{Requeue: true}, nil
 	}
 
-	err = r.reconcilePolicyFile(sp, l)
+	if valErr := oh.Validate(); valErr != nil {
+		if err := nodeStatus.SetNodeStatus(context.TODO(), statusv1alpha1.ProfileStateError); err != nil {
+			r.metrics.IncSelinuxProfileError(reasonCannotUpdatePolicyStatus)
+			r.record.Event(sp, event.Warning(reasonCannotUpdatePolicyStatus, err))
+			return reconcile.Result{}, errors.Wrap(err, "setting node status to in progress")
```
jhrozek, line 233: typo in the errors.Wrap text

jhrozek (commented):
The only thing I found is the error message. If you end up getting a green test run, let's mark the PR as lgtm and merge, but if you end up having to re-run CI, please change the error message.

Otherwise, really impressive work!

api/selinuxprofile/v1alpha2/common.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
saschagrunert, line 2: nit: 2021

saschagrunert (approved):
Awesome! Just a single non blocking nit from my side.

PR #680, opened 2021-11-17
Load/Unload AppArmor profiles into hosts

### LINKED ISSUE #65: Create AppArmor profile for the operator
A tailor-made AppArmor profile to restrict functionality to the minimum required for the operator to run.

#### Why is this needed:
Better define the attack surface, giving users a clear view of what the operator can/cannot do.

/area security
/priority important-longterm

pjbgf:
#### What type of PR is this?

/kind feature

Implements the logic of loading/unloading profiles into cluster hosts providing a MVP for the AppArmor functionality.

Partially fixes #65 

#### Does this PR have test?

Tests will be included in follow-up PRs.

AppArmor profile management operations requires the container to operate as `privileged` and with `host pid ns` .
To counter balace the impact of such privilege requirements, any host level operation are being executed atomically within blocks such as:
https://github.com/pjbgf/security-profiles-operator/blob/83ecb88c77354afa5eaea34916017c92457c4123/internal/pkg/daemon/apparmorprofile/apparmor.go#L95

If we were to go ahead with such an approach, we could use this opportunity to reduce the complexity around the init containers. And then concentrate on ways to further protect supply chain attacks leveraging the container level privileges.

```release-note
`spod` can load and unload AppArmor profiles into clusters host servers.
`spod` now runs as `root` and `privileged`.
```


pjbgf:
Some tests are failing due to the lack of libapparmor in some environments. I will fix this once we agree the current approach is the way to go.

saschagrunert (approved):
LGTM, thanks for adding this @pjbgf! :pray:

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -426,11 +426,33 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
...
+		var userRoot int64 = 0
+		// a more privileged mode is required when apparmor is enabled
+		// TODO: review security model and provide a dynamic approach that can be case specific
+		templateSpec.Containers[bindata.ContainerIDDaemon].SecurityContext = &corev1.SecurityContext{
+			AllowPrivilegeEscalation: &truly,
+			Privileged:               &truly,
+			ReadOnlyRootFilesystem:   &falsely,
+			RunAsUser:                &userRoot,
+			RunAsGroup:               &userRoot,
+		}
```
pjbgf, line 446: @saschagrunert @jhrozek @JAORMX this ensures that spod is privileged/root only when apparmor is enabled - which would work as a feature switch and therefore cause no impact to existing execution levels for Seccomp/SELinux. PTAL

I will follow-up with some materials to foster discussion around current security model and a few ideas on changes we could adopt - but that will be a separate thread/PR.

Dockerfile.ubi
```diff
@@ -29,11 +29,20 @@ RUN microdnf install -y \
...
-        zlib-static
+        zlib-static \
+        autoconf \
+        automake \
+        libtool \
+        which \
+        dejagnu \
+        bison \
+        flex \
+        perl
```
jhrozek, line 40: It is fine to leave for now, but my plan with `Dockerfile.ubi` is to be as close as possible to what OpenShift will ship in the future. Are there any other plans with this Dockerfile?
The thing is that installing build tools and building stuff from scratch is a bit of a no-no for OpenShift.
Anyway, this is totally fine for this PR, just a food for thought (and something I wanted to discuss on our Tuesday meeting).

internal/pkg/manager/spod/spod_controller.go
jhrozek: I think this is fine, but have you considered instead the model we have used for SELinux where the privileged operations are executed by a separate workload, selinuxd in this case?
But as I said, I'm OK with this because it doesn't increase the set of privileges for neither the default case or the case we'll be enabling for openshift, but only for this one case.
Thank you for thinking about security!

jhrozek (commented):
Looks good to me. I'm not sure if you are waiting for more feedback from other developers as wel, so I won't use the looks-good-to-me tag. Please feel free to ping me on Slack if I should add it and the PR to be merged.

JAORMX:
My main concern is very similar to @jhrozek 's, by including the AppArmor dependencies into the main ContainerFiles, we are baking in assumptions and code that will not be used in SELinux cases. While I'm definitely up for including the AppArmor functionality, I'd say it should take a similar approach as we did for SELinux. Create a separate workload that will *just* do the Apparmor functionality. For SELinux, we did this with [selinuxd](https://github.com/JAORMX/selinuxd), however, the AppArmor pieces could be included in this repo too. We actually want to bring up the same topic for selinuxd... should we have it in this repo or should it have a repo of its own also in kubernetes-sigs?

However, the main idea remains: Let's have the AppArmor functionality be in its own container, thus keeping the dependencies (and the privilege) in there. Only when AppArmor is enabled do we spawn that container (as we do with SELinux).

@pjbgf what do you think?

Dockerfile.ubi
pjbgf: @jhrozek I was actually torn between fixing the errors I had supporting AppArmor for the OpenShift building process by a) leveraging conditional build and removing it completely or b) building libapparmor from source. I don't really have a good answer as to why I chose the latter. :smile: 

Users in RHEL/CentOS/Fedora will tend to use SELinux and users in Debian/Ubuntu/etc will tend to use AppArmor. Realistically I don't think we need to support both simultanebously, apart from cases in which a cluster have node pools that go across both spectrum.

For the time being I will just remove AppArmor from the OpenShift build through conditional build. If the need/use case arise in the future we can always retrofit.

hack/update-btf
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-BTFHUB_COMMIT=807c5c41894143513100ac58fa07e72df221baa1
+BTFHUB_COMMIT=9d50bd1fb3cc00fca591711f15702cec35347915
```
pjbgf, line 4: @saschagrunert upstream `807c5c41894143513100ac58fa07e72df221baa1` no longer exists, so had to update this. was this a commit from the main branch? The repo now has no commits older than 14h, so it is concerning from a security standpoint to depend on a project that "rewrites its history". I will raise an issue upstream.

pjbgf:
Removed AppArmor from `Docker.ubi`. Security model to be reviewed on follow-up PR. 
@JAORMX @jhrozek @saschagrunert PTAL

pjbgf:
e2e-fedora issue seems unrelated, retrying again

internal/pkg/manager/spod/spod_controller.go
JAORMX: Did we get to any consensus here? Is AppArmor gonna switch to a similar sidecar container model that the SELinux part uses? or will we take this as-is

pjbgf: @jhrozek  @JAORMX we spoke a bit off-line on this on last catch-up. I am correct assuming we are in consensus on the approach: merge as-is and then propose a better apporach on follow-up PR. Would you be OK with that?

pjbgf:
/override pull-security-profiles-operator-verify

overriding due to the current verify flakeness

PR #683, opened 2021-11-22
Allow fetching running spod name from within it

JAORMX:
#### What type of PR is this?

/kind cleanup

Currently, the SPOD is a singleton that denotes the DaemonSet instance
running on the nodes to do the SPO work. This is mostly just enforced by
the fact that the profilerecorder will fetch the default SPOD to check
its configuration.

This changes that by adding an environment variable on the running pod
that allows one to fetch the SPOD name from the env variable. There is
now a utility function to help with that.

The result is that we would (in theory) now be able to run multiple SPOd
instances that cover different node sets. However, this also just makes the
codebase less hardcoded and more adaptable.

None

#### Does this PR have test?

N/A.

```release-note
NONE
```

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -34,6 +34,10 @@ const (
 	// Service Account for the security-profiles-operator daemon.
 	SPOdServiceAccount = SPOdName
 
+	// SPOdNameEnvKey allows one to query the name of the SPOd instance
+	// from within the daemon.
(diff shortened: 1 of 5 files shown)


JAORMX:
Pull-request updated, HEAD is now a7560ba6142cf8bc0b4c26c97af85684a5166928

saschagrunert (approved):

PR #684, opened 2021-11-22
tests: Don't wait for all pods in all namespaces when testing against an OpenShift cluster

jhrozek:
#### What type of PR is this?
/kind cleanup

Commit cb9760d75d9d08ea624a624446d8bc38685756db made our tests much more
stable, but the part that waits for all pods in all namespaces doesn't
work well on OCP, because there's typically too many pods and the check
takes too much time.

In addition, at least at the moment, the tests against an OCP cluster
are manually executed against an already provisioned cluster.

The two other supported cluster types (vanilla and kind) have not changed
behaviour and still wait on all pods in all namespaces.

None

#### Does this PR have test?
N/A

N/A

```release-note
None
```

### DIFF
--- a/test/e2e_test.go
+++ b/test/e2e_test.go
@@ -48,8 +48,7 @@ const (
 )
 
 func (e *e2e) TestSecurityProfilesOperator() {
-	e.logf("Waiting for all pods to become ready")
-	e.waitFor("condition=ready", "pods", "--all", "--all-namespaces")
+	e.waitForReadyPods()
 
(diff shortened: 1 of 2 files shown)


jhrozek:
Ugh, I really can't program can I?
/hold

jhrozek:
OK, the last version succeeded. I was looking at test logs from the old PR (confused emoji)
/hold cancel

saschagrunert (approved):

PR #685, opened 2021-11-22
secprofnodestatus: Link the node profile status with profile using a custom label

### LINKED ISSUE #408: Limit the `spo.x-k8s.io/profile-name` label content length or limit the profile name length to make them usable for labels
Limit the profile name length to 63 characters. Other ways of solving the problem described below are acceptable, too..

#### Why is this needed:
We added the `spo.x-k8s.io/profile-name` label (`StatusToProfLabel` in the code) as a label of the node statuses to be able to list node statuses per profile. This is useful, but the profile names are not limited and at the same time k8s labels are limited to 63 characters. If someone creates a profile with a longer name, the label wouldn't be assigned.

#### User story covered
As an application SRE, Rajesh needs to be able to see the state of the installed profile(s) and verify that it has indeed been installed on the system.

jhrozek:
#### What type of PR is this?
/kind cleanup

We were using the profile name as the value of the spo.x-k8s.io/profile-name
label to be able to list node profile statuses for a profile. This is
user-friendly but would fail in case the profile name was longer than 64
characters as that's the label size limit.

Instead, let's rename the label to spo.x-k8s.io/profile-id. In case the
profile name is shorter than the limit, let's still use the name because
it's user-friendly, but in case the name is longer, let's instead use a
sha1 hash of the name with the profile Kind as a prefix.

Using sha1 is OK in this case, because we're not using it for any crypto
purposes, just to shorten a name.

Fixes #408

#### Does this PR have test?
Yes

N/A

```release-note
The securityprofilenodestatus CR now links with the security profile its status
it represents using label spo.x-k8s.io/profile-id. If the profile name is less
than 64 characters long, then the label value is the profile name, otherwise it's
kind-sha1hashofthename.

This change supports profile names whose names are over 64 characters.

```

### DIFF
--- a/internal/pkg/util/names.go
+++ b/internal/pkg/util/names.go
@@ -16,7 +16,15 @@ limitations under the License.
 
 package util
 
-import "k8s.io/apimachinery/pkg/types"
+import (
+	"crypto/sha256"
+	"fmt"
+	"io"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
 
 func NamespacedName(name, namespace string) types.NamespacedName {
 	return types.NamespacedName{
@@ -33,4 +41,45 @@ func Contains(a []string, b string) bool {
 		}
 	}
 	return false
+}
+
+// LengthName creates a string of maximum defined length.
+func lengthName(maxLen int, hashPrefix, format string, a ...interface{}) (string, error) {
+	friendlyName := fmt.Sprintf(format, a...)
+	if len(friendlyName) < maxLen {
+		return friendlyName, nil
+	}
+
+	// If that's too long, just hash the name. It's not very user friendly, but whatever
+	hasher := sha256.New()
+	if _, err := io.WriteString(hasher, friendlyName); err != nil {
+		return "", errors.Wrap(err, "writing string")
+	}
+
+	hashStr := fmt.Sprintf("%x", hasher.Sum(nil))
+	hashUseLen := maxLen - len(hashPrefix) - 1 // -1 for the dash separator
+	if hashUseLen < maxLen {
+		hashStr = hashStr[:hashUseLen]
+	}
+	hashedName := fmt.Sprintf("%s-%s", hashPrefix, hashStr)
+
+	if len(hashedName) > maxLen {
+		return "", errors.New("shortening string")
+	}
+	return hashedName, nil
+}
+
+func dnsLengthName(hashPrefix, format string, a ...interface{}) string {
+	const maxDNSLen = 63
+
+	// (jhrozek): I think it makes sense to make the utility function return
+	// error, but here I think it's OK to just ignore
+	// nolint:errcheck
+	name, _ := lengthName(maxDNSLen, hashPrefix, format, a...)
+	return name
+}
+
+func KindBasedDNSLengthName(obj client.Object) string {
+	kind := obj.GetObjectKind().GroupVersionKind().Kind
+	return dnsLengthName(kind, "%s-%s", kind, obj.GetName())
 }
--- a/api/secprofnodestatus/v1alpha1/secprofnodestatus_types.go
+++ b/api/secprofnodestatus/v1alpha1/secprofnodestatus_types.go
@@ -41,8 +41,9 @@ const (
 
 // Common labels of the node status objects.
 const (
-	// Identifies the profile by name so that the admin can list all node statuses for a certain profile.
-	StatusToProfLabel = "spo.x-k8s.io/profile-name"
+	// StatusToProfLabel identifies the profile by name, or if the name is too long, by a hash so that
+	// the admin can list all node statuses for a certain profile.
+	StatusToProfLabel = "spo.x-k8s.io/profile-id"
 	// Identifies the node on which the profile is installed so that the admin can list profiles per node.
 	StatusToNodeLabel = "spo.x-k8s.io/node-name"
 	// Allows the admin to filter out node statuses with a certain state (e.g. show me all that failed).
--- a/internal/pkg/manager/nodestatus/nodestatus.go
+++ b/internal/pkg/manager/nodestatus/nodestatus.go
@@ -133,16 +133,6 @@ func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 		"Profile.Kind", prof.GetObjectKind().GroupVersionKind(),
 	)
 
-	// get all the other statuses
-	profName := instance.Labels[statusv1alpha1.StatusToProfLabel]
-	if profName == "" {
-		return reconcile.Result{}, errors.New("Unlabeled node status")
-	}
-
-	if prof.GetName() != profName {
-		return reconcile.Result{}, errors.New("status doesn't match owner")
-	}
-
 	// Initialize status if it hasn't happened already
 	if prof.GetStatusBase().Status == "" {
 		lprof.Info("Initializing Profile status")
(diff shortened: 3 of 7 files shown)


jhrozek:
hmm, apart from the verify failures the test also seem to be failing on vanilla kube..let me investigate..
/hold

JAORMX:
I think it could be the case that we have a SELinux and a Seccomp policy with the same name... Seems to me like that would cause issues. Why not identify profiles with a combinarion of `<kind> + <name>` instead?

jhrozek:
> I think it could be the case that we have a SELinux and a Seccomp policy with the same name... Seems to me like that would cause issues. Why not identify profiles with a combinarion of `<kind> + <name>` instead?

Yes, good idea. The labels now look like this:
```
spo.x-k8s.io/profile-id: SeccompProfile-log-enricher-trace
```
for short names and like this:
```
spo.x-k8s.io/profile-id: SeccompProfile-8b3e816a82b57243d0a7c6500d49a6424f69221a
```
for long names.

The tests should pass now, I was an idiot and thought the `nodes` array contains all nodes, while in fact it contains worker nodes only.

jhrozek:
ugh I suck. I'll fix this tomorrow.

jhrozek:
@saschagrunert @pjbgf @JAORMX I'm having trouble running the make verify target locally to get the same error or understand how to fix it.

Locally I'm running:
```
jhrozek@kyubey:~/devel/security-profiles-operator|label_len ⇒  ./build/golangci-lint run --fix
jhrozek@kyubey:~/devel/security-profiles-operator|label_len ⇒  gci -w -local internal/pkg/util/names_test.go
jhrozek@kyubey:~/devel/security-profiles-operator|label_len ⇒  git st
On branch label_len
Your branch is ahead of 'origin/master' by 2 commits.
  (use "git push" to publish your local commits)

nothing to commit, working tree clean
```
I have the same golangci version and I'm running go 1.17. How do I fix the linter issues?

jhrozek:
OK, "solved". I spun up a local toolbox container and installed the bpf libraries using the hack-XXX scripts. For some reason that did help.

jhrozek:
FWIW, I think the fedora test failure is unrelated. That test had passed in the previous iteration of this PR.

internal/pkg/util/names.go
```diff
@@ -34,3 +44,42 @@ func Contains(a []string, b string) bool {
...
+	if len(friendlyName) < maxLen {
+		return friendlyName, nil
+	}
+
+	// If that's too long, just hash the name. It's not very user friendly, but whatever
+	//
+	// We can suppress the gosec warning about sha1 here because we don't use sha1 for crypto
+	// purposes, but only as a string shortener
+	// nolint:gosec
+	hasher := sha1.New()
```
pjbgf, line 60: Completely understand the rational here. I would probably use sha256 though, given the project's focus in security we should refrain from taking dependencies on algorithmns or libraries that are known to be weak/vulnerable.

This may also be a problem for users in high compliance sectors, as this is not NIST compliant.

pjbgf (changes requested):
nice one @jhrozek, we may need a small change on the `sha1`. apart from that it looks good.

internal/pkg/util/names.go
jhrozek: Thank you for the review. I understand where you're coming from, but the problem is the 64 characters length that labels are restricted to in k8s. Consider this (it's python but I hope it's readable nonetheless):
```
In [5]: hashlib.sha1('foo'.encode()).hexdigest()
Out[5]: '0beec7b5ea3f0fdbc95d0dd47f3c5bc275da8a33'

In [6]: hashlib.sha256('foo'.encode()).hexdigest()
Out[6]: '2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae'

In [7]: len(hashlib.sha1('foo'.encode()).hexdigest())
Out[7]: 40

In [8]: len(hashlib.sha256('foo'.encode()).hexdigest())
Out[8]: 64
```
so sha256 doesn't already fit on its own:
```
kubectl label sp log-enricher-trace hash=2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae
error: invalid label value: "hash=2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae": must be no more than 63 characters
```

One thing we could do (but..it's kindof a hack..) is to trim the length a bit, e.g. use an equivalent of git's abbrev and fall back to using the full hash without prefix if there is a collision, which frankly is quite unlikely. And even if there was a collision, it wouldn't be that much of a deal as the hash is not used for any security critical functionality, just a link between two objects.

pjbgf: If you are cool with using sha256, trimming the tail end of it is a common approach which slighty increases chance of collision but that isn't a problem on that case - and less likely to occur than sha1 anyway. 

Otherwise, would you be opposed to simply restricting the size of profile name and using that instead? 

In both the approaches above, any concerns around name collision through profiles with the same name deployed in different namespaces?

Now if this can be non-deterministic, we could also just append the hex of some random values at the tail end using `rand.Read(buf)`.

jhrozek: > If you are cool with using sha256, trimming the tail end of it is a common approach which slighty increases chance of collision but that isn't a problem on that case - and less likely to occur than sha1 anyway.

How much is reasonable to trim? As little as possible to fill the 63 characters?

> Otherwise, would you be opposed to simply restricting the size of profile name and using that instead?

I would not be opposed to that. The only catch is that we use sometimes derive security profile resources from other resources (e.g. we derive from recording name), so we need to be careful and restrict them all or we might be unable to generate the profiles later.

> In both the approaches above, any concerns around name collision through profiles with the same name deployed in different namespaces?

Both objects are namespaced and live in the same namespace, so I think using the namehash is no different than using the plain name. Or did you have some other issue in mind?

> Now if this can be non-deterministic, we could also just append the hex of some random values at the tail end using rand.Read(buf).

I honestly prefer deterministic at least to make testing easier.

jhrozek: ok, let me start by trimming the hash and in parallel look at limiting the name. We can compare the two approaches then (and see test results!)

pjbgf: > How much is reasonable to trim? As little as possible to fill the 63 characters?

Yep, as little as possible would be ideal.

> I would not be opposed to that. The only catch is that we use sometimes derive security profile resources from other resources (e.g. we derive from recording name), so we need to be careful and restrict them all or we might be unable to generate the profiles later.

Very good point. That's definitely something we want to get right and aligned sooner rather than later.

> Both objects are namespaced and live in the same namespace, so I think using the namehash is no different than using the plain name. Or did you have some other issue in mind?

That's right - same issue in both cases. I was thinking this could be a problem when we deploy SPO to watch multiple namespaces for Security Profiles, which I think we still support - but not completely sure.

jhrozek:
OK, for now I've changed the hash to use sha256. Still todo is to look at name length validation and check the multi-namespace labels.

jhrozek:
OK, for limiting the name we have basically two options:
 - let the object be created with a long name and either mark it with an error and refuse to process it further or just issue a warning (because the name is only used in labels for the  `secprofnodestatus` object which is not super critical)
 - add a validating webhook. This is probably the more correct way, but honestly I dislike adding another webhook just for this case..

pjbgf:
fedora-e2e failed after running for over 50 minutes - retriggering it now.

pjbgf:
> let the object be created with a long name and either mark it with an error and refuse to process it further or just issue a warning (because the name is only used in labels for the secprofnodestatus object which is not super critical)

happy to take this approach for now.

pjbgf (approved):
/lgtm

the validation can be done on a follow-up PR.

PR #695, opened 2021-11-25
Add OpenShift deployment manifests

### LINKED ISSUE #689: Add OpenShift deployment manifest

We should have a `deploy/openshift.yaml` deployment manifest that easily allows folks to deploy this on OpenShift.

#### Why is this needed:

We currently have two deployment manifests:

* `deploy/operator.yaml`
* `deploy/namespace-operator.yaml`

These allow folks to deploy the operator on vanilla kubernetes with a couple of different configurations (cluster-scoped operator and namespace scoped operator).

We also have the means of deploying the operator and testing it in OpenShift, so this should be reflected in the manifests.

We have some Makefile targets that allow one to deploy testing version of the operator to OpenShift, but this is not ideal and is not very friendly for newcomers.

#### User story covered

As an Infrastructure SRE, Kirsti wants to deploy the SPO in OpenShift.

JAORMX:
#### What type of PR is this?

/kind cleanup

This adds explicit manifests to deploy the operator in OpenShift.

This was done using kustomize overlays, and two overlays were introduced:

* `openshift`: Deploy the SPO in OpenShift with SELinux enabled by default.
* `openshift-dev`: Deploy the SPO in OpenShift assuming the images are deployed
                   in the cluster's internal registry (useful for testing).

The `openshift-dev` is used explicitly by the `deploy-openshift-dev` Makefile
target which takes care of uploading the images to the registry.

this also adds an environment variable to the manager to enable SELinux by
default in the SPOD we deploy by on creation.

Fixes #689

#### Does this PR have test?

No

```release-note
An OpenShift deployment manifest was included in deploy/openshift.yaml
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,6 +131,8 @@ $(BUILD_DIR)/kustomize: $(BUILD_DIR)
 deployments: $(BUILD_DIR)/kustomize manifests generate ## Generate the deployment files with kustomize
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/cluster -o deploy/operator.yaml
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/namespaced -o deploy/namespace-operator.yaml
+	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/openshift -o deploy/openshift.yaml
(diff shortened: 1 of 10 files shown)


jhrozek (approved):
I tested the openshift makefile target and SPO was deployed fine with selinuxd running. The only difference between the -dev and non-dev manifest is the image references, so the deployment yaml should work fine when we release new images as well.

PR #696, opened 2021-11-25
hack: Run deployments verify multiple times to ignore ordering issues

JAORMX:
#### What type of PR is this?

/kind flake

When there's more than one CRD coming from a package, controller-gen
might generate them in a random order. This doesn't play well with our
verify target. To address this, we run it multiple times to ignore such
issues.

None

#### Does this PR have test?

N/A.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -297,7 +297,7 @@ verify-go-mod: update-go-mod ## Verify the go modules
 
 .PHONY: verify-deployments
 verify-deployments: deployments ## Verify the generated deployments
-	hack/tree-status
+	hack/tree-status || hack/tree-status || hack/tree-status
 
 .PHONY: verify-go-lint
 verify-go-lint: $(BUILD_DIR)/golangci-lint ## Verify the golang code by linting


JAORMX:
/retest

e2e is flaky

Makefile
```diff
@@ -297,7 +297,7 @@ verify-go-mod: update-go-mod ## Verify the go modules
 
 .PHONY: verify-deployments
 verify-deployments: deployments ## Verify the generated deployments
-	hack/tree-status
+	hack/tree-status || hack/tree-status || hack/tree-status
```
saschagrunert, line 300: I'd rather remove the `verify-deployments` from the verify target for now.

JAORMX: There is still some value in running this as it indicates us if a contributor has run `make deployments`

saschagrunert: But then we have to run `make deployments` as well, not only `hack/tree-status`, right?

JAORMX: duh... you're right. I got confused

PR #698, opened 2021-11-25
spod: Make /etc/selinux.d an emptyDir instead of hostPath

jhrozek:
#### What type of PR is this?
/kind cleanup

Since the directory only contains policy definitions, not the actual
policy result it is not needed to be a hostPath. The actual policy is
persisted in /etc/selinux (without the .d).

The original idea behind having a hostPath was to run selinuxd on the
nodes directly to lock down even e.g. the kubelet. But that will not
happen, at least not in the short term, so let's reduce the number of
our hostPath mounts a little.

None

#### Does this PR have test?
N/A tested by existing tests

N/A

```release-note
The directory /etc/selinux.d used to be mounted on the hosts in previous SPO versions.
This is no longer the case, the directory was converted to an emptyDir instead,
reducing the number of required host mounts.

```

### DIFF
--- a/internal/pkg/manager/spod/bindata/spod.go
+++ b/internal/pkg/manager/spod/bindata/spod.go
@@ -582,10 +582,7 @@ semodule -i /opt/spo-profiles/selinuxrecording.cil
 					{
 						Name: "selinux-drop-dir",
 						VolumeSource: corev1.VolumeSource{
-							HostPath: &corev1.HostPathVolumeSource{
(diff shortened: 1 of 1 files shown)


JAORMX (commented):
lgtm

I'll merge it once it passes CI

saschagrunert (approved):

JAORMX:
/hold

debugging this a little before we merge

PR #702, opened 2021-11-25
Make short running containers work with bpf recorder

saschagrunert:
#### What type of PR is this?

/kind feature

A couple of enhancements have been done to make the bpf recorder work
containers having a short lifetime:

- Add a PID lock, which makes callers of `SyscallsForProfile` wait for
  events to be processed.
- Do not process events on the startup module load test.
- Reduce the amount of time for retrying the container ID retrieval.
- Increase logging verbosity to be able to trace the retries, especially
  if the container ID is not yet set.
- Added a `kubectl run` e2e test to proof the feature

None
#### Does this PR have test?
Yes

None

```release-note
None
```


saschagrunert:
Note: the mutex lock does not solve the issue when running multiple short running recordings in parallel. I have an idea how to solve this and follow  up on the topic in another PR.

PR #707, opened 2021-11-28
selinux: Fix selinuxd policy for RHEL-8/CentoOS-8/RHCOS

### LINKED ISSUE #699: selinuxd SELinux policy is broken for CentOS-based selinuxd deployments
#### What happened:
We test selinuxd with SELinux enabled and its own label in our fedora e2e tests, but then the openshift deployment uses CentOS based images for selinuxd. Because selinuxd calls out to libsemanage and is not totally statically linked, its behaviour is dependent on what platform it is running on.
And it seems that the policy works fine for Fedora-based images (as seen in our fedora e2e tests) but not the CentOS based images that are used when deploying on OCP.

We then see AVC denials like this:
```
type=AVC msg=audit(1637847205.755:121): avc:  denied  { rename } for  pid=241207 comm="selinuxdctl" name="tmp" dev="nvme0n1p4" ino=73403084 scontext=system_u:system_r:selinuxd.process:s0:c586,c793 tcontext=system_u:object_r:selinux_config_t:s0 tclass=dir permissive=0
```

selinuxd works

install an SELinux policy

this is only an issue on CentOS-8/RHEL-8/RHCOS based systems

#### Environment:

- Cloud provider or hardware configuration: AWS
- OS (e.g: `cat /etc/os-release`):RHCOS 4.9
- Kernel (e.g. `uname -a`): Not important
- Others: Not important

jhrozek:
#### What type of PR is this?
/kind regression

While our update of selinuxd's policy worked fine (see CI), it broke
selinuxd's usage on OCP.

This patch brings back the needed changes for selinuxd to work fine
again on RHEL-8 based systems.

Fixes #699

#### Does this PR have test?
No, but we should have CentOS-8 based tests in CI I guess.

@JAORMX what do you think about just updating the policy quickly for now and then having a separate Fedora-latest and Centos-8? For OCP4 we could select the Centos-8 based one as well as default to it, whoever is crazy/brave enough to deploy Fedora in their cluster would have to kustomize their deployment themselves.

```release-note
NONE
```


JAORMX:
Seems to me like these changes will break our fedora tests

JAORMX:
huh, I thought this would have broken on Fedora. I guess the older policy is still applicable?

JAORMX:
@jhrozek ah, seems you need to rebase

jhrozek:
> huh, I thought this would have broken on Fedora. I guess the older policy is still applicable?

This PR only adds to the Fedora policy, so far they are not mutually exclusive.

JAORMX:
@jhrozek got it; that makes sense. Still gotta rebase though, your PR is missing the AppArmor bits of the rendered manifests.

jhrozek:
On Mon, Nov 29, 2021 at 04:35:32AM -0800, Juan Osorio Robles wrote:
> @jhrozek got it; that makes sense. Still gotta rebase though, your PR is missing the AppArmor bits of the rendered manifests.

Oops, done.

saschagrunert (approved):

PR #719, opened 2021-11-30
Finish profilerecorder unit tests

saschagrunert:
#### What type of PR is this?

/kind cleanup

This adds all other required unit test cases and bump the coverage around 98%.

None
#### Does this PR have test?
Yes

None

```release-note
None
```

### DIFF
--- a/internal/pkg/daemon/profilerecorder/impl.go
+++ b/internal/pkg/daemon/profilerecorder/impl.go
@@ -18,6 +18,7 @@ package profilerecorder
 
 import (
 	"context"
+	"io/ioutil"
 
 	"github.com/containers/common/pkg/seccomp"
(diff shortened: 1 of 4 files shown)


JAORMX:
@saschagrunert I'm very much into this extended test coverage!

This is more of a nit and I'll merge this anyway... but could we cover the profilerecorder unit tests with different `t.Run` calls so the documentation would now be a test run name as opposed to a comment? This would make it easier to debug if we would break a test in the future.

JAORMX (approved):
/lgtm

saschagrunert:
> This is more of a nit and I'll merge this anyway... but could we cover the profilerecorder unit tests with different `t.Run` calls so the documentation would now be a test run name as opposed to a comment? This would make it easier to debug if we would break a test in the future.

Oh yeah, I can follow-up on that :+1:

PR #732, opened 2021-12-01
Sort auto-generated CRDs

pjbgf:
#### What type of PR is this?

/kind bug
/kind failing-test

The YAML CRD autogeneration logic is non-deterministic, making files containing multiple CRDs to break
the make verify CI tests. This PR sorts them to ensure they are always in the same order.

This also updates gcb-docker-gcloud's version - another reason for `make verify` failures.

None

#### Does this PR have test?

N/A

```release-note
NONE
```


JAORMX:
/lgtm

woo-hoo! thanks!!

Issue #653, opened 2021-11-09
Improve visibility of SPO's security posture


pjbgf:
#### What would you like to be added:

This is a grouping issue to track the different actions around improving and documenting SPO's security posture.

- [ ] Review gaps from the [OpenSSF scorecard](https://github.com/ossf/scorecard)
  - [x] Branch-Protection #660 
  - [x] CII-Best-Practices 
  - [ ] Fuzzing
  - [x] Pinned-Dependencies
  - [x] Signed-Releases
  - [x] Token-Permissions #737 
- [x] Document SPO security model (https://github.com/kubernetes-sigs/security-profiles-operator/pull/693)
  - [x] Container-Host Isolation
  - [x] Control Plane RBAC

#### Why is this needed:

Improve the shared understanding of SPO's current security position, document strengths and shortcomings allowing users to assess its security model.

#### User story covered

WIP - Pending PR to add new persona/story.

pjbgf:
For visibility, here's the current OpenSSF scorecard:
![image](https://user-images.githubusercontent.com/5452977/144217858-151d641a-d7f4-4350-8769-16459b1ee0d4.png)

pjbgf:
The `CII-Best-Practices` has been completed today and the project has given the `passing status` as it can be seen [here](https://bestpractices.coreinfrastructure.org/en/projects/5368).

The only caveat here is around fuzzing of memory-unsafe languages. Currently the only scope for this is the eBPF implementation. Given its experimental nature, I tagged it as `N/A`. Once this feature stabilises and is shipped to users, this will need to be reviewed. (FYI @saschagrunert)

pjbgf:
Note to self: stop marking PR as `partially fixes`.

pjbgf:
Although the upstream card changed its format, we can still see the impact of the recent PRs.
Before we were failing in 4 out of 8 categories, now we are scoring 8.6/10. :tada: 

![image](https://user-images.githubusercontent.com/5452977/146226637-b55045bf-2cd1-4047-be78-56b8782c522e.png)

pjbgf:
Updated scorecard:
![image](https://user-images.githubusercontent.com/5452977/204080754-bff8c8e0-8e46-4891-a17e-df8c5877fca0.png)

PR #737, opened 2021-12-01
Remove unecessary token permissions

pjbgf:
#### What type of PR is this?

/kind cleanup

Apply the least-privilege principle to the github token used within Github Actions.
This is a best practice as recommended by the [ossf](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions).

Part of #653

#### Does this PR have test?

N/A

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,19 @@ on:
   pull_request:
 env:
   GO_VERSION: '1.17'
+permissions:
+  actions: none
+  checks: none
+  contents: none
+  deployments: none
+  id-token: none
+  issues: none
+  discussions: none
+  packages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
 jobs:
   macos:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -6,6 +6,19 @@ on:
...
+  contents: none
+  deployments: none
+  id-token: none
+  issues: none
+  discussions: none
+  packages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
```
pjbgf, line 21: The documentation seems to imply that using `permissions:` without specifying each one of the options have the same effect as setting `none` to all of them.

I ended up choosing the verbose approach as it made explicit the permission model _intented_ without requiring awareness of this detail by whoever stumbles into this file.

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#permissions

saschagrunert (approved):
Great!

PR #740, opened 2021-12-01
openshift: Deploy webhook using internal CA instead of cert-manager

jhrozek:
#### What type of PR is this?
/kind cleanup

OpenShift does not include cert-manager by default and moreover includes
its own way of generating keys and certificates for services:

- https://docs.openshift.com/container-platform/4.7/nodes/pods/nodes-pods-secrets.html#nodes-pods-secrets-certificates-creating_nodes-pods-secrets
-  https://docs.openshift.com/container-platform/4.7/security/certificate_types_descriptions/service-ca-certificates.html

and providing the CA bundles for webhook configuration:
  
- https://docs.openshift.com/container-platform/4.7/security/certificates/service-serving-certificate.html#add-service-certificate-mutating-webhook_service-serving-certificate

While we could have add a dependency on the OpenShift Operator level, it is not needed and we can instead just use the built-in CA.

None

#### Does this PR have test?
Not specifically, but we can run existing tests on OCP

Marking this as WIP because I haven't tested myself past deployment, but I'm interested in early feedback.

```release-note
When deploying on OpenShift, cert-manager is no longer required.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,8 @@ $(BUILD_DIR)/kubernetes-split-yaml: $(BUILD_DIR)
 deployments: $(BUILD_DIR)/kustomize $(BUILD_DIR)/kubernetes-split-yaml manifests generate ## Generate the deployment files with kustomize
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/cluster -o deploy/operator.yaml
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/namespaced -o deploy/namespace-operator.yaml
+	cp -f deploy/base/certificate.yaml deploy/overlays/openshift/no_certificate.yaml
+	sed -i '/^---$$/a $$patch: delete' deploy/overlays/openshift/no_certificate.yaml
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/openshift -o deploy/openshift.yaml
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/openshift-dev -o deploy/openshift-dev.yaml
 
@@ -410,9 +412,6 @@ push-openshift-dev: set-openshift-image-params openshift-user image
 
 .PHONY: do-deploy-openshift-dev
 do-deploy-openshift-dev: $(BUILD_DIR)/kustomize
-	@echo "Deploying cert-manager"
-	oc apply -f https://github.com/jetstack/cert-manager/releases/download/v1.6.1/cert-manager.yaml
-	oc wait --for=condition=Ready pod -lapp.kubernetes.io/instance=cert-manager -ncert-manager
 	@echo "Deploying"
 	oc apply -f deploy/openshift-dev.yaml
 	@echo "Setting triggers to track image"
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -36,8 +36,6 @@ dependencies:
       match: jetstack/cert-manager
     - path: installation-usage.md
       match: jetstack/cert-manager
-    - path: Makefile
-      match: jetstack/cert-manager
 
   - name: kind
     version: 0.11.1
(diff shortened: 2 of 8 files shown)


jhrozek:
self-nack, the non-dev openshift deployment is missing some annotations:
```
diff deploy/openshift-dev.yaml deploy/openshift.yaml
1705c1705
<         image: image-registry.openshift-image-registry.svc:5000/openshift/security-profiles-operator:latest
---
>         image: gcr.io/k8s-staging-sp-operator/security-profiles-operator:latest
1759c1759
<         image: image-registry.openshift-image-registry.svc:5000/openshift/security-profiles-operator:latest
---
>         image: gcr.io/k8s-staging-sp-operator/security-profiles-operator:latest
1867,1868d1866
<   annotations:
<     service.beta.openshift.io/serving-cert-secret-name: webhook-server-cert
1884,1885d1881
<   annotations:
<     service.beta.openshift.io/serving-cert-secret-name: metrics-server-cert
```

Makefile
```diff
@@ -154,6 +154,8 @@ $(BUILD_DIR)/kubernetes-split-yaml: $(BUILD_DIR)
 deployments: $(BUILD_DIR)/kustomize $(BUILD_DIR)/kubernetes-split-yaml manifests generate ## Generate the deployment files with kustomize
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/cluster -o deploy/operator.yaml
 	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/namespaced -o deploy/namespace-operator.yaml
+	cp -f deploy/base/certificate.yaml deploy/overlays/openshift/no_certificate.yaml
```
jhrozek, line 157: @JAORMX this feels like a bit of a hack, but I wanted to make sure we include nothing from the certificate.yaml file. Just pointing this out myself for review..

JAORMX: I'm not sure I get the reasoning for this. the overlay should have everything to build the `openshift.yaml` and `openshift-dev.yaml` files.

JAORMX (changes requested):

jhrozek:
On Thu, Dec 02, 2021 at 07:23:58AM -0800, Juan Osorio Robles wrote:
> @JAORMX requested changes on this pull request.
> 
> 
> 
> > @@ -154,6 +154,8 @@ $(BUILD_DIR)/kubernetes-split-yaml: $(BUILD_DIR)
>  deployments: $(BUILD_DIR)/kustomize $(BUILD_DIR)/kubernetes-split-yaml manifests generate ## Generate the deployment files with kustomize
>  	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/cluster -o deploy/operator.yaml
>  	$(BUILD_DIR)/kustomize build --reorder=none deploy/overlays/namespaced -o deploy/namespace-operator.yaml
> +	cp -f deploy/base/certificate.yaml deploy/overlays/openshift/no_certificate.yaml
> 
> I'm not sure I get the reasoning for this. the overlay should have everything to build the `openshift.yaml` and `openshift-dev.yaml` files. 

Right, so the `deploy/base/certificate.yaml` file contains manifests to
create an Issuer and two Certificate, all of the three manifests use
kinds from the cert-manager.io/v1 API group. This API group will not be
available if we don't install cert-manager on OCP, so what we need is to
skip the creation of these objects altogether.

I didn't find a better way with kustomize to skip a manifest as a whole
other than mark each object to delete with `$patch: delete`. And because
I want to be sure we won't miss any newly added object from
`deploy/base/certificate.yaml` I copy the file and then sed the `---`
YAML separators to be instead `---\n$patch:delete`.

It's still hacky, of course, but I didn't find a better way.

jhrozek:
heh, previously e2e-ubuntu was passing and e2e-fedora was failing and now it's the other way around.
Either way, this PR only touched the openshift overlays which are not tested anyway..
/retest

jhrozek:
@JAORMX PTAL

saschagrunert (approved):

PR #741, opened 2021-12-01
Pin container base images

pjbgf:
#### What type of PR is this?

/kind cleanup

#### What this PR does

Pins most docker base images.

#### Why we need it:

Pinned dependencies reduce several security risks:

- They ensure that checking and deployment are all done with the same software, reducing deployment risks, simplifying debugging, and enabling reproducibility.
- They can help mitigate compromised dependencies from undermining the security of the project (in the case where you've evaluated the pinned dependency, you are confident it's not compromised, and a later version is released that is compromised).
- They are one way to counter dependency confusion (aka substitution) attacks, in which an application uses multiple feeds to acquire software packages (a "hybrid configuration"), and attackers fool the user into using a malicious package via a feed that was not expected for that package.

More information refer to [ossf](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies)

Partially fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/653

#### Does this PR have test?

N/A

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# to pin base image below may require further changes on build/release processes
 FROM quay.io/security-profiles-operator/build:latest AS build
 
 COPY . /work
--- a/Dockerfile.build-image
+++ b/Dockerfile.build-image
@@ -12,7 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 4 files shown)


pjbgf:
`fedora-e2e` flakeness:
`--- FAIL: TestSuite/TestSecurityProfilesOperator/cluster-wide:_Seccomp:_Verify_profile_recording_bpf (233.12s)`

saschagrunert (approved):

PR #743, opened 2021-12-02
WIP: Split fedora-e2e flake tests

pjbgf:
#### What type of PR is this?

/kind cleanup
/kind flake

Some tests are intermittendly failing (a.k.a. flaking). They are most often happening as part of the `fedora-e2e` test suite which  currently takes around 40 minutes to run.

By splitting them from the more reliable tests allows us to better decide how to approach the problem - (e.g. auto-retry). 

None

#### Does this PR have test?

N/A

```release-note
NONE
```

### DIFF
--- a/test/e2e_flake_test.go
+++ b/test/e2e_flake_test.go
@@ -1,0 +1,75 @@
+//go:build flake
+// +build flake
+
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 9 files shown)


test/e2e_flake_test.go
```diff
@@ -0,0 +1,59 @@
...
+	})
+
+	e.Run("cluster-wide: Seccomp: Verify profile recording bpf", func() {
+		e.enableBpfRecorderInSpod()
+		e.testCaseBpfRecorderKubectlRun()
+		e.testCaseBpfRecorderStaticPod()
+		e.testCaseBpfRecorderMultiContainer()
+		e.testCaseBpfRecorderDeployment()
+		e.testCaseBpfRecorderParallel()
+	})
```
saschagrunert, line 55: Are those tests really flaking?

pjbgf: I don't think they are all flaking individually, but as a group yes. So thought it would make sense to keep them in their current grouping - happy to change if you think that would be better.

I started this PR segregating the BPF E2E tests, but ended up making the split by flaky tests as there were others that could benefit by an auto retry. 

@saschagrunert do you think is OK to keep these grouped until they are consistently reliable?

saschagrunert: Yes, sounds good to me. :)

pjbgf:
Moving back to WIP, will review in the coming days.

Issue #736, opened 2021-12-01
Do memory and CPU benchmark


saschagrunert:
Benchmarking the SPOD containers with a manual test and adjust the resources accordingly before doing the release.

saschagrunert:
I’ll do the BPF recorder first tomorrow.

saschagrunert:
bpf recorder results while doing a test with 10 and later with 50 pod recordings in parallel on 3 worker nodes:
![screenshot](https://user-images.githubusercontent.com/695473/144398525-da5c73d7-957b-4d9e-9680-c05e11fee144.png)

Looks like our current memory restrictions are good enough:
https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/internal/pkg/manager/spod/bindata/spod.go#L501-L506

saschagrunert:
Did a 10 minute test with 50 pods recordings via the log enricher, the new memory values seems to pay out:
![screenshot](https://user-images.githubusercontent.com/695473/145055181-bec0a1dc-63be-4018-a706-793e0dcd45fa.png)

saschagrunert:
@jhrozek I'm done with testing, feel free to reopen if you see anything left open.

jhrozek:
On Tue, Dec 07, 2021 at 07:13:54AM -0800, Sascha Grunert wrote:
> @jhrozek I'm done with testing, feel free to reopen if you see anything left open.

The only thing I found on my side is the webhook race which is already
filed an a selinuxd memory leak for which I still
don't have a patch. Looks like most of the issues were triggered by
seccomp due to the bigger size of the logs..

PR #748, opened 2021-12-03
Increase ephemeral storage for log-enricher

saschagrunert:
#### What type of PR is this?

/kind bug

The log-enricher gets killed by the kubelet when running intensive (50 pods) seccomp recording tests:
```
Container log-enricher exceeded its local ephemeral storage limit \"20Mi\".
```
We now increase the memory to have more room left.

Refers to #736 
None

```release-note
None
```


jhrozek (approved):
/lgtm

pjbgf (approved):

PR #763, opened 2021-12-14
Switch back to dev VERSION

saschagrunert:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:
Switching back to dev after the release.
#### Which issue(s) this PR fixes:

Refers to #613 
#### Does this PR have test?
None

#### Special notes for your reviewer:
None
#### Does this PR introduce a user-facing change?

```release-note
None
```

PR #764, opened 2021-12-14
Increase manager memory constraints

saschagrunert:
#### What type of PR is this?

/kind bug

While doing some long term testing of the operator I saw that the
manager is near its memory constraints, which causes it to be killed
from time to time. To avoid that we now bump the memory to a more stable
level.

/hold
should merge after https://github.com/kubernetes-sigs/security-profiles-operator/pull/763
#### Does this PR have test?

None
screenshot

```release-note
Increase manager memory limit to 128MiB
```

### DIFF
--- a/deploy/base/manager_deployment.yaml
+++ b/deploy/base/manager_deployment.yaml
@@ -28,10 +28,10 @@ spec:
             allowPrivilegeEscalation: false
           resources:
             requests:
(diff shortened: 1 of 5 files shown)


jhrozek (approved):
/lgtm
I think these limits are still quite sane

pjbgf (approved):

PR #809, opened 2022-02-11
Automatically determine cert-manager requirement

saschagrunert:
#### What type of PR is this?

/kind feature

We now deploy the cert manager related resources by using the internal
manager. This way we can pro grammatically decide if cert-manager is
required or not and adjust the deployment accordingly.

None
#### Does this PR have test?
None

None

```release-note
Automatically determine if cert-manager is required or not, for example in OpenShift deployments.
```


deploy/overlays/openshift/kustomization.yaml
```diff
@@ -8,11 +8,3 @@ bases:
 
 patchesStrategicMerge:
   - deployment.yaml
```
saschagrunert, line 10: The only difference between the OpenShift deployment and the standard one is that we enable selinux. @jhrozek do we wanna push this logic into the manager as well?

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -107,6 +110,7 @@ func (r *ReconcileSPOd) Healthz(*http.Request) error {
 //
 // OpenShift (This is ignored in other distros):
 // +kubebuilder:rbac:groups=security.openshift.io,namespace="security-profiles-operator",resources=securitycontextconstraints,verbs=use
+// +kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=get;list;watch
```
saschagrunert, line 113: I dislike that. Do we have a better way to find out that we're running in OpenShift or not?

jhrozek: In the ComplianceAsCode project, we detect OCP by looking for an instance of `apiVersion` `config.openshift.io/v1`, `Kind: ClusterOperator`, `name: openshift-apiserver`.  Looking for the instance might be an overkill, we do it to make sure that the version matches `4.y.z`, but I guess looking for `ClusterOperator` might be a bit safer than looking for `SCCs`.

deploy/overlays/openshift/kustomization.yaml
jhrozek: I think it would be nice, to enable SELinux by default through the manager as well.

saschagrunert: So maybe if it’s not set explicitly in the config (not sure how to determine that), then we enable it automatically on OpenShift.

saschagrunert:
There seems to be an issue with the operator removal test.

internal/pkg/manager/spod/spod_controller.go
saschagrunert: Fixed :+1:

saschagrunert:
@jhrozek PTAL

internal/pkg/manager/spod/bindata/webhook.go
```diff
@@ -0,0 +1,319 @@
...
+
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
+)
+
+var (
+	replicas                int32 = 3
+	defaultMode             int32 = 420
+	failurePolicy                 = admissionregv1.Fail
+	caBundle                      = []byte("Cg==")
+	bindindPath                   = "/mutate-v1-pod-binding"
```
jhrozek, line 41: nit: s/bindind/binding/
OK to fix in another patch.

jhrozek (approved):
/lgtm
Looks good!
The only thing we need to fix as a follow-up is that the e2e tests still deploy cert-manager unconditionally, I'd like to make sure that the tests on openshift all pass even w/o it. But that's follow up work.

PR #826, opened 2022-03-09
feat: make configurable the security context of each container part of SPOD in the SecurityProfilesOperatorDaemon cofiguration

ccojocar:
#### What type of PR is this?

feature

This PR extends the SecurityProfilesOperatorDaemon configuration to allow the security context to be specified for each container part of SPOD pod. This is required for instance to configure the SELinux options for nodes which doesn't enforce SELinux by default and do not provide the spc_t type. Such an example is the Flatcar Linux distribution.

#### Does this PR have test?

Yes 

```release-note

```


api/spod/v1alpha1/spod_types.go
```diff
@@ -33,6 +33,14 @@ type SelinuxOptions struct {
 	AllowedSystemProfiles []string `json:"allowedSystemProfiles,omitempty"`
 }
 
+// SecurityContext define the security context of a container part of SPOD daemon pod
+type SecurityContext struct {
```
jhrozek, line 37: Is it a good idea to name the struct the same as the one coming from `corev1` ?

ccojocar:
> So one question that I wonder after reading the patch is you think this fine-grained configuration has uses beyond setting the SELinux context. If it's about setting the selinux context only, wouldn't it be easier to just auto-detect if the nodes support selinux (or have that configurable) rather than expose all the options?

Configuring SELinux is one issue I stumbled across, and I wasn't able to get the daemon running without changing the hard-coded value. I see that it can be useful to modify or to overwrite other security parameters such as the seccomp profile.

api/spod/v1alpha1/spod_types.go
ccojocar: I am open for a suggestion if you might think that this can create some confusion.

jhrozek: What about making it either prefixed with `SPO` or such or naming it after the fact that it pairs the container and the contex like `SpoContSecCtx`.

saschagrunert:
> I can just detect if the selinux is enabled and add the type if is in enforcing mode.

I would prefer that route :+1:

ccojocar:
closing this in favour of #828

PR #828, opened 2022-03-10
fix: disable selinux from SPOD security context when selinux is not in enforcing mode

ccojocar:

This PR disables the SELinux form security context of SPOD deployment when SELinux is not running in the enforcing mode. This is required in order to have a successful deployment on nodes which do not enforce SELinux and do not provide the spc_t type label.  Such an example is the Flatcar Linux distribution.

This check is not optimal because it is done on the manager and not on each node where the daemonset is supposed to be deployed. That will most likely require to deploy a daemonset in advance of the deploy just to perform the selinux check on each node to disable it form the deployment when is not supported. Such a solution seems to me a bit too cumbersome. Do you have any better suggestion? 

```release-note
Disable SELinux option from SPOD security context when SELinux is not in enforcing mode
```

### DIFF
--- a/internal/pkg/manager/spod/setup.go
+++ b/internal/pkg/manager/spod/setup.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/crossplane/crossplane-runtime/pkg/resource"
+	selinux "github.com/opencontainers/selinux/go-selinux"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -123,6 +124,16 @@ func getEffectiveSPOd(dt *daemonTunables) *appsv1.DaemonSet {
 
 	sepolImage := &refSPOd.Spec.Template.Spec.InitContainers[1]
 	sepolImage.Image = dt.selinuxdImage // selinuxd ships the policies as well
+
+	// Disable SELinux configuration when SELinux is not in enforcing mode
+	if selinux.EnforceMode() != selinux.Enforcing {
+		for i := range refSPOd.Spec.Template.Spec.InitContainers {
+			refSPOd.Spec.Template.Spec.InitContainers[i].SecurityContext.SELinuxOptions = nil
+		}
+		for i := range refSPOd.Spec.Template.Spec.Containers {
+			refSPOd.Spec.Template.Spec.Containers[i].SecurityContext.SELinuxOptions = nil
+		}
+	}
 	return refSPOd
 }
 
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/jetstack/cert-manager v1.7.1
 	github.com/maxbrunsfeld/counterfeiter/v6 v6.5.0
 	github.com/nxadm/tail v1.4.8
+	github.com/opencontainers/selinux v1.10.0
 	github.com/openshift/api v0.0.0-20220209124712-b632c5fc10c0
 	github.com/pjbgf/go-apparmor v0.0.7
 	github.com/pkg/errors v0.9.1
--- a/vendor/github.com/opencontainers/selinux/LICENSE
+++ b/vendor/github.com/opencontainers/selinux/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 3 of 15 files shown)


saschagrunert (commented):
I think we can add a release note for this :) 

Otherwise
/lgtm

jhrozek:
lgtm also. I honestly don't see why did the fedora e2e test fail, all invidivual tests show either as pass or skip..

ccojocar:
> I think we can add a release note for this :)

@saschagrunert How do I add a release note? Thanks

saschagrunert:
> > I think we can add a release note for this :)
> 
> @saschagrunert How do I add a release note? Thanks

By adding it to the ```release-note block in the PR

ccojocar:
Thanks @saschagrunert, I added a brief release note.

JAORMX:
@ccojocar Flatcar is working on upgrading to a newer SELinux ref policy which includes `spc_t`. Currently though, containers there rely on the virt modules for labeling. So, while `spc_t` might not be there. e.g. in the Flatcar policy that's being used, `container_t` does not exist, and instead there's `svirt_lxc_net_t`. There might be an alternative label that represents `spc_t`, I'm looking at options. However, instead of turning SELinux options off; why not disable SELinux support in the SPO instead? We do have a configuration flag for this, and thus it wouldn't deploy the SELinux bits.

internal/pkg/manager/spod/setup.go
```diff
@@ -123,6 +124,16 @@ func getEffectiveSPOd(dt *daemonTunables) *appsv1.DaemonSet {
...
+
+	// Disable SELinux configuration when SELinux is not in enforcing mode
+	if selinux.EnforceMode() != selinux.Enforcing {
+		for i := range refSPOd.Spec.Template.Spec.InitContainers {
+			refSPOd.Spec.Template.Spec.InitContainers[i].SecurityContext.SELinuxOptions = nil
+		}
+		for i := range refSPOd.Spec.Template.Spec.Containers {
+			refSPOd.Spec.Template.Spec.Containers[i].SecurityContext.SELinuxOptions = nil
+		}
+	}
```
JAORMX, line 136: I don't think we should be doing this, instead you could just turn off SELinux in the SPO deployment.

JAORMX (changes requested):

ccojocar:
>  However, instead of turning SELinux options off; why not disable SELinux support in the SPO instead? We do have a configuration flag for this, and thus it wouldn't deploy the SELinux bits.

@JAORMX You misunderstood the fix. This is not related to SELinux container which manage the SELinux profile. I am aware that that option can be turned off and it is turned off by default. 

The issue is that `spc_t` is hardcoded in deployment for all containers part of SPOD daemon. Including the daemon container itself. This doesn't work on a Linux distribution which doesn't support this SELinux type. Also it doesn't make sense to be there if SELinux is not in enforcement mode on k8s nodes.

JAORMX:
> The issue is that `spc_t` is hardcoded in deployment for all containers part of SPOD daemon. Including the daemon container itself. This doesn't work on a Linux distribution which doesn't support this SELinux type.

I understand  this; in that sense, it makes sense not to require this or make this configurable in the SPO.

> Also it doesn't make sense to be there if SELinux is not in enforcement mode on k8s nodes.

I, however, don't agree with this statement. Turning SELinux to `enforcing` would be disruptive in deployments that do have `spc_t` and would require re-deploying the SPO as the pods would need to be relabeled. e.g. this would be problematic in Openshift. I wonder if we can come to a solution that doesn't cause this issue.

ccojocar:
> I understand this; in that sense, it makes sense not to require this or make this configurable in the SPO.

The initial solution in #826 was to make the security context of each container part of SPO configurable. I think there was a concern that this is exposing too much the API in the CRD.

JAORMX:
@ccojocar sorry for the confusion. I'll reach out to the other maintainers and we'll get this sorted. Thanks a lot for the contributions!

saschagrunert:
Started a discussion in https://kubernetes.slack.com/archives/C013FQNB0A2/p1647006231779659

saschagrunert (approved):

ccojocar:
Closing this in favour of https://github.com/kubernetes-sigs/security-profiles-operator/pull/850. 

The SELinux configuration cannot be detected in the manager container because it requires to mount the SELinux pseudo file system which will entitle some additional privileges which needs to be granted to the manager container. Discussion in https://kubernetes.slack.com/archives/C013FQNB0A2/p1647525440776799

PR #830, opened 2022-03-10
fix: health probe of seccomp controller should be ready if a node supports seccomp

ccojocar:

#### What type of PR is this?
bug 

This fix avoids an endless restart of seccomp container when there isn't any seccompprofile
defined into the cluster. In this situation the reconcile loop never gets executed, and the
seccomp container keeps being restarted by Kubernetes.

This is currently happening when the operator is initially installed.

The seccompprofile controller should be ready even if there is no  profile defined but the node supports seccomp.

```release-note

```

### DIFF
--- a/internal/pkg/daemon/seccompprofile/seccompprofile.go
+++ b/internal/pkg/daemon/seccompprofile/seccompprofile.go
@@ -39,7 +39,6 @@ import (
 
 	seccompprofileapi "sigs.k8s.io/security-profiles-operator/api/seccompprofile/v1beta1"
 	statusv1alpha1 "sigs.k8s.io/security-profiles-operator/api/secprofnodestatus/v1alpha1"
-	"sigs.k8s.io/security-profiles-operator/internal/pkg/atomic"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/controller"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/daemon/metrics"
@@ -88,7 +87,6 @@ type Reconciler struct {
 	record  event.Recorder
 	save    saver
 	metrics *metrics.Metrics
-	ready   atomic.Bool
 }
 
 // Name returns the name of the controller.
@@ -122,8 +120,20 @@ func (r *Reconciler) Setup(
 
 // Healthz is the liveness probe endpoint of the controller.
 func (r *Reconciler) Healthz(*http.Request) error {
-	if !r.ready.Get() {
-		return errors.New("not ready")
+	return r.checkSeccomp()
+}
+
+// checkSeccomp verifies if the seccomp is supported by the node
+func (r *Reconciler) checkSeccomp() error {
+	if !seccomp.IsSupported() {
+		err := fmt.Errorf("node %q does not support seccomp", os.Getenv(config.NodeNameEnvKey))
+		if r.record != nil {
+			r.metrics.IncSeccompProfileError(reasonSeccompNotSupported)
+			r.record.Event(&seccompprofileapi.SeccompProfile{},
+				event.Warning(reasonSeccompNotSupported, err, os.Getenv(config.NodeNameEnvKey),
+					"node does not support seccomp"))
+		}
+		return err
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


JAORMX:
@ccojocar I thought the SPO was installing some base seccomp profiles as part of the setup process. I'm wondering why this is not happening... This might be an installation issue.

ccojocar:
> I thought the SPO was installing some base seccomp profiles as part of the setup process. I'm wondering why this is not happening... This might be an installation issue.

@JAORMX I don't see any base seccomp profile as being part of the default installation. Anyhow I don't think that the readiness of the operator should depend on a default profile being installed into the cluster.  What if that profile doesn't meet the requirements of the cluster operator? Does the cluster operator still need to install that profile just to have the operator properly running, any maybe afterwards to delete it?

JAORMX:
> > I thought the SPO was installing some base seccomp profiles as part of the setup process. I'm wondering why this is not happening... This might be an installation issue.
> 
> @JAORMX I don't see any base seccomp profile as being part of the default installation. Anyhow I don't think that the readiness of the operator should depend on a default profile being installed into the cluster. What if that profile doesn't meet the requirements of the cluster operator? Does the cluster operator still need to install that profile just to have the operator properly running, any maybe afterwards to delete it?

you are right, we are not installing a base seccomp profile by default. It is a good practice that an admin would install one, but we shouldn't be relying on it for health checks. Given this, I think this PR is appropriate and it LGTM. Thanks for taking the time to fix this!

JAORMX (approved):

ccojocar:
I am not sure whether the failure of `e2e-fedora` test is related to this change:
```
        --- PASS: TestSuite/TestSecurityProfilesOperator/cluster-wide:_SELinux:_base_case_(install_policy,_run_pod_and_delete) (82.75s)
    --- FAIL: TestSuite/TestSecurityProfilesOperator/namespaced:_SELinux:_base_case_(install_policy,_run_pod_and_delete) (72.45s)
```

It seems that only the namespaced version of the test fails. Do you have an idea what's going on? Thanks

JAORMX:
@ccojocar the test fails since the policy is installed but takes too long to remove (or doesn't get removed at all). Are we seeing this same issue in other PRs? cc @jhrozek

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -88,7 +87,6 @@ type Reconciler struct {
 	record  event.Recorder
 	save    saver
 	metrics *metrics.Metrics
-	ready   atomic.Bool
```
saschagrunert, line 91: The idea of this indicator was to become ready once one successful reconciliation loop has been done. We indeed install a bunch of default profiles via the manager, which then should get reconciled by the daemon set: https://github.com/kubernetes-sigs/security-profiles-operator/blob/664598434c3c896d6a69a35f8313a07abdcdf482/internal/pkg/manager/spod/bindata/default_profiles.go#L28

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -122,8 +120,21 @@ func (r *Reconciler) Setup(
...
-	if !r.ready.Get() {
-		return errors.New("not ready")
+	return r.checkSeccomp()
+}
+
+// checkSeccomp verifies if the seccomp is supported by the node.
+func (r *Reconciler) checkSeccomp() error {
+	if !seccomp.IsSupported() {
+		err := errors.New("seccomp not supported")
+		err = fmt.Errorf("node %q: %w", os.Getenv(config.NodeNameEnvKey), err)
```
saschagrunert, line 130: nit: I'd merge both lines into something like 
```suggestion
		err := fmt.Errorf("seccomp not supported on node %q", os.Getenv(config.NodeNameEnvKey))
```

ccojocar: They were merged together in the previous commit but it was a lint warning. I split them in order to fix that warning. See https://github.com/kubernetes-sigs/security-profiles-operator/runs/5498882895?check_suite_focus=true#step:5:114

internal/pkg/daemon/seccompprofile/seccompprofile.go
ccojocar: I saw this profile being sometimes installed. Somehow it doesn't get always installed which leads to the failure mentioned above. 

I would rather not have this profile installed by default because we don't have any nginx running in the cluster.

PR #833, opened 2022-03-11
feat: add a list in profilerecording which allows to define specific containers

ccojocar:
- fix: extend the profilerecording with a container list
- feat: annotate only specific containers in case they are defined in the profilerecording

```release-note

```


internal/pkg/webhooks/utils/utils.go
```diff
@@ -45,6 +45,16 @@ func RemoveIfExists(list []string, item string) []string {
 	return list
 }
 
+// Exists checks an item is present in the provided list.
+func Exists(list []string, item string) bool {
```
saschagrunert, line 49: We already have this function in https://github.com/kubernetes-sigs/security-profiles-operator/blob/dd72d74e37a41afa7084d9bf680a3b9dcc95666e/internal/pkg/util/names.go#L36-L44

saschagrunert (commented):
I think that's actually a good feature! Do you think we can make it work in the same way for the log enricher as well as the bpf recorder?

api/profilerecording/v1alpha1/profilerecording_types.go
```diff
@@ -55,6 +55,12 @@ type ProfileRecordingSpec struct {
 	// label selector semantics. An empty podSelector matches all pods in this
 	// namespace.
 	PodSelector metav1.LabelSelector `json:"podSelector"`
+
+	// Containers a set of containers to record. This allows to select
```
saschagrunert, line 59: nit:
```suggestion
	// Containers is a set of containers to record. This allows to select
```

internal/pkg/webhooks/utils/utils.go
ccojocar: I removed the duplicated function and used the one you are mentioning here. Thanks for pointing this out.

api/profilerecording/v1alpha1/profilerecording_types.go
ccojocar: Fixed the comment

ccojocar:
> I think that's actually a good feature! Do you think we can add an e2e test for the bpf and log enricher recorder?

I can add an E2E test.

ccojocar:
@saschagrunert I added the E2E tests. Please could you have a look again? Thanks!

test/tc_profilerecordings_test.go
```diff
@@ -215,6 +223,32 @@ func (e *e2e) profileRecordingMultiContainer(
...
+
+	e.kubectl("delete", "pod", podName)
+
+	const profileNameNginx = recordingName + "-nginx"
+	profileNginx := e.retryGetSeccompProfile(profileNameNginx)
+	e.Contains(profileNginx, "close")
+
+	const profileNameRedis = recordingName + "-redis"
+	profileRedis := e.retryGetSeccompProfile(profileNameRedis)
+	e.Empty(profileRedis)
```
saschagrunert, line 246: We probably should check that redis has not been recorded here.

ccojocar:
@saschagrunert e2e tests passed now. Please could you have another look? Thank you!

saschagrunert (approved):
Thanks!

PR #846, opened 2022-03-18
Add ability to patch "problematic" pods

JAORMX:
#### What type of PR is this?

/kind feature
/kind design
/kind api-change

It's often useful to know what pods are having issues with policies (be
it SELinux or Seccomp). To enable administrators to easily fetch pods
with denials registered to them, this adds the ability for the log
enricher to detect this and label a pod appropriately with the following
label:

```
spo.x-k8s.io/had-denials: ''
```

By relying on a label, we're able to easily fetch the pods with a call
such as follows:

```
kubectl get pods -A -l spo.x-k8s.io/had-denials=
```

None

#### Does this PR have test?

Yes

```release-note
This PR includes the ability to tag pods that present denials from either Seccomp or SELinux. This will happen through the 'spo.x-k8s.io/had-denials' label.
```

### DIFF
--- a/api/spod/v1alpha1/spod_types.go
+++ b/api/spod/v1alpha1/spod_types.go
@@ -68,6 +68,10 @@ type SPODSpec struct {
 	// tells the operator whether or not to enable AppArmor support for this
 	// SPOD instance.
 	EnableAppArmor bool `json:"enableAppArmor,omitempty"`
+	// tells the operator whether or not to apply labels to pods that present
+	// security policy-related denials. Note that this will be done cluster-wide.
+	// Note that this currently requires the log enricher to be enabled.
+	LabelPodDenials bool `json:"labelPodDenials,omitempty"`
 	// If specified, the SPOD's tolerations.
 	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -66,6 +66,7 @@ const (
 	jsonFlag           string = "json"
 	selinuxFlag        string = "with-selinux"
 	apparmorFlag       string = "with-apparmor"
+	labelDenialsFlag   string = "label-denials"
 	defaultWebhookPort int    = 9443
 )
 
@@ -172,6 +173,13 @@ func main() {
 			Name:    "log-enricher",
 			Aliases: []string{"l"},
 			Usage:   "run the audit's log enricher",
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:  labelDenialsFlag,
+					Usage: "whether to label problematic pods or not",
+					Value: false,
+				},
+			},
 			Action: func(ctx *cli.Context) error {
 				return runLogEnricher(ctx, info)
 			},
@@ -435,10 +443,15 @@ func runBPFRecorder(_ *cli.Context, info *version.Info) error {
 	return bpfrecorder.New(ctrl.Log.WithName(component)).Run()
 }
 
-func runLogEnricher(_ *cli.Context, info *version.Info) error {
+func runLogEnricher(clictx *cli.Context, info *version.Info) error {
 	const component = "log-enricher"
 	printInfo(component, info)
-	return enricher.New(ctrl.Log.WithName(component)).Run()
+
+	e, err := enricher.New(ctrl.Log.WithName(component), clictx.Bool(labelDenialsFlag))
+	if err != nil {
+		return fmt.Errorf("building log enricher: %w", err)
+	}
+	return e.Run()
 }
 
 func runNonRootEnabler(_ *cli.Context, info *version.Info) error {
--- a/internal/pkg/daemon/enricher/container.go
+++ b/internal/pkg/daemon/enricher/container.go
@@ -22,6 +22,11 @@ import (
 	"time"
 
 	"github.com/ReneKroon/ttlcache/v2"
+	"github.com/go-logr/logr"
+	"golang.org/x/net/context"
+	"golang.org/x/sync/errgroup"
+	v1 "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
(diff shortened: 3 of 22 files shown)


internal/pkg/daemon/enricher/container.go
```diff
@@ -124,6 +120,10 @@ func (e *Enricher) getContainerInfo(
 						recordProfile: recordProfile,
 					}
 
+					if e.labelProblematicContainers {
+						go labelContainer(e.impl, e.clientset, *info, e.logger)
```
saschagrunert, line 124: We can make this a method to avoid passing `*e`'s variables. :)

saschagrunert, line 124: I think this can be racy. If a pod is already labeled or about to be labeled it should be skipped.

cmd/security-profiles-operator/main.go
```diff
@@ -154,6 +155,14 @@ func main() {
...
 			Usage:   "run the audit's log enricher",
 			Action:  runLogEnricher,
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "label-problematic",
+					Aliases: []string{"l"},
+					Value:   labelProblematicContainers,
+					Usage:   "the port on which to expose the webhook service (default 9443)",
+				},
+			},
```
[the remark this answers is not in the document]
JAORMX: right! That's the intent, I haven't submitted that part.

internal/pkg/daemon/enricher/container.go
JAORMX: true! I'll do that

cmd/security-profiles-operator/main.go
```diff
@@ -154,6 +155,14 @@ func main() {
 			Aliases: []string{"l"},
 			Usage:   "run the audit's log enricher",
 			Action:  runLogEnricher,
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:    "label-problematic",
+					Aliases: []string{"l"},
+					Value:   labelProblematicContainers,
+					Usage:   "the port on which to expose the webhook service (default 9443)",
```
jhrozek, line 163: typo

cmd/security-profiles-operator/main.go
```diff
@@ -415,7 +424,11 @@ func runBPFRecorder(ctx *cli.Context) error {
 func runLogEnricher(ctx *cli.Context) error {
 	const component = "log-enricher"
 	printInfo(component)
-	return enricher.New(ctrl.Log.WithName(component)).Run()
+	e, err := enricher.New(ctrl.Log.WithName(component), labelProblematicContainers)
+	if err != nil {
+		return errors.Wrap(err, "getting log enricher")
```
jhrozek, line 429: copy-paste error

internal/pkg/daemon/enricher/container.go
```diff
@@ -124,6 +116,10 @@ func (e *Enricher) getContainerInfo(
 						recordProfile: recordProfile,
 					}
 
+					if e.labelProblematicContainers {
+						go e.impl.LabelProblematicPod(e.clientset, info.containerName, info.podName, info.namespace, e.logger)
```
jhrozek, line 120: Do you think it would decrease the noise to only label pods that are not being recorded? Or do you think that using this feature is more or less exclusive with recording? (Either way, nothing that should block this patch, just thinking out loud).

For future I guess we could have a simple namespace selector or even a list of NSs in case we want to restrict what pods are labeled this way (here I'm thinking about that use-case with watching just-recorded policies for additional AVCs...)

jhrozek (commented):
nice work, some nits and a question inline

internal/pkg/daemon/enricher/container.go
JAORMX: So, this currently tails audit logs, so it will label ANY pod running on a given node. We could start limiting that with a label selector, but that would only take effect until after the pod has been parsed from the node itself.

jhrozek (commented):
Sascha is better equipped to review the code, but I like the approach.
I would prefer if the code was split into refactoring and adding functionality next time though :-)
Do I read the code correctly that enabling the new boolean also enables enricher? If yes, you might want to spell this out in the help text and/or the docs.
Speaking of docs, amending installation-usage.md would be nice.

JAORMX:
@jhrozek I'll add some docs about this.

Also, enabling the `labelProblematicPods` flag does not enable the enricher. The enricher needs to be enabled as a pre-requisite.

api/spod/v1alpha1/spod_types.go
```diff
@@ -68,6 +68,10 @@ type SPODSpec struct {
 	// tells the operator whether or not to enable AppArmor support for this
 	// SPOD instance.
 	EnableAppArmor bool `json:"enableAppArmor,omitempty"`
+	// tells the operator whether or not to apply labels to pods that present
+	// security policy-related denials. Note that this will be done cluster-wide.
+	// Note that this currently requires the log enricher to be enabled.
+	LabelProblematicPods bool `json:"labelProblematicPods,omitempty"`
```
saschagrunert, line 74: Can we be more specific on the word "Problematic"? Like, "LabelPodDenials" or something similar?

internal/pkg/daemon/enricher/container.go
```diff
@@ -53,102 +62,139 @@ func (e *Enricher) getContainerInfo(
...
-		containerID,
-	); !errors.Is(err, ttlcache.ErrNotFound) {
-		cInfo, ok := info.(*containerInfo)
-		if !ok {
-			return nil, errors.New("cache value is not container info")
+func (e *Enricher) populateCacheEntryForContainer(
+	ctx context.Context, targetContainerID string, pod *v1.Pod, eg *errgroup.Group,
+) {
+	eg.Go(func() error {
+		var errorToRetry error
```
saschagrunert, line 125:
```suggestion
	eg.Go(func() (errorToRetry error) {
```

internal/pkg/daemon/enricher/impl.go
```diff
@@ -108,13 +112,61 @@ func (d *defaultImpl) NewForConfig(
...
+			return !kerrors.IsNotFound(inErr)
+		},
+	); err != nil {
+		l.Error(
+			err, "unable to patch container to mark it as problematic",
+			"pod.Namespace", pod.GetNamespace(), "pod.Name", pod.GetName(),
+			"containerName", containerName,
+		)
+	}
+}
```
saschagrunert, line 153: I would add that method to the enricher, since it looks like we only have `doLabelProblematicPodPod` to mock.

internal/pkg/daemon/enricher/impl.go
```diff
@@ -108,13 +112,61 @@ func (d *defaultImpl) NewForConfig(
...
+) error {
+	labels := pod.GetLabels()
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+	labels[problematicContainerLabelKey] = ""
+	pod.SetLabels(labels)
+	_, err := c.CoreV1().Pods(pod.GetNamespace()).Update(
+		ctx, pod, metav1.UpdateOptions{},
+	)
```
saschagrunert, line 166: I think this is the only method which requires mocking and therefore should be part of the `defaultImpl`.

saschagrunert (commented):
Cool! I think we have to agree on a naming here and I have a bunch of nits regarding the interface mocking. Otherwise LGTM

JAORMX:
@saschagrunert thanks for the suggestions! Fixed

saschagrunert (approved):
LGTM, PTAL @jhrozek

jhrozek (approved):
The only remaining comment I have is to amend the docs. But that can be done afterwards so now I'm adding
/lgtm

PR #850, opened 2022-03-21
fix: add an environment variable for disabling the selinux option from security context of SPOd deployment

ccojocar:

This adds an environment variable for disabling the the selinux option form security context of SPOd deployment. This is required for Linux distributions which doesn't enforce SELinux by default and do not provide the `spc_t` type label (e.g. Flatcar Linux). 

See other solutions proposed in:

https://github.com/kubernetes-sigs/security-profiles-operator/pull/828
https://github.com/kubernetes-sigs/security-profiles-operator/pull/826

```release-note
Add SPOD_SELINUX_DISABLE environment variable for disabling the the SELinux option from security context of SPOd deployment

```

### DIFF
--- a/deploy/overlays/flatcar/deployment.yaml
+++ b/deploy/overlays/flatcar/deployment.yaml
@@ -1,0 +1,13 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: security-profiles-operator
+  namespace: security-profiles-operator
+spec:
+  template:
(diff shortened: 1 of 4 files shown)


saschagrunert (approved):
/ok-to-test

JAORMX:
we agreed on slack to change the approach to configure this via the SPOD object as opposed to using an env variable.

PR #882, opened 2022-04-01
Make `$(BUILD_DIR)/kustomize` target PHONY

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now manually check if the binary is available because its execution seems to modify it, which means that it will attempt to load it again.

None
#### Does this PR have test?
None

None

```release-note
None
```


jhrozek (approved):
yay
/lgtm

PR #883, opened 2022-04-04
Make webhooks' failurePolicy and namespaceSelector configurable

### LINKED ISSUE #672: Consider restricting webhook namespaces to a subset
At the moment, our webhooks affect all namespaces. It would be nice if we could restrict the webhooks to a subset of namespaces instead. This could be done in the manifests only, by taking the namespaceSelector and/or objectSelector into use.

#### Why is this needed:
I think that unless the administrator explicitly says so, mutating webhooks shouldn't affect the whole cluster, especially if the failurepolicy is set to FAIL.

Note: this might be distribution-specific and I'm fine adding the namespaceSelector to the openshift kustomize files. Maybe for the generic case, we can use the namespace-scope way of running SPO. But I suspect that even for profileBindings, having a way to set e.g. all namespaces labeled with "apps" get to set bindings but no others might be useful.

#### User story covered
I don't think we have one, but I also think that 'overall cluster stability" is kindof an implicit one.

jhrozek:
#### What type of PR is this?
/kind feature

By default, our webhooks target all namespaces. While from a security
standpointm this is OK, because the webhooks react to our CRs and
therefore an explicit RBAC rule must be added to manage them, there is
still a risk that if there is a bug in the webhooks, the function of
critical namespaces might be affected.

This patch exposes the failureMode and the namespaceSelector as
configurable through a newly added webhookOptions object in the spod CR
to enable admins to select which namespaces should be watched by the
webhook and what should happen on failure.

In addition, namespace runlevels 0 and 1 are ignored by defualt on OCP.

Fixes #672

#### Does this PR have test?
Yes, an e2e test was added.

I'll point a question inline.

```release-note
The spod CR gains a new field webhookOptions which allows the webhooks' failurePolicy and namespaceSelector to be configurable.
```


installation-usage.md
```diff
@@ -932,6 +932,29 @@ security_profiles_operator_seccomp_profile_audit_total{container="log-container"
...
+
+In order to change webhook's configuration, the `spod` CR exposes an object `webhookOptions` that allows
+the `failurePolicy` and the `namespaceSelector` to be set. This way you can set the webhooks to "soft-fail"
+or restrict them to a subset of a namespaces so that even if the webhooks had a bug that would prevent them
+from running at all, other namespaces or resources wouldn't be affected.
+
+For example, to set the `binding.spo.io` webhook's configuration to ignore errors as well as restrict it
+to a subset of namespaces labeled with `spo.x-k8s.io/enable-binding=true`, call:
+```shell
+$ kubectl -nsecurity-profiles-operator patch spod spod -p '{"spec":{"webhookOptions":[{"name":"binding.spo.io","failurePolicy":"Ignore","namespaceSelector":{"matchExpressions":[{"key":"spo.x-k8s.io/enable-binding","operator":"In","values":["true"]}]}}]}}' --type=merge
```
saschagrunert, line 951: Do we wanna multiline this?

installation-usage.md
```diff
@@ -932,6 +932,29 @@ security_profiles_operator_seccomp_profile_audit_total{container="log-container"
 security_profiles_operator_seccomp_profile_audit_total{container="log-container",executable="/usr/sbin/nginx",namespace="default",node="127.0.0.1",pod="log-pod",syscall="write"} 20
 ```
 
+## Configuring webhooks
```
saschagrunert, line 935: The linter won't be happy with this change without changing the table of contents.

jhrozek: Thanks, done

installation-usage.md
jhrozek: I actually added a YAML patch because I always found the JSON patches quite unreadable. How does that look?

saschagrunert (approved):
Great, thank you!

jhrozek:
oops, I can reproduce the issue locally. Stupid tests, finding issues in my code..

saschagrunert (approved):

PR #893, opened 2022-04-13
nodestatus: Only list nodestatuses from the same namespace as the profile

jhrozek:
#### What type of PR is this?
/kind bug

In case there were two profiles with identical names in two namespaces,
the operator would list them all, from both namespaces and then fail
with:

nodestatus.go:179 nodestatus "msg"="Not updating policy: not all statuses are ready"
    "namespace"="default"
    "nodeStatus"="sleep-sh-pod-ip-10-0-164-225.us-east-2.compute.internal"
    "has"=18 "wants"=6

note that has > wants.

This patch adjusts the list so that only nodestatuses from the same
namespace are listed.

None

#### Does this PR have test?
Yes

None

```release-note
NONE
```

### DIFF
--- a/internal/pkg/manager/nodestatus/nodestatus.go
+++ b/internal/pkg/manager/nodestatus/nodestatus.go
@@ -155,7 +155,7 @@ func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 		return reconcile.Result{}, errors.New("status doesn't match owner")
 	}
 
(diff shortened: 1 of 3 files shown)


jhrozek:
oops, looks like I broke something :-/
/hold

jhrozek:
/hold cancel
this should be ready for another round...

saschagrunert (approved):

PR #896, opened 2022-04-19
Fix missing key/value for profile not found log message

saschagrunert:
#### What type of PR is this?

/kind cleanup

The log message should not print `MISSING` which can be fixed by using an arbitrary name as key.

None
#### Does this PR have test?
None

None

```release-note
None
```


jhrozek (approved):
/lgtm

PR #898, opened 2022-04-19
Switch from ring to performance buffer

saschagrunert:
#### What type of PR is this?

/kind feature

To support older kernels we now use the performance buffer ebpf type rather than the ring buffer.

Fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/795
#### Does this PR have test?

None
None

```release-note
Using the ebpf performance buffer type rather than the ring buffer to support older Linux kernel versions.
```


saschagrunert:
Hm, now I'm getting the invalid argument error here :thinking:

JAORMX (approved):

JAORMX (changes requested):
uhm...

JAORMX:
> Hm, now I'm getting the invalid argument error here 🤔

in which logs is that?

saschagrunert:
> > Hm, now I'm getting the invalid argument error here thinking
> 
> in which logs is that?

In the e2e tests, I'm also not able to run the ebpf program locally.

saschagrunert:
I'm not sure this feels like a downgrade to me. The fact that older kernel versions do not support the ringbuffer is something we may auto-fix in the future if those systems switch to newer kernels.

PR #910, opened 2022-04-26
e2e: Various stabilization fixes for our test suite

jhrozek:
#### What type of PR is this?
/kind bug
/kind flake

Since selinuxd switched to using policycoreutils, it is a bit slower.
Let's increase the number of iterations we wait for a policy to be
removed in hopes of getting a more stable CI environment.

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```

### DIFF
--- a/internal/pkg/daemon/enricher/container.go
+++ b/internal/pkg/daemon/enricher/container.go
@@ -121,7 +121,7 @@ func (e *Enricher) populateContainerPodCache(
 }
 
 func (e *Enricher) populateCacheEntryForContainer(
-	ctx context.Context, targetContainerID string, pod *v1.Pod, eg *errgroup.Group,
+	_ context.Context, targetContainerID string, pod *v1.Pod, eg *errgroup.Group,
 ) {
 	eg.Go(func() (errorToRetry error) {
 		// nolint:gocritic // This is what we expect and want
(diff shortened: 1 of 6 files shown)


saschagrunert (approved):

jhrozek:
Going to try one more patch, hold in the meantime, I'm going to unhold when the PR is ready for review
/hold

jhrozek:
Turning this into a PR with several assorted e2e fixes, going to unhold the PR when I'm done playing whack-a-mole.

JAORMX (approved):

jhrozek:
fedora seems to be succeeding. flatcar failed with on the seccomp-metrics test which is still a little flaky. e2e (kind) based with a transient error where a pod didn't start.

I'm going to just close and reopen the PR to force the CI I think the situation has improved quite a bit with the patches.

jhrozek:
OK, this is much better. @saschagrunert @JAORMX PTAL

Note that the metrics test is sometimes still a bit wonky, but that hits the kind-based e2e test mostly which can at least be re-run.
The reason is that the metrics service frontends the spod DS instances and for some reason (which I haven't figured out yet..) each pod in the DS reports a different metric value.

saschagrunert (approved):
Thank you! :pray:

PR #913, opened 2022-04-29
feat: add a syscall allow list in the SPOD configuration

### LINKED ISSUE #897: Enforce cluster-wide a security baseline for seccomp profile installation

In a cluster where different tenants can install seccomp profiles in different namespaces, it is paramount to enforce cluster-wide 
a security baseline for seccomp profile installation. This will prevent that the tenants will introduce system calls which are not allowed by the cluster operators in order to avoid container escapes and such. 

It should be possible to install only a subset of the baseline. For example, a seccomp profile should only be successfully installed when it contains a subset of the system calls available in the baseline profile (aka cluster default seccomp profile).

#### Why is this needed:

This will allow a cluster administrator to enforce which system calls are allowed to be enabled in the cluster via the operator.  This will prevent from introducing dangerous system calls which might lead to container escapes.

We want to prevent that someone extends the seccomp profile installed in a namespace beyond the cluster default profile(https://kubernetes.io/blog/2021/08/25/seccomp-default/#seccompdefault-to-the-rescue).

ccojocar:

Add a syscall allow list in the SPOD configuration. This is to allow an administrator to limit cluster-wide the syscalls which are allowed to be defined in the seccomp profiles. 

Fixes #897

#### Does this PR have test?

Yes 

Yes

```release-note

Add a syscall allow list in the SPOD configuration
```


saschagrunert:
@ccojocar may I ask you to give this one a rebase?

ccojocar:
@saschagrunert All tests passed after rebasing, please could you have a look? Thanks

test/tc_allowed_syscalls_test.go
```diff
@@ -0,0 +1,251 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
saschagrunert, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

saschagrunert (commented):
Hm, the code changes look good from what I can see. 

Can we add docs to the installation and usage guide to describe the feature? I'm curious how it behaves if we change the allowed syscalls for already deployed profiles (we cannot remove them when they're in usage).

ccojocar:
Thanks for review @saschagrunert! I added some documentation which explains how to use the allow syscalls list. 

 >  I'm curious how it behaves if we change the allowed syscalls for already deployed profiles (we cannot remove them when they're in usage).

There is an e2e test for this, namely `testCaseAllowedSyscallsInUse` which essentially removes some syscall from the spod config at runtime while one of the non compliant profile is in use. This will result in the operator removing the profile as soon as the pod referencing the profile terminates.

saschagrunert (approved):
LGTM

PR #916, opened 2022-05-03
metrics: Add SELinux related audit metrics to the enricher

### LINKED ISSUE #609: SELinux recording metrics are missing
#### What happened:
The SELinux recording feature, especially in the enricher in missing metrics.

We should mirror the seccomp code path and have metrics

record a selinux profile

N/A

#### Environment:
N/A

jhrozek:
#### What type of PR is this?
/kind feature

On dispatching an AVC line from the enricher via GRPC, also issue a
metric with the details of the AVC, similarly to what we do in the
seccomp case.

Fixes #609

#### Does this PR have test?
No, I'm not sure if our way of getting metrics through the service
is stable enough for tests.

N/A

```release-note
The security_profiles_operator_selinux_profile_audit_total metric was actually
enabled and uses the appropriate labels scraped from the audit.log file.
```


internal/pkg/daemon/metrics/grpc.go
```diff
@@ -97,22 +97,31 @@ func (m *Metrics) AuditInc(
 
 		switch r.GetType() {
 		case api.AuditRequest_SECCOMP:
+			if r.GetSeccompReq() == nil {
+				return fmt.Errorf("missing seccomp request details")
```
saschagrunert, line 101:
```suggestion
				return errors.New("missing seccomp request details")
```

internal/pkg/daemon/enricher/enricher.go
```diff
@@ -421,6 +421,27 @@ func (e *Enricher) dispatchSelinuxLine(
 		"tclass", auditLine.tclass,
 	)
 
+	metricsType := apimetrics.AuditRequest_SELINUX
+	if err := e.impl.SendMetric(
+		metricsClient,
+		&apimetrics.AuditRequest{
+			Type:       metricsType,
```
saschagrunert, line 428:
```suggestion
			Type:       apimetrics.AuditRequest_SELINUX,
```

internal/pkg/daemon/metrics/grpc.go
```diff
@@ -97,22 +97,31 @@ func (m *Metrics) AuditInc(
...
 				r.GetNamespace(),
 				r.GetPod(),
 				r.GetContainer(),
 				r.GetExecutable(),
-				r.GetSyscall(),
+				r.GetSeccompReq().GetSyscall(),
 			)
 		case api.AuditRequest_SELINUX:
+			if r.GetSelinuxReq() == nil {
+				return fmt.Errorf("missing selinux request details")
```
saschagrunert, line 113:
```suggestion
				return errors.New("missing selinux request details")
```

api/grpc/metrics/api.proto
```diff
@@ -29,13 +29,23 @@ message AuditRequest {
...
+    message SeccompAuditReq {
+        string syscall = 1;
+    }
+    message SelinuxAuditReq {
+        string perm = 1;
+        string scontext = 2;
+        string tcontext = 3;
+        string tclass = 4;
+    }
     Type type = 1;
```
saschagrunert, line 41: In theory we could remove the type in favor of checking those against `nil`:
```
SeccompAuditReq seccompReq = 7;
SelinuxAuditReq selinuxReq = 8;
```

internal/pkg/daemon/metrics/grpc.go
jhrozek: thanks

api/grpc/metrics/api.proto
jhrozek: yes, good idea

jhrozek:
Thank you for the review @saschagrunert

internal/pkg/daemon/metrics/metrics.go
```diff
@@ -153,7 +157,10 @@ func New() *Metrics {
 				metricsLabelPod,
 				metricsLabelContainer,
 				metricsLabelExecutable,
-				metricsLabelSyscall,
+				metricsLabelPerms,
+				metricsLabelScontext,
+				metricsLabelTcontext,
+				metricsLabelTclass,
```
JAORMX, line 163: wouldn't doing this dramatically increase the cardinality of this metric? I really think we shouldn't do this...

JAORMX (changes requested):
I'm concerned about this dramatically increasing the cardinality of the SELinux metrics...

From [an article on the topic](https://www.robustperception.io/cardinality-is-key)

> As a general rule of thumb I'd avoid having any metric whose cardinality on a /metrics could potentially go over 10 due to growth in label values. The way I think of it is that if it has already grown to be 10 today, it might be 15 in a year's time. It can be additionally okay to have no more than a literal handful within a given Prometheus that go to 100. These are only rules of thumb, if you know for example that the number of replicas of your service will always be tiny then you can exceed these a bit, conversely if you're expecting thousands of replicas I'd carefully consider every time series on the /metrics.

What are you trying to get with these metrics? Maybe there can be another solution instead of adding these new labels.

internal/pkg/daemon/metrics/metrics.go
jhrozek: What real-world implication does increasing the cardinality have?

jhrozek:
> I'm concerned about this dramatically increasing the cardinality of the SELinux metrics...
> 
> From [an article on the topic](https://www.robustperception.io/cardinality-is-key)
> 
> > As a general rule of thumb I'd avoid having any metric whose cardinality on a /metrics could potentially go over 10 due to growth in label values. The way I think of it is that if it has already grown to be 10 today, it might be 15 in a year's time. It can be additionally okay to have no more than a literal handful within a given Prometheus that go to 100. These are only rules of thumb, if you know for example that the number of replicas of your service will always be tiny then you can exceed these a bit, conversely if you're expecting thousands of replicas I'd carefully consider every time series on the /metrics.
> 
> What are you trying to get with these metrics? Maybe there can be another solution instead of adding these new labels.

Have a way of observing what is being recorded. There's two reasons:
 - for the recording itself
 - I'm still thinking about "watching" workloads with a policy when introducing the policy into production and I was thinking about using the log enricher so that you could see what AVCs does the workload trigger and can adjust either the workload or the policy. 

If the number of labels is a concern, I guess we could remove /some/ but at least the source and target contexts should stay so that you could see what is touching what and go to see the full log enricher logs.

jhrozek:
btw the problem with "watching log enricher logs" is that each spod instance has their own. The metrics allow you to watch all of them from a central place.

internal/pkg/daemon/metrics/metrics.go
JAORMX: The main issue (real-world implication) is that increasing cardinality of metrics will lead to overloading Prometheus. Thus, by not adding only what's minimally needed for us to debug problems at scale we make the SPO a noisy neighbour in terms of metrics

saschagrunert (approved):
/hold for @JAORMX

jhrozek:
OK, I removed the perm and the class. This still retains scontext and tcontext which should give an idea of who is touching what and brings the amount of labels to just one more than the already existing seccomp metric

JAORMX:
> OK, I removed the perm and the class. This still retains scontext and tcontext which should give an idea of who is touching what and brings the amount of labels to just one more than the already existing seccomp metric

I think this is a good compromise. Thanks!

JAORMX (approved):

jhrozek:
ugh another CI failure, let me reopen the PR..

saschagrunert (approved):

PR #923, opened 2022-05-06
fix: remove the control-plane taint to allow calico controller to be scheduled

ccojocar:

```release-note

```


saschagrunert (approved):
Awesome, thank you!

JAORMX (approved):

ccojocar:
It seems that two e2e tests failed after removing the taint. This reason is that they were never executed before because the list of nodes provided to the tests was always empty due to the taint. They should should run now and complete successfully.

jhrozek (approved):
Thank you very much for fixing the tests so quickly!
/lgtm

PR #944, opened 2022-05-23
deploy: Annotate the NS to allow running privileged

jhrozek:
#### What type of PR is this?
/kind bug

On some distributions, like OCP 4.11, where PSA is enabled and the
restricted PSA is the default, pods must either run with a very limited
privileges or the namespace must be annotated to allow privileged pods.

Because SPO can't run with such limited privileges (for example selinuxd
or the non-root-enabler must run as root), we need to take the second
option and annotate the namespace.

None

#### Does this PR have test?
N/A

Should I also amend the docs?

```release-note
The security-profiles-operator namespace is now labeled with the following labels:
    pod-security.kubernetes.io/enforce: privileged
    pod-security.kubernetes.io/audit: privileged
    pod-security.kubernetes.io/warn: privileged
To account for clusters that are enabling PSA and defaulting to the restricted one.

When using another namespace or creating the namespace with other means,
please ensure that the namespace has the above labels.
```

### DIFF
--- a/deploy/base/ns.yaml
+++ b/deploy/base/ns.yaml
@@ -3,3 +3,7 @@ apiVersion: v1
 kind: Namespace
 metadata:
   name: security-profiles-operator 
+  labels:
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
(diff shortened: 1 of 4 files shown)


JAORMX (approved):
/lgtm

PR #988, opened 2022-06-08
examples: Use a nonprivileged nginx pod as well

jhrozek:
#### What type of PR is this?
/kind cleanup
/kind failing-test

We use the example in the e2e tests, but the image we used was a
privileged one, which doesn't work by default on OCP.

N/A

#### Does this PR have test?
N/A

I wasn't sure if this is the proper fix, becuase the policy used in
the example was created for the privileged container in the first
place. Alternatives:
 - use a custom workload in the test and just don't test the example
 - Decode() the YAML and overwrite the image on OCP only

```release-note
NONE
```

### DIFF
--- a/examples/pod.yaml
+++ b/examples/pod.yaml
@@ -17,4 +17,4 @@ spec:
       # seccompProfile:
         # type: Localhost
         # localhostProfile: operator/security-profiles-operator/nginx-1.19.1.json
-    image: nginx:1.19.1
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

jhrozek:
Closing per github discussion. Will either decode the pod and set the SA or something else so that we keep a test with the privileged container.

PR #1015, opened 2022-07-01
tests: Modify the example pod if running on OCP

jhrozek:
#### What type of PR is this?
/kind cleanup
/kind failing-test

#### What this PR does / why we need it:
The test pod in testCaseRunPod uses a container running as root. This
doesn't work by default on OCP, where the security defaults force all
pods to run as non-root regardless of what their Dockerfile says.

As pointed out in another PR that tried to use a non-root container, there
is value in testing a root container as well. To solve this, this patch
adds an openshift-specific annotation, forces the pod to run as root and
in order for all this to be allowed, sets the SA that binds to a role
that allows using privileged SCCs.

Tested on OCP 4.11.

#### Which issue(s) this PR fixes:
None

#### Does this PR have test?
N/A

#### Special notes for your reviewer:
N/A

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #1016, opened 2022-07-01
tests: Add the recording RBAC roles and SA dynamically, re-add on every operator redeploy

jhrozek:
#### What type of PR is this?
/kind bug
/kind failing-test

No idea why we didn't catch this before when running the RBAC roles and
SA on OCP before, but redeploying the operator wouldn't re-add the RBAC
role, role binding and the SA required to grant the ability to use the
SCC needed to record replicated controllers.

This patch adds the needed manifests as part of deploying the operator.

Note that most of the code is only executed on OCP, on other distros,
only the SA is added so that we can always reference the SA regardless
of the distribution.

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```


saschagrunert (approved):

JAORMX (approved):

jhrozek:
/hold
for some reason the tests work fine for me locally, wtf. Submitted some extra debugging and marking as hold

jhrozek:
this PR now superseded PR #1015

jhrozek:
@saschagrunert can you review this PR, please?

PR #1047, opened 2022-07-26
Update Ubuntu Vagrant image to 22.04

JAORMX:
#### What type of PR is this?

/kind bug
/kind cleanup

22.04 is the new LTS and is what we should be using instead. 21.10
(which we were using) is already EOL.

None

#### Does this PR have test?

ubuntu e2e

```release-note
NONE
```


saschagrunert (approved):

saschagrunert:
@JAORMX lol we mered that into the `update-ubuntu` branch rather than `main`.

PR #1054, opened 2022-07-27
Switch to macos-12 runners

saschagrunert:
#### What type of PR is this?

/kind cleanup

The macos-10.xx runners will be removed soon so we should adapt to the new versions which seems to support VirtualBox as well.

None
#### Does this PR have test?
None

None

```release-note
None
```


JAORMX (approved):
/lgtm
/approve

Thanks @saschagrunert !

PR #1055, opened 2022-07-27
deploy: make sure that manifests start with a separator

jhrozek:
#### What type of PR is this?
/kind bug

Our operator-sdk bundle build command cat(1)s several files together,
but since the CSV, which is the last in the list, was missing a
separator on the first line, the last YAML object in the previous file
(deployment.yaml) was concatenated with the CSV.

Let's make sure the YAML documents in deploy/base start with a
separator.

None

#### Does this PR have test?
N/A will be tested by the existing ones

N/A

```release-note
NONE
```

### DIFF
--- a/deploy/base/clusterserviceversion.yaml
+++ b/deploy/base/clusterserviceversion.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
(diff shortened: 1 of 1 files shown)


JAORMX (approved):
good catch!
/lgtm
/approve

saschagrunert (approved):

PR #1058, opened 2022-07-28
Fix Flatcar Linux e2e tests

### LINKED ISSUE #1049: Flatcar e2e tests are broken
#### What happened:

The Flatcar VM fails to boot:

```
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: core
    default: SSH auth method: private key
Timed out while waiting for the machine to boot. This means that
Vagrant was unable to communicate with the guest machine within
the configured ("config.vm.boot_timeout" value) time period.

If you look above, you should be able to see the error(s) that
Vagrant had when attempting to connect to the machine. These errors
are usually good hints as to what may be wrong.

If you're using a custom box, make sure that networking is properly
working and you're able to connect to the machine. It is a common
problem that networking isn't setup properly in these boxes.
Verify that authentication configurations are also setup properly,
as well.

If the box appears to be booting properly, you may want to increase
the timeout ("config.vm.boot_timeout") value.
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Checking if box 'flatcar-stable' version '3227.2.0' is up to date...
==> default: Running provisioner: install-dependencies (shell)...
An error occurred in the underlying SSH library that Vagrant uses.
The error message is shown below. In many cases, errors from this
library are caused by ssh-agent issues. Try disabling your SSH
agent or removing some keys and try again.

If the problem persists, please report a bug to the net-ssh project.

timeout during server version negotiating
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Checking if box 'flatcar-stable' version '3227.2.0' is up to date...
==> default: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> default: flag to force provisioning. Provisioners marked to run always will still run.
```

The machine should boot and we should be able to run e2e tests

Run the flatcar e2e test in CI

#### Environment:

- Kernel (e.g. `uname -a`):
- Others:

ccojocar:
- test: use LTS image instea of stable for Flatcar e2e
- test: enable the Flatcar Linux e2e

#### What type of PR is this?

/kind failing-test

It switches the Flatcar Linux vagrant box to LTS image. It seems that the sable image is somehow broken and doesn't boot.

Fixes #1049

#### Does this PR have test?

yes

```release-note
NONE

```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -131,7 +131,7 @@ jobs:
 
   e2e-flatcar:
    needs: image
-#
+
    runs-on: macos-12
    timeout-minutes: 90
    env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,39 +129,39 @@ jobs:
       - name: Run E2E tests
         run: $RUN hack/ci/e2e-ubuntu.sh
 
-  #e2e-flatcar:
-  #  needs: image
-
-  #  runs-on: macos-12
-  #  timeout-minutes: 90
-  #  env:
-  #    RUN: ./hack/ci/run-flatcar.sh
-  #  steps:
-  #    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b
-  #    - name: Vagrant box version
-  #      id: vagrant-box
-  #      run: |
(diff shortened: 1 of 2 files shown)


ccojocar:
The metrics e2e test seems to be flaky.

JAORMX (changes requested):

JAORMX (approved):
/lgtm
/approve

PR #1067, opened 2022-08-03
helm: Package and save chart in a GitHub release.

stephen-fox:
#### What type of PR is this?
/kind cleanup

This adds a GitHub Actions workflow that runs "helm package"
on the helm chart and saves it as a workflow artifact. If the
workflow was triggered by a GitHub release being published,
it automatically adds the package to the release that triggered
the workflow.

Note: The package is only attached to a release when it is
published. A draft release will not trigger the workflow.
Refer to the GitHub documentation for more information:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

The reason for running "helm package" on both pull requests
and release-publishings is to ensure a change introduced in
a PR does not break the workflow automation (it would be very
unfortunate to discover that the workflow broke when we create
a GitHub release).

None

#### Does this PR have test?
The workflow tests that "helm package" creates a file and that
the file exists. The GitHub "upload-artifact" action is configured
to fail if it cannot find a file to upload (sadly, like GitLab, this is not
the default behavior).

I spoke with @saschagrunert and @JAORMX about the automation
working this way (that is: packaging the chart and attaching it to
an existing release).

```release-note
NONE
```

### DIFF
--- a/.github/workflows/helm-chart-package.yaml
+++ b/.github/workflows/helm-chart-package.yaml
@@ -1,0 +1,53 @@
+name: helm-chart-package
+on:
+  pull_request:
+  release:
+    types: [published]
+permissions:
+  actions: none
+  checks: none
+  contents: write
+  deployments: none
+  id-token: none
+  issues: none
+  discussions: none
+  packages: none
+  pull-requests: none
+  repository-projects: none
+  security-events: none
+  statuses: none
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
Awesome, thank you!
/lgtm

PR #1059, opened 2022-07-28
Run metrics e2e tests in single-node environments only

JAORMX:
#### What type of PR is this?

/kind flake

#### What this PR does / why we need it:

This is meant to remove some of the flakiness in our tests. We make the
assumption that our vanilla tests are running in single-node mode only,
so we take advantage of that fact and set a boolean that indicates that
it's a single-node test environment.

Note that this also does changes necessary for the ubuntu setup to
run the Seccomp tests.

#### Which issue(s) this PR fixes:

None

#### Does this PR have test?

Yes, the metrics e2e tests

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #1069, opened 2022-08-04
Run metrics e2e tests in single-node environments only (cherry-pick)

jhrozek:
#### What type of PR is this?
/kind flake

This PR cherry-picks one patch from PR #1059 and adds one more patch in
a similar manner, also running a metrics check in a single-node environment
only.

The second patch prevents frequent flakes in an OCP six-node environment.

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```


saschagrunert (approved):
Thank you for picking this up!

PR #1070, opened 2022-08-04
Two minor test deflakes

jhrozek:
#### What type of PR is this?
/kind flake

- tests: Clean up selinuxprofiles when tests end, too - avoids the test hanging in case a selinux profile is not cleaned up correctly for one reason or another
- tests: Wait for profile name to be reconciled before creating pod - avoids a race in case a profile takes a little while to reconcile and the pod would have been created too fast

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```


jhrozek:
funny that a deflake PR would flake out, but it's in the metrics test which is addressed by another deflake PR.

/test pull-security-profiles-operator-test-e2e

saschagrunert (approved):

PR #1075, opened 2022-08-08
chore: set explicit timeouts on webhooks

ccojocar:
#### What type of PR is this?

/kind bug

It sets explicit timeouts shorter than the default. This is required by our webhook configuration static check which doesn't allowed longer timeouts than 5 seconds.

```release-note
NONE

```

### DIFF
--- a/deploy/overlays/webhook/webhook_config.yaml
+++ b/deploy/overlays/webhook/webhook_config.yaml
@@ -9,6 +9,7 @@ metadata:
 webhooks:
   - name: binding.spo.io
     failurePolicy: Fail
+    timeoutSeconds: 5
     sideEffects: None
     rules:
       - operations: ["CREATE", "UPDATE", "DELETE"]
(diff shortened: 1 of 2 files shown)


deploy/overlays/webhook/webhook_config.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
 webhooks:
   - name: binding.spo.io
     failurePolicy: Fail
+    timeoutSeconds: 5
```
jhrozek, line 12: If this is realistically something that deployers want configurable, why not make it configurable? (What if your policy will say no longer than 4 seconds next year)?

ccojocar: Happy to make it configurable. How do you suggest to do that in this situation?

jhrozek: ugh, I'm sorry, only now did I notice that this is the overlay, for some reason I thought you were modifying the go sources in bindata with a hardcoded config. Sorry, going to ack this now.

jhrozek (approved):
/lgtm

deploy/overlays/webhook/webhook_config.yaml
ccojocar: Thanks for reviewing it!

jhrozek (approved):
/lgtm

PR #1076, opened 2022-08-08
helm: Added chart installation instructions.

stephen-fox:
#### What type of PR is this?
/kind documentation

This PR adds installation documentation for the recently-added helm chart.

None

#### Does this PR have test?
N/A

```release-note
NONE
```

### DIFF
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -130,7 +130,7 @@ kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/
 kubectl --namespace cert-manager wait --for condition=ready pod -l app.kubernetes.io/instance=cert-manager
 # Install the chart from a release URL (note: helm also allows users to
 # specify a file path instead of a URL, if desired):
-helm install security-profiles-operator https://github.com/kubernetes-sigs/security-profiles-operator/releases/download/v0.4.3/security-profiles-operator-0.4.3.tgz
+helm install security-profiles-operator https://github.com/kubernetes-sigs/security-profiles-operator/releases/download/v0.4.4-dev/security-profiles-operator-0.4.4-dev.tgz
 ```
 
 ## Set logging verbosity
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -7,6 +7,7 @@
 - [Install operator](#install-operator)
   - [Installation using OLM from operatorhub.io](#installation-using-olm-from-operatorhubio)
   - [Installation using OLM using upstream catalog and bundle](#installation-using-olm-using-upstream-catalog-and-bundle)
+  - [Installation using helm](#installation-using-helm)
 - [Set logging verbosity](#set-logging-verbosity)
 - [Configure the SELinux type](#configure-the-selinux-type)
 - [Restrict the allowed syscalls in seccomp profiles](#restrict-the-allowed-syscalls-in-seccomp-profiles)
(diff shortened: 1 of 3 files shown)


installation-usage.md
```diff
@@ -116,6 +117,22 @@ manifest=https://raw.githubusercontent.com/kubernetes-sigs/security-profiles-ope
...
+as an artifact, and can be installed by executing the following shell commands:
+
+```shell
+# Install cert-manager if it is not already installed (TODO: The helm
+# chart might do this one day - see issue 1062 for details):
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.8.0/cert-manager.yaml
+kubectl --namespace cert-manager wait --for condition=ready pod -l app.kubernetes.io/instance=cert-manager
+# Install the chart from a release URL (note: helm also allows users to
+# specify a file path instead of a URL, if desired):
+helm install security-profiles-operator https://github.com/kubernetes-sigs/security-profiles-operator/releases/download/v0.5.0/security-profiles-operator-0.1.0.tgz
```
jhrozek, line 133: so what are the two NVRs? I guess the first one is SPO version?
What I'm wondering is whether it makes sense to add any new NVRs we reference to dependencies.yaml and have them checked so that they don't get out of sync

stephen-fox: The first version (v0.5.0) is an imaginary release of the security-profiles-operator (since the current release does not have a helm chart attached). The second version (0.1.0) is the helm chart version. With regards to helm - there are unfortunately two versions: the application version and the chart version. There's no reason why they could not be the same, especially if we're tying each chart release to a SPO release.

I'm unfamiliar with `zeitgeist`, but it appears something like the following would work:

In `dependencies.yaml`:

```yaml
dependencies:
  - name: spo-version
    version: v0.4.1
    refPaths:
    - path: installation-usage.md
      match: SPO_VERSION

  - name: spo-helm-chart-version
    version: 0.1.0
    refPaths:
    - path: installation-usage.md
      match: SPO_HELM_CHART_VERSION
```

And then in `installation-usage.md`:

```sh
# ...
SPO_VERSION=v0.4.1
SPO_HELM_CHART_VERSION=0.1.0
helm install security-profiles-operator https://github.com/kubernetes-sigs/security-profiles-operator/releases/download/${SPO_VERSION}/security-profiles-operator-${SPO_HELM_CHART_VERSION}.tgz
```

What do you think, Jakub?

jhrozek: So I'm somewhat surprised that we don't check the SPO version against what we have in `bundle/manifests/security-profiles-operator.clusterserviceversion.yaml` and `VERSION` yet, but I was proposing using zeitgeist mostly because I knew that we have the version information somewhere already and didn't want the versions to drift.

I'm not super familiar with zeitgeist yet either, but I don't think we would have to split the variables from the `helm install` line either unless you find it more readable, we could even keep the version there as is:
```
 helm install security-profiles-operator https://github.com/kubernetes-sigs/security-profiles-operator/releases/download/v0.5.0/security-profiles-operator-0.1.0.tgz
```
and then match with:
```yaml
- name: spo-version
    version: v0.4.1
    refPaths:
    - path: installation-usage.md
      match: helm install security-profiles-operator
```
(I think/hope :-))

I don't have a strong opinion on the helm chart version; I've only ever used helm as a user, so my only answer is "do whatever is normally done with other projects".

jhrozek: btw since the version numbers are bumped during releases and it's @saschagrunert who does the releases maybe we could also wait for his opinion so I don't screw up his process :-)

stephen-fox: Let me try that out in the mean time.

stephen-fox: That does appear to work. I'd be happy to add other things to the `dependencies.yaml` file as well.

JAORMX: I'd rather have the Helm chart version match the operator version. That'll make things easier to track. I do agree with @jhrozek about tracking versions with zeitgeist, that has proven very helpful.

stephen-fox: I'll have a commit for this pushed in a bit. Unfortunately, it appears zeitgeist checks the entire line for the specified text. In the case of `installation-usage.md`, there are two version numbers in the `helm install ...` arguments. In other words - it is possible that one version might be up-to-date while the second gets left behind by accident.

We could do the shell variable thing I posted earlier (e.g., `SPO_VERSION=0.4.3`) and then reference that in the command. Or, we could modify the GitHub Actions Workflow to remove the version number from the chart package file. Regarding the latter - I'm not sure what the typical pattern is for helm chart file names. The `helm package` subcommand always generates a file with the version number included in it.

There is also the matter of whether or not helm chart versions should have a `v` prefix (e.g., v0.0.1 versus 0.0.1). The `helm create ...` subcommand generates a chart whose version numbers do not include the leading `v`. The documentation does not seem to have a strict opinion about this. While we do not have to make a decision about the leading "v" in the current PR... it adds more annoying complexity.

stephen-fox:
😭

JAORMX:
`make bundle`

JAORMX:
@stephen-fox I recently did some changes to the bundle, so you probably need to run the generation again... we should document all this

JAORMX (approved):
/lgtm
/approve

Great job! Thanks for adding this

PR #1083, opened 2022-08-10
nodestatus: Only log low-level messages if verbosity is increased

### LINKED ISSUE #1066: the nodestatus module is too verbose
#### What happened:
There's a lot of messages in the nodestatus.go module that use the Info level while using the increased verbosity configurable in spod would probably be a better idea.

no messages about setting the nodestatus to something without setting extra verbosity

just install a seccomp profile

This is split off issue #990 

#### Environment:

- Kernel (e.g. `uname -a`):
- Others:

jhrozek:
#### What type of PR is this?
/kind cleanup

Unless the nodestatus controller is being debugged, then these messages
are really not that important.

Fixes #1066

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```


JAORMX (approved):
/lgtm
/approve

PR #1097, opened 2022-08-19
deploy: Set the NS for spod SA explicitly to allow privileged pods under OCP 4.12+

jhrozek:
#### What type of PR is this?
/kind bug

OCP 4.12+ is trying to make PSA and SCCs coexist. In order to find out
whether the workload can avoid being restricted by the default
("restricted") PSA profile, there is a label synchronizer that takes a
look at the available RBAC permissions and if there is an SA that is
allowed to use a privilged SCC, then the namespace is allowed to be
labeled with `pod-security.kubernetes.io/enforce: privileged`.
More details can be found at:
    https://github.com/openshift/enhancements/blob/master/enhancements/authentication/pod-security-admission-autolabeling.md

It seems that this mechanism relies on having the namespace set
explicitly for RoleBinding's `.subjects[].namespace` to be set when
`.subjects[].kind` is ServiceAccount even though this is not required by
the RBAC system to successfully bind the SA to a Role

This might be a bug:
    https://issues.redhat.com/browse/OCPBUGS-160
however, setting the namespace explicitly does no harm and lets us run
the pods in a privileged manner on OCP 4.12+

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```

### DIFF
--- a/bundle/manifests/spod_rbac.authorization.k8s.io_v1_rolebinding.yaml
+++ b/bundle/manifests/spod_rbac.authorization.k8s.io_v1_rolebinding.yaml
@@ -12,3 +12,4 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: spod
+  namespace: security-profiles-operator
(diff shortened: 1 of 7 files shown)


JAORMX:
@jhrozek seems you missed `make bundle`

jhrozek:
> @jhrozek seems you missed `make bundle`

I did. Fixed!

PR #1105, opened 2022-08-22
Fix tag usage and image loading in image builds

### LINKED ISSUE #1104: Image build CI broken
#### What happened:

The CI for the jobs `image` and `ubi-image` are broken right now:
- https://github.com/kubernetes-sigs/security-profiles-operator/runs/7946447237?check_suite_focus=true
- https://github.com/kubernetes-sigs/security-profiles-operator/runs/7946447601?check_suite_focus=true

JAORMX:
#### What type of PR is this?

/kind bug
/kind regression

We cannot both push and load at the same time, so logic was added to
make that work. We also now only do container scanning on the first tag
since we cannot use multiple tags as the output of the metadata-action
would normally provide.

Fixes #1104

#### Does this PR have test?

The image build tests

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -96,7 +96,7 @@ jobs:
       - name: Run container image vulnerability scanner
         uses: aquasecurity/trivy-action@d63413b0a4a4482237085319f7f4a1ce99a8f2ac
         with:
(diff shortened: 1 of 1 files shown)


pjbgf (approved):
/lgtm

saschagrunert (approved):

PR #1110, opened 2022-08-23
Process bpf events in parallel

saschagrunert:
#### What type of PR is this?

/kind feature

We now spawn a go routine for each event to be able to process them in parallel.

None

#### Does this PR have test?

None

None

```release-note
Updated BPF event processing to run in parallel.
```

### DIFF
--- a/internal/pkg/daemon/bpfrecorder/bpfrecorder.go
+++ b/internal/pkg/daemon/bpfrecorder/bpfrecorder.go
@@ -39,6 +39,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/jellydator/ttlcache/v3"
 	seccomp "github.com/seccomp/libseccomp-golang"
+	"golang.org/x/sync/semaphore"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


internal/pkg/daemon/bpfrecorder/bpfrecorder.go
```diff
@@ -546,16 +546,17 @@ func toStringByte(array []byte) string {
 func (b *BpfRecorder) processEvents(events chan []byte) {
 	b.logger.Info("Processing events")
 	for event := range events {
-		func() {
+		eventCopy := event
+		go func() {
```
JAORMX, line 550: While this is a good enhancement, I'm slightly concerned that we'd hit a goroutine limit in bigger deployments. Pheraps after this merges it would be good to do some stress testing of this part of the code-base.

saschagrunert: In theory we can get 4194304 go routines for 64bit systems (max amount of pids), which could cause memory spikes. So yeah either we limit it by using a worker pool or we adjust the memory constraints accordingly.

JAORMX:
/approve

@pjbgf @jhrozek could you also take a look?

JAORMX (approved):

JAORMX:
/lgtm cancel

waiting for feedback from other maintainers

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
```diff
@@ -545,68 +546,83 @@ func toStringByte(array []byte) string {
 
 func (b *BpfRecorder) processEvents(events chan []byte) {
 	b.logger.Info("Processing events")
+
+	// Allow up to 1000 goroutines to run in parallel
+	const workers = 1000
```
pjbgf, line 551: Feels like something that could/should be configurable.

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
jhrozek: Would it help, as an additional enhancement, to have a pool of gouroutines? I'm sure we're not the first ones with this requirement, so I bet there exists a library that already does this..

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
saschagrunert: Hm, how would users decide which value to choose here? I'm not sure about making it configurable for end users without confusing them.

brness: maybe this could be positively related with spod request and limits.

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
JAORMX: I think @saschagrunert 's solution to use a semaphore is equivalent to having a pool. I don't have an opinion in terms of making that configurable as @pjbgf suggested, but it might be worth doing.

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
pjbgf: @saschagrunert the worry was mostly around giving users a way to control this in case it got out of hand.
But I am happy to ahead as-is and review before next release.

Issue #744, opened 2021-12-02
The recording webhook's resource updating is racy


jhrozek:
#### What happened:
I deleted pods in a deployment to trigger recording a policy and saw:
```
Error from server: admission webhook "recording.spo.io" denied the request: failed to update resource profilerecording: Operation cannot be fulfilled on profilerecordings.security-profiles-operator.x-k8s.io "nginx-selinux-policy": the object has been modified; please apply your changes to the latest version and try again
Error from server: admission webhook "recording.spo.io" denied the request: update resource on removing pod: failed to update resource profilerecording status: Operation cannot be fulfilled on profilerecordings.security-profiles-operator.x-k8s.io "nginx-selinux-policy": the object has been modified; please apply your changes to the latest version and try again
Error from server: admission webhook "recording.spo.io" denied the request: update resource on removing pod: failed to update resource profilerecording status: Operation cannot be fulfilled on profilerecordings.security-profiles-operator.x-k8s.io "nginx-selinux-policy": the object has been modified; please apply your changes to the latest version and try again
```

#### What you expected to happen:
No errors

#### How to reproduce it (as minimally and precisely as possible):
 - create a profilerecording
 - create a largish deployment (I managed to reproduce with about ~50 pods) matching the profilerecording
 - kubectl delete the pods in the deployment
 
#### Anything else we need to know?:
N/A

#### Environment:

- Cloud provider or hardware configuration: AWS
- OS (e.g: `cat /etc/os-release`): RHCOS 4.9
- Kernel (e.g. `uname -a`): N/A
- Others: N/A

jhrozek:
/remove-lifecycle stale
this has a PR on review

PR #1112, opened 2022-08-23
recording webhook: Run only one replica in the recording deployment, make container replica tracking more predictable

jhrozek:
#### What type of PR is this?
/kind bug

This PR contains a self-contained change from the branch where I work on
merging recorded policies. Since the changes are useful (IMO) on their own
and the branch has been getting quite big, I decided to submit them separately.

More information about the individual commits are below and also in the
commit messages.

- webhook: Only run one replica since the webhook is stateful
Since the webhook tracks the replicas seen and the associated replica
number, it cannot run with multiple replicas. If if did, it's possible
that each pod under admission hits a different replica, which means that
the recorded policies use the same replica number and the recorded
policies are overwritten. This is especially true if the hook recorder
is used where the replica number is used as the output file.

Longer term, we should remove the container replica tracking from the
webhook so that they are again truly stateless, but in the meantime,
let's only use one webhook instance.

Related:
    https://bugzilla.redhat.com/show_bug.cgi?id=2089329

- e2e: use readinessProbe to make sure we record containers when they're actually ready
If no readiness probe is used, then we might end up recording the
container before it is truly ready and the policy might not contain
all the calls we need.

- webhooks: Fix setting finalizers
Only one call to set the status was used to set both finalizers and the
status, which doesn't work. Let's use a separate call for setting the
status and a separate call for setting the finalizer.

- webhooks: Make setting finalizers and status more resilient with retries
Especially with multiple replicas, we've seen the webhooks error out due
to a conflict. This is problematic because the webhooks have a hard-fail
policy.

Let's retry multiple times instead.

- e2e: test that recording finalizer prevents deleting the resource
Adds a test for the finalizer.

- recording webhook: don't use the replica name to track pods being recorded, use labels instead
The recording webhook used to track container replicas in a sync.Map,
but its usage was problematic because:
 - all replicas were deleted in case any of replicas with the same
   generatedName were used. This meant replica numbers used for
   recorded policies were being reused and not all containers would
   end up having a recorded policy
 - if a replica was removed, the profilerecording would lose the
   status and finalizers
 - using the replica tracking in the sync.Map means that the webhook
   is maintaining its state, which it shouldn't. While this patch
   doesn't remove the state completely it reduces its usage.

Instead of the above, this patch maintains the finalizer and the status
as a combination of examining the pod being admitted as well as the
currently existing pods which are listed during admission.

The replica number is then always increasing by one e.g. across scaleups
or scaledowns of replicated pods.

Finally, because the webhook is watching all namespaces, but the
sync.Map was only storing plain names, let's also add namespace as a
prefix for the syncMap. This allows to empty the sync.Map once no pods
for a single recording exist, keeping the memory usage low over time.

Fixes: #744

#### Does this PR have test?
yes

I'll point them out inline

```release-note
The webhooks that handle profile recording and binding now run only one replica
to avoid races due to the recording webhook keeping an internal state.
```


internal/pkg/manager/spod/bindata/webhook.go
```diff
@@ -37,7 +37,7 @@ import (
 )
 
 var (
-	replicas                int32 = 3
+	replicas                int32 = 1
```
jhrozek, line 40: This brings a related question -- since we only run a single replica, but the webhook listens to all namespaces, wouldn't it become a bottleneck?

I think the most systematic solution would be to actually remove the replica tracking from the webhook and make it truly stateless. The replica tracking is used for two things (AFAICS):
 1. when recording with the hook, to use in the filename recorded into
 2. to discern between different container replicas

For both, I guess we could use a random number instead, but I was thinking if it was acceptable to deprecate support for the hook based recording and make merging of policies the default once we have it? Frankly, I don't see much reason in having separate policies for instances of individual containers, wouldn't you want to have a single policy instead? The hook based recording, if we didn't want to deprecate it, could even use random numbers instead.

Alternatively, we could have a single gRPC endpoint just to track the replicas, but that brings even more complexity.

test/tc_profilerecordings_test.go
```diff
@@ -352,6 +352,13 @@ spec:
       containers:
       - name: nginx
         image: quay.io/security-profiles-operator/test-nginx-unprivileged:1.21
+        ports:
+        - containerPort: 8080
+        readinessProbe:
```
jhrozek, line 425: This might be something to recommend in our docs

internal/pkg/manager/spod/bindata/webhook.go
JAORMX: I'd be up for removing support for the hook. It adds extra complexity and we do have other alternatives to offer. We really shouldn't have anything stateful in the webhook... ideally, if we need to keep track of something, we should have a separate system to track the stateful bits... e.g. we might want to start leveraging Redis and probing it from the webhook. Having a performant distributed cache we could query would help.

jhrozek: Yes, I was considering just moving the sync.Map to the operator (controller) and have it fronted by a gRPC API to get the next replica number, but the simpler we can make the webhook the better

JAORMX: That's an option, but I'd just highly suggest using a dedicated service that's meant for caching, e.g. Redis. It's less code we'd have to maintain.

saschagrunert: +1 for removing the hook

jhrozek: Great. I'll work on removing the hook and then I'll see if we can adjust the code to get away with the using e.g. random numbers for the individually recorded container replicas. Afterwards, we'll be able to go back to multiple webhook pods.

Please let me know if I should make that work part of this PR or a follow-up.

saschagrunert: I think it can be part of this PR :)

jhrozek: done

jhrozek:
oops unit tests

jhrozek:
welp:
```
cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
```

jhrozek:
> welp:
> 
> ```
> cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
> ```

ah! Previously we were installing `oci-seccomp-bpf-hook` which was dragging in `gcc` through the `kernel-devel` dependency. So at least it is not a regressions.

jhrozek:
rebased to pick up the recent fixes for the flaky tests -- most of the recording tests are marked as flaky and I want to make sure they pass as well (they did locally, but I was typically running a subset only)

saschagrunert (approved):
Awesome work, thank you so much!

jhrozek:
ubuntu flaky e2e tests passed:
https://github.com/kubernetes-sigs/security-profiles-operator/runs/8223349760?check_suite_focus=true#step:9:6819
/hold until the others pass as well

jhrozek:
hmm, in flatcar tests I still see:
```
    e2e_flaky_test.go:23: Skipping flaky tests
```
going to look into it. But I don't think that flatcar tests anything that fedora and ubuntu wouldn't test..

jhrozek:
fedora tests passed:
https://github.com/kubernetes-sigs/security-profiles-operator/runs/8223349942?check_suite_focus=true#step:9:18556

PR #1125, opened 2022-08-26
workflows: Set the E2E_TEST_FLAKY_TESTS_ONLY=true variable inside the test VM

jhrozek:
#### What type of PR is this?
/kind bug
/kind failing-test

Turns out that the flaky e2e tests were not run at all because we were
setting the `E2E_TEST_FLAKY_TESTS_ONLY=true` variable for the $RUN command
that just runs "ssh sudo -c" which clears the environment.

Let's set the variable inside the test VM instead so that the flaky
tests run properly.

None

#### Does this PR have test?
Implicitly

N/A

```release-note
NONE
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -97,9 +97,9 @@ jobs:
         run: $RUN hack/ci/e2e-fedora.sh
       - name: Run Flaky E2E tests
         continue-on-error: true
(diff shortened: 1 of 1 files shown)


JAORMX:
oops!

Good catch!

/lgtm

PR #1151, opened 2022-09-06
Fix olm CI tests

saschagrunert:
#### What type of PR is this?

/kind failing-test

An approach to fix the failing OLM tests.

None
#### Does this PR have test?
None

None

```release-note
None
```

### DIFF
--- a/examples/olm/install-resources.yaml
+++ b/examples/olm/install-resources.yaml
@@ -2,7 +2,8 @@
 apiVersion: v1
 kind: Namespace
 metadata:
-    name: security-profiles-operator
+  name: security-profiles-operator
+  labels:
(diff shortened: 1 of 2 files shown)


examples/olm/install-resources.yaml
```diff
@@ -3,7 +3,6 @@ apiVersion: v1
 kind: Namespace
 metadata:
     name: security-profiles-operator
-    openshift.io/cluster-monitoring: "true"
```
jhrozek, line 6: Why not put the cluster-monitoring into labels? But if you just want to fix the tests in the fastest way possible, then this is fine.

jhrozek (approved):

PR #1153, opened 2022-09-06
tests: Run the flaky tests, too

jhrozek:
#### What type of PR is this?
/kind bug

Turns out that my previous patch to run the flaky tests was incomplete.
We need some more changes:
    - the `testify.m` parameter must be, for some reason, specified last
      or else the `go test` call fails with "no Go files in XYZ".
    - after the namespaced tests (last in the regular non-flaky tests),
      we need to reset the namespace or else we test everything in the
      test-ns namespace
    - we need to tear down the operator after the namespaced tests to
      be able to roll it out again in the flaky tests

The tests are still flaky, but at least they run now.

None

#### Does this PR have test?
implicit

N/A

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,7 +396,7 @@ test-e2e: ## Run the end-to-end tests
 test-flaky-e2e: ## Only run the flaky end-to-end tests
 	CGO_LDFLAGS= \
 	E2E_SKIP_FLAKY_TESTS=false \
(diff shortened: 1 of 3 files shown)


pjbgf (approved):
/lgtm

Thanks for working on this @jhrozek! :bow:

PR #1156, opened 2022-09-08
recording: Emit event when attempting to record a privileged container + add docs

jhrozek:
#### What type of PR is this?
/kind cleanup
/kind documentation

- docs: Document that the log based recorder can't record privileged pods - adds docs that make it clear that we can't record privileged containers using the log enricher
- recording: Issue an event if a privileged pod is recorded with a log-based recorder - when log enricher is used to record privileged containers, attach an event to the recording object

None

#### Does this PR have test?
No

N/A

```release-note
If the log-based recorder is in use and the user attempts to either record a container which already had its SecurityContext set or attempts to record a privileged container (which ignores both seccomp profiles and selinux contexts), the profile recording webhook issues a warning event.
```

### DIFF
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -453,6 +453,12 @@ test-recording-redis   Installed   15s
 Recording a SELinux profile would work the same, except you'd use `kind: SelinuxProfile`
 in the `ProfileRecording` object.
 
+Please note that log based recording does not have any effect if the recorded container
+is privileged, that is, the container's security context sets `privileged: true`. This
+is because privileged containers are not subject to SELinux or seccomp policies at all
+and the log based recording makes use of a special seccomp or SELinux profile respectively
+to record the syscalls or SELinux events.
+
 #### eBPF based recording
 
 The operator also supports an [eBPF](https://ebpf.io) based recorder. This
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -514,7 +514,7 @@ func runWebhook(ctx *cli.Context, info *version.Info) error {
 	setupLog.Info("registering webhooks")
 	hookserver := mgr.GetWebhookServer()
 	binding.RegisterWebhook(hookserver, mgr.GetClient())
-	recording.RegisterWebhook(hookserver, mgr.GetClient())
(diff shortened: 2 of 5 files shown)


JAORMX (approved):

JAORMX:
/lgtm cancel
@jhrozek This is actually a user facing change, so this should have a relevant release note.

installation-usage.md
```diff
@@ -453,6 +453,12 @@ test-recording-redis   Installed   15s
 Recording a SELinux profile would work the same, except you'd use `kind: SelinuxProfile`
 in the `ProfileRecording` object.
 
+Please note that log based recording does not have any effect if the recorded container
+is privileged, that is, the container's security context sets `privileged: true`. This
```
saschagrunert, line 457: Do we wanna behave in the same way if we already have a seccomp profile applied to the container?

JAORMX: I think there might be a use-case for still recording a workload even if there's a profile applied to it. e.g. verifying that the applied profile is sufficient.

saschagrunert: But for the log enricher we would override the existing profile.

JAORMX: For Seccomp? I thought it was only the case for SELinux.

jhrozek: > Do we wanna behave in the same way if we already have a seccomp profile applied to the container?

saschagrunert: The webhook just updates the security context for log based seccomp profile recordings: https://github.com/kubernetes-sigs/security-profiles-operator/blob/1e4dfea95c075717e0fa9189380ce72f06da6a3b/internal/pkg/webhooks/recording/recording.go#L250-L268

saschagrunert: > > Do we wanna behave in the same way if we already have a seccomp profile applied to the container?
> 
> I'm sorry, can you elaborate? Are you thinking about the case where someone records a workload while already specifying the profile in the first place?

Yes, exactly. We could assume that users do not understand how it works in detail, so I'm wondering if we wanna make them aware that we overwrote their already used seccomp profile.

jhrozek: OK, makes sense. btw it seems like we were overwriting the context for SELinux recording as well, so I added events to both.

saschagrunert (approved):

PR #1178, opened 2022-09-20
BUG 2082099: Fix finalizer string too long

Vincent056:
/kind bug

This PR shorten the finalizer string if it is over the limit size of 63

#### Does this PR have test?

Yes

No

```release-note
Fix the finalizer string too long, shorten the length of the node name if the finalizer string combined length is over the size of 63
```

####Which issue(s) this PR fixes:

Fixes: [BZ 2082099](https://bugzilla.redhat.com/show_bug.cgi?id=2082099)

### DIFF
--- a/internal/pkg/nodestatus/nodestatus.go
+++ b/internal/pkg/nodestatus/nodestatus.go
@@ -33,6 +33,11 @@ import (
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/util"
 )
 
+const (
+	// finalizer name length limit.
+	finalizerNameLenLimit = 63
+)
+
 type StatusClient struct {
 	pol             client.Object
 	nodeName        string
@@ -42,13 +47,18 @@ type StatusClient struct {
 
 func NewForProfile(pol client.Object, c client.Client) (*StatusClient, error) {
 	nodeName, ok := os.LookupEnv(config.NodeNameEnvKey)
+	finalizerString := nodeName + "-deleted"
 	if !ok {
 		return nil, errors.New("cannot determine node name")
+	}
+	// Make sure the length of finalizer is not longer than 63 characters
+	if len(nodeName)+len("-deleted") > finalizerNameLenLimit {
+		finalizerString = nodeName[:finalizerNameLenLimit-len("-deleted")] + "-deleted"
 	}
 	return &StatusClient{
 		pol:             pol,
 		nodeName:        nodeName,
-		finalizerString: nodeName + "-delete",
+		finalizerString: finalizerString,
 		client:          c,
 	}, nil
 }
--- a/internal/pkg/nodestatus/nodestatus_test.go
+++ b/internal/pkg/nodestatus/nodestatus_test.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


saschagrunert:
/ok-to-test

Thank you for the fix @Vincent056!

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -42,13 +42,18 @@ type StatusClient struct {
 
 func NewForProfile(pol client.Object, c client.Client) (*StatusClient, error) {
 	nodeName, ok := os.LookupEnv(config.NodeNameEnvKey)
+	finalizerString := nodeName + "-deleted"
 	if !ok {
 		return nil, errors.New("cannot determine node name")
 	}
+	// Make sure the length of finalizer is not longer than 63 characters
+	if len(nodeName) > 55 {
```
saschagrunert, line 50: We can make it a `const` to fix the CI.

jhrozek:
An e2e test would also be nice.

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -33,6 +33,11 @@ import (
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/util"
 )
 
+const (
+	// node name length limit
+	nodeNameLenLimit = 55
```
jhrozek, line 38: This is a complete nit, but can you change a comment to include that the finalizer can be up to 63 chars in lenght and 55 + len("-deleted") = 63? We might also want to use a constant instead of `-deleted` so that both are next to each other.

internal/pkg/nodestatus/nodestatus_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
+)
+
+// Expected shorten the node name if length exceed the limit
```
jhrozek, line 27: yup, validation error

jhrozek (commented):
just some nits

jhrozek:
```
internal/pkg/nodestatus/nodestatus_test.go:23: File is not `gci`-ed with --skip-generated -s standard,default,prefix(sigs.k8s.io/security-profiles-operator) (gci)
	"github.com/stretchr/testify/require"
```

So far I've had the best results with fixing the gci issues by running gci manually, something like:
```
go install github.com/daixiang0/gci@latest
gci write path/to/file
```
There is also the `-l` option that allows you to specify "local" imports that should be sorted last, but I could never get the flag combination to work, so I sort them manually. Normally the SPO expects the imports to be sorted:

- stdlib, alphabetically
- third party, alphabetically
- local, alphabetically

```
internal/pkg/nodestatus/nodestatus_test.go:60: File is not `gofumpt`-ed (gofumpt)
```

`/build/golangci-lint run --fix --build-tags no_bpf` should do the trick

Don't forget to remove and reinstall golangci-lint if you have an old version locally.

Vincent056:
@jhrozek It seems like `t.Parallel called after t.Setenv; cannot set environment variables in parallel tests [recovered]`, and not setting parallel is causing verify to fail

jhrozek:
> @jhrozek It seems like `t.Parallel called after t.Setenv; cannot set environment variables in parallel tests [recovered]`, and not setting parallel is causing verify to fail

Can you try adding `//nolint:paralleltest` with the explanation you used to suppress the warning? There are some other uses of this stanza in SPO, you can git grep for its usage to see exactly where and how it's used.

jhrozek (approved):
/lgtm
I restarted the flatcar test, the failure looks like a install-time fluke

PR #1179, opened 2022-09-20
Add support for merging profiles recorded by a single profileRecording

### LINKED ISSUE #677: recording of replicated pods should contain the union of the recorded policies
#### What happened:
Recording of a replicated pod creates a policy per replica. Depending on what does each replica do during the recording, each policy might be a bit different with no easy way to merge the policies other than manually.

Because the pods managed the replicating controller, be it a Deployment, a DeamonSet or a ReplicaSet should be interchangeable, I would expect the result to be an union of the recorded policies.

Record a deployment, check out that the result present a policy per pod.

N/A

#### Environment:

- Cloud provider or hardware configuration: N/A
- OS (e.g: `cat /etc/os-release`): N/A
- Kernel (e.g. `uname -a`): N/A
- Others: N/A

jhrozek:
#### What type of PR is this?
/kind feature
/kind api-change

- profilerecording API: Add an optional mergeStrategy field
The profileRecording CRD gets a new field mergeStrategy. This field defaults to "none" which is the same behavior as earlier. When set to "containers", all policies belonging to the same container and the same recordingn will be merged.

- api: add a profileSpecBase interface
Adds a new interface that all profiles capable of producing recordable profiles should implement. The interface consists of two methods:
    IsPartial() - if the policy is partial and should not be reconciled yet
    ListProfilesByRecording() - lists all profiles recorded by a particular recording
These will come handy when merging the profiles. Also adds several labels that these profiles will use to identify the recording and mark the profile as partial.

- profilerecording: Create partial profiles if recording sets mergeStrategy=containers
When the mergeStrategy field of profileRecording is set to containers, the profilerecorder controller in the deamon would create profiles with
the label profilebase.ProfilePartialLabel set to true. The profiles would furthermore contain a label linking to the profileRecording and
the name of the container which will later aid in reconstructing the merged profile.
When the partial profiles are created, a finalizer is set on the profileRecording which prevents its deletion unless the partial profiles are either deleted manually by the admin or merged.

RBAC changes - the profilerecording controller (running under the spod identity) is now allowed to list,watch and get profilerecordings

- nodestatus: Add a new profile status "Partial" for profiles yet to be merged
Adds a new profile status "Partial" to both the nodestatus CRD and by extension the profiles itself. This status is used to denote profiles that are not to be used right away, but those that are about to be merged.
Apart from managing the status itself, the node finalizer would also remove the finalizer preventing the recording from being deleted if all the recordings go away. This covers the use case where someone records partial policies and then just wants to remove them instead of merging.

- seccomp/selinux: do not handle partial seccomp and selinux profile
Skips reconciling partial selinux and seccomp profiles. SPO should only reconcile those when they are merged.

- api: Add a generic Go interface to merge partial profiles
In order to avoid repeating ourselves in the merge code, let's add a new interface MergeableProfile with a single method Merge() that is implemented by all profiles that can be recorded.
Moves UnionSyscalls from the seccomp controller to the seccomp types along with its unit tests.

- controller: Add a new recording merger controller
Adds a new controller called recordingmerger that watches for profilerecordings that are about to be deleted. For those recordings, all partial profiles are merged into a final profile and created.

RBAC changes: adds delete/deletecollection for both seccomp and selinux profiles to the SPO controllers in order to remove the partial profiles.

- e2e tests: Add tests for profilemerging
Adds e2e tests that record a number of partial profiles where one of the pods in the recorded deployment performs a different action than the others, then merges the partial profiles and ensures that the merged profile is in fact a union of the partial profiles.

- bundle: update with the recent changes
Does what it says

Fixes #677

#### Does this PR have test?
Yes

I'll point some inline

```release-note
The profileRecording CR contains a new optional field mergeStrategy, defaulting
to none. When this field is set to containers, the recorded profiles will be set
as partial, both using a label and the profile status. These profiles are not
reconciled until the profileRecording exists, at which point the partial profiles
are merged and a union of the partial profiles is created and finally reconciled.

This allows for easier recording of policies e.g. during longer e2e runs which
are recording the policies.
```


jhrozek:
I'll put some questions for reviewers tomorrow.

internal/pkg/daemon/profilerecorder/profilerecorder.go
```diff
@@ -423,9 +433,28 @@ func (r *RecorderReconciler) collectLogProfiles(
 func (r *RecorderReconciler) collectLogSeccompProfile(
 	ctx context.Context,
 	enricherClient enricherapi.EnricherClient,
+	parsedProfileName *parsedAnnotation,
 	profileNamespacedName types.NamespacedName,
 	profileID string,
 ) error {
+	labels, err := profileLabels(
```
jhrozek, line 440: there's a bit of code duplication between the three recording functions, something to clean up perhaps

api/selinuxprofile/v1alpha2/selinuxprofile_types.go
```diff
@@ -98,6 +101,30 @@ type SelinuxProfile struct {
 	Status SelinuxProfileStatus `json:"status,omitempty"`
 }
 
+func (sp *SelinuxProfile) Merge(other profilebasev1alpha1.MergeableProfile) error {
+	// TODO(jhrozek): should we be defensive about checking if other attributes match as well? (e.g. inherit)
```
jhrozek, line 105: should we? The profiles to merge /should/ come from the same recording, but I guess additional hardening wouldn't hurt?

api/seccompprofile/v1beta1/seccompprofile_types.go
```diff
@@ -183,3 +212,30 @@ type SeccompProfileList struct {
 func init() { //nolint:gochecknoinits // required to init scheme
 	SchemeBuilder.Register(&SeccompProfile{}, &SeccompProfileList{})
 }
+
+func UnionSyscalls(baseSyscalls, appliedSyscalls []*Syscall) []*Syscall {
```
saschagrunert, line 216: Do we wanna make this private? This function should be only used internally since it does not handle arches and arguments.

api/selinuxprofile/v1alpha2/selinuxprofile_types.go
saschagrunert: I'd say we make this private so that other projects do not use it via the API.

saschagrunert (commented):
Great work, I just have a bunch of comments regarding the API visibility.

api/selinuxprofile/v1alpha2/selinuxprofile_types.go
jhrozek: By making this private do you mean "not exposed by the API modules directly"?
I'm thinking how to make this private while still letting the merge controller use an interface and not care about what profile it is that they are merging..would something like the following work for you?

```
type MergeableProfile interface {
 merge()
}

type MergeableSeccompProfile struct {
   SeccompProfile
}

func (msp *MergeableSeccompProfile) merge(....) error {
   ...
}

type MergeableSelinuxProfile struct {
   SelinuxProfile
}

func (msp *MergeableSelinuxProfile)  merge(....) error {
   ...
}
```
this might a private API in the merge controller or somewhere else in utils/ but not exposed outside SPO...

jhrozek: correction: The `Merge` method could be exported outside its module, just not in the API. I guess that would still satisfy what you asked for?

api/profilebase/v1alpha1/profilebase.go
```diff
@@ -17,13 +17,68 @@ limitations under the License.
...
+	cli client.Client,
+	recordingName string,
+	recordingNamespace string,
+	list client.ObjectList,
+) ([]metav1.Object, error) {
+	if err := cli.List(
+		ctx,
+		list,
+		client.InNamespace(recordingNamespace),
+		client.MatchingLabels{v1alpha1.ProfileToRecordingLabel: recordingName}); err != nil {
```
JAORMX, line 64: We might need to pre-parse the `recordingName` to ensure it's usable as a label. Remember labels have a length limitation. We might need to limit recording names or hash in case the length is too long.

api/seccompprofile/v1beta1/seccompprofile_types.go
JAORMX: I agree, I don't think we need to make this public for now.

api/seccompprofile/v1beta1/seccompprofile_types.go
```diff
@@ -183,3 +212,30 @@ type SeccompProfileList struct {
...
+	allSyscalls := make(map[seccomp.Action]map[string]bool)
```
JAORMX, line 217: uhm... you already have a notion of how long the syscall lists will be from the two objects being passed. Why not directly allocate this map with the length of the longest list of the two? That way you reduce the number of allocations made at runtime.

api/seccompprofile/v1beta1/seccompprofile_types.go
```diff
@@ -183,3 +212,30 @@ type SeccompProfileList struct {
...
+	}
+	for _, s := range appliedSyscalls {
+		if _, ok := allSyscalls[s.Action]; !ok {
+			allSyscalls[s.Action] = make(map[string]bool)
+		}
+		for _, n := range s.Names {
+			allSyscalls[s.Action][n] = true
+		}
+	}
+	finalSyscalls := make([]*Syscall, 0, len(appliedSyscalls)+len(baseSyscalls))
```
JAORMX, line 232: doesn't this allocate more memory than you might need? some, if not most, of the entries of these profiles will be the same. Seems to me like it would be preferable to allocate only for the length of the longest array. The rest of the allocations would be handled via the `append` function.

JAORMX (changes requested):

api/profilebase/v1alpha1/profilebase.go
jhrozek: Good call!

Do you (or @sascha) have a preference? I think I prefer limiting the names, just because the hashes are user-unfriendly and there's really no strict need to keep them unique (unlike e.g. node labels).

An easy way might be to add an event in the webhook recorder that says the recording name is too long and no recording will happen, but I'm not sure if it's super well discoverable. Unfortunately we don't have a profilerecording controller except the merger. Actually, thinking out loud, the merger might be a good place to do that, even if it's not strictly its job. (I'd prefer to keep the webhooks as simple as possible).

This wouldn't prevent the recordings from being created with names that are too long, though we'd need an admission webhook for that which I think is really an overkill..

api/seccompprofile/v1beta1/seccompprofile_types.go
jhrozek: I can fix these comments (they are valid) just note that I just lifted the existing function from the seccomp controller :-) and didn't write it from scratch.

api/seccompprofile/v1beta1/seccompprofile_types.go
jhrozek: yes, probably

api/profilebase/v1alpha1/profilebase.go
JAORMX: No preferences on my side. I'd love it if we could just leverage the CRD definitions to limit the name size (as we can add validations to fields in the CRD)... but I'm not sure and I don't think that's possible. So... whatever you see fit. All I'd like is that this is taken into account.

api/seccompprofile/v1beta1/seccompprofile_types.go
JAORMX: @jhrozek ah! I hadn't noticed.

api/selinuxprofile/v1alpha2/selinuxprofile_types.go
jhrozek: Done, please check

api/seccompprofile/v1beta1/seccompprofile_types.go
jhrozek: I moved UnionSyscalls into an util module where it can be reused by both the deamon and the controller

api/profilebase/v1alpha1/profilebase.go
jhrozek: OK, there's a new comment that emits an event in the webhook as well as error when collecting the recording. Instead of the error we could be more permissive and just create the recording as usual (so not partial) and without the label.

api/seccompprofile/v1beta1/seccompprofile_types.go
jhrozek: Done in a separate commit.

api/seccompprofile/v1beta1/seccompprofile_types.go
jhrozek: Done in a separate commit.

jhrozek:
@JAORMX I tried to implement your suggestion wrt the ownership instead of finalizers and I couldn't make it work. I think you meant to use [foreground cascading deletion](https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#use-foreground-cascading-deletion) right?
That works fine, but you need to delete the object (in our case profile recording) while passing a `DeleteOption` (see the doc). When just deleting the object as the admin would probably do and as `kubectl delete` does, the dependant objects were just removed without adding the finalizer. Adding the finalizer myself as a hack didn't work either. We can discuss this more when you're back (I have the experimental patch still in a special branch).

installation-usage.md
```diff
@@ -576,6 +577,116 @@ NAME                 STATUS      AGE
...
+profile. This is mostly visible when recording pods managed by a replicating
+controller (Deployment, DaemonSet, etc.). A realistic example might
+be a workload being recorded in a test environment where the recorded
+Deployment consists of several replicas, only one of which is receiving
+the test traffic. After the recording is complete, only the container that
+was receiving the traffic would have container all the syscalls that were
+actually used.
+
+In this case, it might be useful to merge the per-container profiles
+into a single profile. This can be done by setting the `mergeStrategy`
```
rhmdnd, line 592: For my own understanding, but this is useful because it makes the one true profile easier to find, correct?

Otherwise, I could fumble through the profiles and grab one that doesn't contain all the syscalls, since it wasn't used at runtime?

installation-usage.md
```diff
@@ -576,6 +577,116 @@ NAME                 STATUS      AGE
...
+attribute to `containers` in the `ProfileRecording`. Note that the following
```
rhmdnd, line 593: Are `none` (the default) and `containers` the only two options for `mergeStrategy`?

api/profilebase/v1alpha1/profilebase.go
rhmdnd: > I'm not sure myself what's preferable, but error seems somehow "safer" than guessing.

What would happen if we didn't error out? Would we truncate and continue?

api/profilerecording/v1alpha1/profilerecording_types.go
```diff
@@ -40,6 +40,23 @@ const (
...
+	ProfileMergeContainers ProfileMergeStrategy = "containers"
+)
+
+const (
+	// ProfileToRecordingLabel is the name of the ProfileRecording CR that produced this profile.
+	ProfileToRecordingLabel = "spo.x-k8s.io/recording-id"
+	// ProfileToContainerLabel is the name of the container that produced this profile.
+	ProfileToContainerLabel = "spo.x-k8s.io/container-id"
+	// RecordingHasUnmergedProfiles is a finalizer that indicates that the recording has partial policies. Its
+	// main use is as to hold off the deletion of the recording until all partial profiles are merged.
```
rhmdnd, line 56: nit: s/as//

api/profilerecording/v1alpha1/profilerecording_types.go
```diff
@@ -50,6 +67,13 @@ type ProfileRecordingSpec struct {
 	// +kubebuilder:validation:Enum=bpf;logs
 	Recorder ProfileRecorder `json:"recorder"`
 
+	// Whether or how to merge recorded profiles. Can be one of "none" or "containers".
```
rhmdnd, line 70: Ok - this answers a question I had elsewhere.

What other types of merge strategies do we anticipate? If I use the `containers` strategy, my profile recording will be an aggregate of everything that happened across the containers. Do we anticipate merging recordings across other entities , like Deployments and ReplicaSets?

For example, would I be able to distill recordings from multiple Deployments if I set `deployment` as the merge strategy (assuming we implemented something like that?)

To merge, or not to merge, feels like a binary decision and the setting here eludes to potential functionality that adds another dimension to record merging, so I'm trying to understand that bit.

api/secprofnodestatus/v1alpha1/secprofnodestatus_types.go
```diff
@@ -63,9 +65,10 @@ func LowerOfTwoStates(currentLowest, candidate ProfileState) ProfileState {
 	orderedStates := make(map[ProfileState]int)
 	orderedStates[ProfileStateError] = 0       // error must always have the lowest index
 	orderedStates[ProfileStateTerminating] = 1 // If one is set as terminating; all the statuses will end here too
-	orderedStates[ProfileStatePending] = 2
-	orderedStates[ProfileStateInProgress] = 3
-	orderedStates[ProfileStateInstalled] = 4
```
rhmdnd, line 68: Since this is a public method, does changing the mapping between the state and the integer matter?

jhrozek:
just a rebase, no other changes.

installation-usage.md
jhrozek: That, but also, the recording merging might be more useful than the default no merging in case the profile is recorded in an e2e test suite, where you might test several different test-cases and what you want to allow is probably a union of the cases. Another use-case are replicated pod controllers (daemonSets, deployments) where you run several replicas of the same pod, but the test would only trigger an action on one of them (e.g. you have 10 replicas of nginx listening on port 80, which probably gives you `bind()` and `listen()`, but if you curl from one of them, only that replica would record `accept()`).

You are right that the profile merging is a usability vs security tradeoff. If you wanted to be very secure, you'd review every recorded policy, but since the target "audience" are workloads that are privileged in some way and have no policy at all, even recording a merger of several would probably be quite a bit more secure that none.

You raised some good points, though. Should I add more context to the docs? Please feel free to suggest what might be missing or needs clarification, I might suffer from "tunnel vision" being familiar with the code.

installation-usage.md
jhrozek: right now, yes. I was also wondering about an option called `workload` or even per-workload options like `deployment` like you suggested earlier which might produce just one policy per workload, but so far I haven't seen a good use-case for that.

api/profilerecording/v1alpha1/profilerecording_types.go
jhrozek: I was thinking that maybe the workload level might be a good candidate for future or even a profile recording level (so, everything recorded using a single profileRecording would result in a single policy), but the container-level merging seemed like a good start in terms of being less granular than per-container-instance which is the `none` default but still not merging too much. As I said elsewhere, the profile merging is a bit of a security/usability tradeoff where you trade not having to deal with potentially dozens of policies after a recording session vs. the danger that one of the policies record something you don't want in the resulting policy.

I wonder if @saschagrunert or @JAORMX have other opinions?

api/profilerecording/v1alpha1/profilerecording_types.go
jhrozek: thanks, fixed

jhrozek:
@rhmdnd for some reason I couldn't reply to this comment inline, replying as a comment instead:


>>    I'm not sure myself what's preferable, but error seems somehow "safer" than guessing.
> What would happen if we didn't error out? Would we truncate and continue?

So maybe the comment is confusing, it was referencing a potential issue if the profilerecording was too long, then its name wouldn't be possible to use as a label value (k8s would error out on using the label value, in this case when the `list()` method is called). I opted for erroring out proactively, what I meant with guessing was trying to list the profiles based on e.g. the recorded container name or the profile name itself. But that seems too hacky. Ultimately I think we want to make use of a new k8s feature which allows to validate CRD attributes without resorting to webhooks (see issue #1199) and just limit the length of the name of the profile bindings.

api/profilerecording/v1alpha1/profilerecording_types.go
saschagrunert: I agree leaving the option for further higher level merges is a good path forward.

saschagrunert (approved):
Great work!

PR #1189, opened 2022-09-27
Allow the access() syscall from the operator seccomp profile

jhrozek:
#### What type of PR is this?
/kind bug

OpenShift uses a fork of Go which in turn uses OpenSSL for crypto
instead of BoringSSL:
https://developers.redhat.com/blog/2019/06/24/go-and-fips-140-2-on-red-hat-enterprise-linux

In order to support running SPO that was compiled with that version of
Go, we need to allow the access() system call to avoid a crash.

None

#### Does this PR have test?
N/A, I tested manually on OCP 4.12

N/A

```release-note
NONE
```


pjbgf (commented):
/lgtm

PR #1216, opened 2022-10-11
Fix flaky e2e issue

Vincent056:
Fix the pod creation error in seccomp profile binding test, and fix seccomp profile finalizer test.

In seccomp profile finalizer test, when we first delete the profile recording before the pod is removed, it will not be removed because of the finalizer, however as soon as the pod is removed, the profile recording will be removed, hence in the next step, we were not able to find the object.

Flaky e2e issue

#### Does this PR have test?

N/A

None

```release-note

```


JAORMX (approved):
/lgtm

PR #1217, opened 2022-10-11
Add Affinity to SPOD configuration

ccojocar:
#### What type of PR is this?

/kind api-change 

It adds Affinity to the SPOD configuration. 

This allows to control the scheduling of spod daemonset pod in a cluster which has nodes with different types of runtimes (e.g. micro-vm, docker, cri-o) which are not tainted but just labeled. 

```release-note
Added Affinity to SPOD configuration.
```


ccojocar:
The unit tests failure on macos doesn't seem to be related to this change. It looks like a transient failure. The test passed when executed locally on macos.

jhrozek:
> The unit tests failure on macos doesn't seem to be related to this change. It looks like a transient failure. The test passed when executed locally on macos.

Yes, I've seen those, too. I kicked the test now.

jhrozek:
The code looks good, I just wonder if you think this is worth documenting in the `installation-usage.md` document?

ccojocar:
> The code looks good, I just wonder if you think this is worth documenting in the installation-usage.md document?

I can add a note.

PR #1227, opened 2022-10-13
Add imagePullSecrets property to SPOD configuration

ccojocar:
#### What type of PR is this?

/kind api-change

Adds the `imagePullSecrets` property to SPOD configuration. This is required to pull the images from spod pod from a private registry. For instance when these images are mirrored in a private registry instead of using them from public registry.

```release-note
Add imagePullSecrets to SPOD configuration.

```


saschagrunert (approved):

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -587,6 +587,7 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 
 	templateSpec.Tolerations = cfg.Spec.Tolerations
 	templateSpec.Affinity = cfg.Spec.Affinity
+	templateSpec.ImagePullSecrets = cfg.Spec.ImagePullSecrets
```
saschagrunert, line 590: I'm wondering if we should apply this to all other images we pull as well. :thinking:

saschagrunert:
CI is not happy with the TOC: 
```
+- [Pull images from private registry](#pull-images-from-private-registry)
```

internal/pkg/manager/spod/spod_controller.go
ccojocar: Do you mean in the webhook binding?

saschagrunert: Yeah for example, we also use kube rbac proxy and things like that but their manifests are mostly static.

ccojocar: I configured it also in the webhook default deployment. Are there any other place which we should consider?

saschagrunert: The daemon probably.

saschagrunert (approved):

saschagrunert:
@jhrozek @JAORMX PTAL
/hold for their eyes

internal/pkg/manager/spod/spod_controller.go
ccojocar: I am a bit confused. Isn't the spod the daemon? I think the other resources get deployed declaratively using manifests.

jhrozek (approved):
LGTM, too

ccojocar:
Thanks @jhrozek  for review, @saschagrunert Can we merge this now? Thanks

saschagrunert (approved):
/unhold

PR #1232, opened 2022-10-14
spod: Make the webhook's ObjectSelector configurable

jhrozek:
#### What type of PR is this?
/kind feature
/kind api-change

Implements a suggestion from code review of the PR that made webhooks
opt-in. I wanted to document a way to make the objectSelector
configurable and to my surprise found out that this is not possible
because we don't implement configurable objectSelectors.

None

#### Does this PR have test?
Yes.

I don't think we absulutely need this for 0.5.0, meaning that I wouldn't
delay the release.

```release-note
The spod CR now has a new attribute objectSelector that allows to configure
which objects would SPO's webhooks match on. By default, the selector matches
all, but setting the selector to include e.g. only certain labels might be a way
to further ensure that possible bugs in the webhooks don't affect the rest of
the cluster.
```


jhrozek:
In PR #1207, @ccojocar suggested that we should we document setting the objectSelector to further restrict the objects that the webhooks match. I was surprised to find out that we don't expose objectSelector, but only namespaceSelector, so I just went ahead and exposed the objectSelector as well.

I don't think we need to delay the 0.5.0 release over this, I just didn't want Cosmin's suggestion to get forgotten because I agree that in general it's a good idea to not overuse webhooks in a cluster.

ccojocar (commented):
lgtm, thanks for adding this!

saschagrunert (approved):

PR #1236, opened 2022-10-17
handle node deletion

### LINKED ISSUE #1152: Unable to delete seccompprofile if one of node is deleted
#### What happened:

Unable to delete the seccompprofiles if one of the node deleted/updated because of the finalizers is not updated with the latest nodes. I'm not sure if this bug or as per design.
Thank you

Possible if SPO update seccompprofiles on the Finalizers part to the latest k8s nodes?

- Start minikube with 3 nodes
```$ minikube  start --kubernetes-version=v1.21.14 --nodes=3 --memory 4000 --cpus```
- Install SPO
- Check the default SeccomProfiles, there are 3 finalizers
```
$ kubectl get seccompprofiles -n security-profiles-operator log-enricher-trace -o json | jq .metadata.finalizers
[
            "minikube-m03-delete",
            "minikube-m02-delete",
            "minikube-delete"
]
```
- Delete one of the node
```
$ minikube node delete minikube-m03
```
- The SeccompProfiles still have `minikube-m03` as finalizers
```
$ kubectl get seccompprofiles -n security-profiles-operator log-enricher-trace -o json | jq .metadata.finalizers
[
            "minikube-m03-delete",
            "minikube-m02-delete",
            "minikube-delete"
]
```
- I can't delete the seccompprofiles because the `"minikube-m03-delete"` is still exist (the old node that already deleted), unless if i patch it the finalizers like:
```
$ kubectl patch seccompprofiles log-enricher-trace -n security-profiles-operator -p '{"metadata":{"finalizers":[]}}' --type=merge
```

I observed this on our Amazon EKS cluster, but for this testing i tested it on minikube with the same behavior. On our Amazon EKS clusters we have patching cycle that will delete the old node and bring up the new node (with new latest image). So after the patching the finalizers on the seccompprofiles still showing the old nodes, this make we can't delete the seccompprofles.

#### Environment:

- Kernel (e.g. `uname -a`):
- Others:

### PR DESCRIPTION
**Pull request by Vincent056**
#### What type of PR is this?
/kind bug

This pr fixes seccompprofiles deletion when a node is removed, we added a check to see if the node finalizer is a deleted node, if so, we remove such finalizer so the seccompprofile can be deleted without any issues.

The nodestatus controller detect the extra nodestatus object, and if it can't find the node associated with it, it will try to remove the node status object and node finalizer from corresponding the seccomp profile

Fixes #1152

#### Does this PR have test?

Yes

NONE

```release-note
This pr fixes seccompprofiles deletion when a node is removed, we added a check to see if the node finalizer is a deleted node, if so, we remove such finalizer so the seccompprofile can be deleted without any issues.
```

### DIFF
--- a/deploy/webhook-operator.yaml
+++ b/deploy/webhook-operator.yaml
@@ -2090,6 +2090,14 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
   - services
   verbs:
   - create
@@ -2252,6 +2260,7 @@ rules:
   resources:
   - securityprofilenodestatuses
   verbs:
+  - delete
   - get
   - list
   - watch
--- a/internal/pkg/manager/nodestatus/nodestatus.go
+++ b/internal/pkg/manager/nodestatus/nodestatus.go
@@ -27,6 +27,7 @@ import (
 	"github.com/crossplane/crossplane-runtime/pkg/event"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/selection"
@@ -98,7 +99,8 @@ func (r *StatusReconciler) Healthz(*http.Request) error {
 
 // Security Profiles Operator RBAC permissions to manage Node Statuses
 //nolint:lll // required for kubebuilder
-// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=securityprofilenodestatuses,verbs=get;list;watch
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=securityprofilenodestatuses,verbs=get;list;watch;delete
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch
 
 // Reconcile reconciles a NodeStatus.
 func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (reconcile.Result, error) {
@@ -167,11 +169,26 @@ func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 	// make sure we have all the statuses already
 	hasStatuses := len(nodeStatusList.Items)
 	wantsStatuses := spodDS.Status.DesiredNumberScheduled
-	if wantsStatuses != int32(hasStatuses) {
+	if wantsStatuses > int32(hasStatuses) {
 		logger.Info("Not updating policy: not all statuses are ready",
 			"has", hasStatuses, "wants", wantsStatuses)
 		// Don't reconcile again, let's just wait for another update
 		return reconcile.Result{}, nil
+	} else if wantsStatuses < int32(hasStatuses) {
+		// this happens when nodes are removed from the cluster
+		logger.Info("Removing extra statuses", "has", hasStatuses, "wants", wantsStatuses)
+		nodeName, err := r.removeStatusForDeletedNode(ctx, nodeStatusList, lprof)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("cannot remove extra statuses: %w", err)
+		}
+		if nodeName != "" {
+			// remove the deleted node finalizer string from the profile
+			logger.Info("Removing finalizer from profile", "profile", prof.GetName(), "node", nodeName)
+			if err := util.RemoveFinalizer(ctx, r.client, prof, util.GetFinalizerNodeString(nodeName)); err != nil {
+				return reconcile.Result{}, fmt.Errorf("cannot remove finalizer from profile: %w", err)
+			}
+		}
+		return reconcile.Result{Requeue: true}, nil
 	}
 
 	lowestCommonState := statusv1alpha1.LowestState
@@ -181,6 +198,27 @@ func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (
 	logger.V(config.VerboseLevel).Info("Setting the status to", "Status", lowestCommonState)
 
 	return r.reconcileStatus(ctx, prof, lowestCommonState, lprof)
+}
+
+// removeStatusForDeletedNode removes the status for a node that has been deleted.
+func (r *StatusReconciler) removeStatusForDeletedNode(ctx context.Context,
+	nodeStatusList *statusv1alpha1.SecurityProfileNodeStatusList, logger logr.Logger,
+) (string, error) {
+	for i := range nodeStatusList.Items {
+		nodeName := nodeStatusList.Items[i].NodeName
+		node := &v1.Node{}
+		if err := r.client.Get(ctx, types.NamespacedName{Name: nodeName}, node); err != nil {
+			if util.IsNotFoundOrConflict(err) {
+				logger.Info("Removing node status for removed node", "node", nodeName)
+				if err := r.client.Delete(ctx, &nodeStatusList.Items[i]); err != nil {
+					return "", fmt.Errorf("cannot delete node status: %w", err)
+				}
+				return nodeName, nil
+			}
+			return "", fmt.Errorf("cannot get node: %w", err)
+		}
+	}
+	return "", nil
 }
 
 func (r *StatusReconciler) getDS(ctx context.Context, namespace string, l logr.Logger) (*appsv1.DaemonSet, error) {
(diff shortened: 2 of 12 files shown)


vendor/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil/controllerutil.go
```diff
@@ -387,6 +390,57 @@ func ContainsFinalizer(o client.Object, finalizer string) bool {
 	return false
 }
 
+// RemoveDeletedNodeFinalizer removes finalizers from the object for deleted nodes.
+func RemoveDeletedNodeFinalizer(c client.Client, o client.Object) (error, string) {
```
jhrozek, line 394: Please don't patch files under `vendor/`, that should be an exception (e.g. if we need to bring in a patch that hasn't been released or merged yet)

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -468,6 +468,23 @@ func (r *Reconciler) reconcileDeletion(
 		return ctrl.Result{}, fmt.Errorf("deleting node status/finalizer for deleted SeccompProfile: %w", err)
 	}
 
+	// Remove the finalizer for deleted Nodes
+	err, nodeName := controllerutil.RemoveDeletedNodeFinalizer(r.client, sp)
```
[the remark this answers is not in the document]
Vincent056: that's a good point, I will do that

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -128,6 +127,17 @@ func (nsf *StatusClient) statusObj(
 	}
 }
 
+func (nsf *StatusClient) statusObjNode(
+	nodeName string,
+) *secprofnodestatusv1alpha1.SecurityProfileNodeStatus {
+	return &secprofnodestatusv1alpha1.SecurityProfileNodeStatus{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nsf.pol.GetName() + "-" + nodeName,
```
jhrozek, line 135: could we reuse this in `statusObj` or just use `statusObj` instead?

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -165,6 +175,37 @@ func (nsf *StatusClient) Remove(ctx context.Context, c client.Client) error {
...
+		return fmt.Errorf("cannot clean up deleted node for %s: %w", nsf.pol.GetName(), err)
+	}
+
+	return nil
+}
+
+// remove the finalizer and delete the nodestatus for deleted node.
+func (nsf *StatusClient) deletedNodeCleanUp(ctx context.Context, c client.Client) error {
+	return util.Retry(func() error {
+		nodeName, err := util.RemoveDeletedNodeFinalizer(ctx, c, nsf.pol)
```
jhrozek, line 189: should this be modified to handle more than one deleted node?

jhrozek (commented):
nice progress, have you had a chance to test the PR manually yet while you're working on the tests?

internal/pkg/nodestatus/nodestatus.go
Vincent056: > nice progress, have you had a chance to test the PR manually yet while you're working on the tests?

I changed the implementation, I tested the patch manually, and it works

Vincent056:
@jhrozek for e2e testing, since we can't delete node, what I did is to add a nodestatus object, and a finalizer to the seccomp profile to mimic a deleted node. The nodestatus controller should detect the extra nodestatus object, and if it can't find the node associated with it, it will try to remove the node status object and node finalizer from the seccomp profile

saschagrunert:
> @saschagrunert Could you add me to the white list, so I don't have to bug someone every time for /ok-to-test

If you want to then we could add you as a reviewer to this project. This would also mean that you get notifications for other PRs as well :)

jhrozek:
So the good thing is that the patch works. I tested on OpenShift, by scaling a `MachineSet` up, made sure both the finalizers and the `securityprofilenodestatus` objects reflect the new nodes and then scaled the `MachineSet` back down and made sure the finalizer and the status are gone.

To fix the CI, you should rebase on the current origin/main to fix some of the jobs and make sure the branch you rebase atop contains both PR #1253 and #1252. This should fix the `build/ubi-image`, `build/operator-image` and `test/e2e-flatcar` tests respectively.

Please also run `make bundle` and commit the diff to fix the `olm/tests` job.

jhrozek:
also to get around the `release-note-needed` label, either please move the `NONE` text into the three-backticks below the `NONE` or even better, write a release note sentence into the backticks :-) (This is similar to how we keep a changelog in CO)

deploy/webhook-operator.yaml
```diff
@@ -1202,7 +1202,15 @@ rules:
...
   - ""
   resources:
-  - services
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - admissionregistration.k8s.io
```
jhrozek, line 1211: is this a merge conflict resolution? It doesn't seem like we need to modify the webhook-operator yaml to touch the `mutatingwebhookconfigurations` APIs do we?

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -167,12 +169,28 @@ func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (
...
 	hasStatuses := len(nodeStatusList.Items)
 	wantsStatuses := spodDS.Status.DesiredNumberScheduled
-	if wantsStatuses != int32(hasStatuses) {
+	if wantsStatuses > int32(hasStatuses) {
 		logger.Info("Not updating policy: not all statuses are ready",
 			"has", hasStatuses, "wants", wantsStatuses)
 		// Don't reconcile again, let's just wait for another update
 		return reconcile.Result{}, nil
 	}
+	if wantsStatuses < int32(hasStatuses) {
```
jhrozek, line 178: I think it would be cleaner to do an else-if instead of two ifs
```
if wantsStatuses > int32(hasStatuses) {
 ...
} else if wantsStatuses < int32(hasStatuses) {
 ...
} 

// we know they are equal here
```
this is just to make it clear in the code that we are still continuing the same if

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -167,12 +169,28 @@ func (r *StatusReconciler) Reconcile(_ context.Context, req reconcile.Request) (
...
+			return reconcile.Result{}, fmt.Errorf("cannot remove extra statuses: %w", err)
+		}
+		if nodeName != "" {
+			// remove the deleted node finalizer string from the profile
+			logger.Info("Removing finalizer from profile", "profile", prof.GetName(), "node", nodeName)
+			if err := util.RemoveFinalizer(ctx, r.client, prof, util.GetFinalizerNodeString(nodeName)); err != nil {
+				return reconcile.Result{}, fmt.Errorf("cannot remove finalizer from profile: %w", err)
+			}
+		}
+		return reconcile.Result{}, nil
```
jhrozek, line 192: since this code would only work for one node removal, should we force another reconcile loop here by returning:
```
return reconcile.Result{Requeue: true}, nil
```
this would allow us to both keep the logic simple to just remove one finalizer and one node object for 99% case where just one node goes away, but it would also cover the case where more nodes go away through multiple reconcile attempts.

jhrozek (commented):
just some nits, but this works, great!

Vincent056:
> > @saschagrunert Could you add me to the white list, so I don't have to bug someone every time for /ok-to-test
> 
> If you want to then we could add you as a reviewer to this project. This would also mean that you get notifications for other PRs as well :)

Thanks for adding me as a reviewer!

deploy/webhook-operator.yaml
Vincent056: We didn't touch `mutatingwebhookconfigurations` in this PR, it seems to be auto-generated from `make bundle`

jhrozek: Ah I know now. Because adding the `node` RBAC rules added a first item to the `ClusterRole` the manager is running as, we need to adjust the webhook operator's kustomize manifests:
```
diff --git a/deploy/overlays/webhook/role.yaml b/deploy/overlays/webhook/role.yaml
index 97221506..f5c1444a 100644
--- a/deploy/overlays/webhook/role.yaml
+++ b/deploy/overlays/webhook/role.yaml
@@ -1,3 +1,3 @@
 ---
-- path: /rules/1
+- path: /rules/2
   op: remove
```
apply the above, then run `make deployments`. That did the trick for me.

.github/workflows/build.yml
```diff
@@ -64,7 +64,7 @@ jobs:
       - uses: actions/checkout@v3
       - name: Set up Docker Buildx
         id: buildx
-        uses: docker/setup-buildx-action@c74574e6c82eeedc46366be1b0d287eff9085eb6
+        uses: docker/setup-buildx-action@8c0edbc76e98fa90f69d9a2c020dcb50019dc325
```
jhrozek, line 67: these don't belong in this PR (and neither does the commit message :-)). `git rebase origin/main` should get rid of these.

jhrozek (changes requested):
the last issue is the namespaced webhook manifests and the commit message, otherwise I think we're good.
We're getting there, nice work!

jhrozek:
btw I made a pass on the latest code functionality-wise and everything still works. So let's fix the nits and merge the patch!

Vincent056:
> the last issue is the namespaced webhook manifests and the commit message, otherwise I think we're good. We're getting there, nice work!


Thanks for the help, I was able to fix the RBAC issue.

PR #1239, opened 2022-10-17
Use security context to set up the default seccomp profile for spod pod and security-profiles-operator container

ccojocar:

```release-note

```


saschagrunert:
Yes I can re-kick the tests once the others finished. Do you need this change as part of v0.5.0, @ccojocar ?

ccojocar:
I think the issue is that cri-o runtime requires the profile path to be prefix with `localhost` while other runtime such as docker doesn't: https://github.com/cri-o/cri-o/blob/main/internal/config/seccomp/seccomp.go#L240

I am not sure what's the best way to go about this. How can we detect the runtime and configure the path accordingly?

PR #1252, opened 2022-10-21
Use directly the box url instead of the metadata when downloading the vagrant box for flatcar linux

ccojocar:
#### What type of PR is this?

/kind bug

It seems that vagrant resolve the relative path from the vagrant metadata to a local file path instead of URL.

See the URL in the metadata file https://lts.release.flatcar-linux.net/amd64-usr/current/flatcar_production_vagrant.json
This gets resolved to a file:// instead to a url:// due to https://github.com/hashicorp/vagrant/blob/87689f5f196cf31dea057eb91aa34a09949ebc8a/lib/vagrant/action/builtin/box_add.rb#L412.
It seems that there is not way to configure this currently.

```release-note
Fix integration tests for Flatcar Linux.

```

### DIFF
--- a/hack/ci/Vagrantfile-flatcar
+++ b/hack/ci/Vagrantfile-flatcar
@@ -4,7 +4,7 @@
 Vagrant.configure("2") do |config|
   config.vm.box = "flatcar-lts"
   config.vm.box_check_update = true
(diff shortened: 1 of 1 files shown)


JAORMX:
/lgtm cancel

let's wait for the flatcar job to pass first.

saschagrunert (approved):

PR #1255, opened 2022-10-21
Configure the default local seccomp profile according to the runtime

ccojocar:
#### What type of PR is this?

/kind bug

Currently the spod fails to start on cri-o runtime because unlike containerd or docker, the cri-o runtime expects the local seccomp profile to be prefixed with 'localhost'.

See for more details https://github.com/cri-o/cri-o/blob/1e6fd9c520d03d47835d1d4c3209e0f77c38f542/internal/config/seccomp/seccomp.go#L240

#### Does this PR have test?

yes 

```release-note
Configure the default local seccomp profile according to the runtime (e.g. cri-o expects the profile to be prefixed with `localhost`).
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -477,9 +477,9 @@ OLM_EXAMPLES := \
 
 .PHONY: bundle
 bundle: operator-sdk deployments ## Generate bundle manifests and metadata, then validate generated files.
-	sed -i "s/\(olm.skipRange: '>=.*\)<.*'/\1<$(VERSION)'/" deploy/base/clusterserviceversion.yaml
-	sed -i "s/\(\"name\": \"security-profiles-operator.v\).*\"/\1$(VERSION)\"/" deploy/catalog-preamble.json
-	sed -i "s/\(\"skipRange\": \">=.*\)<.*\"/\1<$(VERSION)\"/" deploy/catalog-preamble.json
+	$(SED) "s/\(olm.skipRange: '>=.*\)<.*'/\1<$(VERSION)'/" deploy/base/clusterserviceversion.yaml
+	$(SED) "s/\(\"name\": \"security-profiles-operator.v\).*\"/\1$(VERSION)\"/" deploy/catalog-preamble.json
+	$(SED) "s/\(\"skipRange\": \">=.*\)<.*\"/\1<$(VERSION)\"/" deploy/catalog-preamble.json
 	cat $(OLM_EXAMPLES) $(BUNDLE_OPERATOR_MANIFEST) deploy/base/clusterserviceversion.yaml | $(OPERATOR_SDK) generate bundle -q --overwrite --version $(VERSION) $(BUNDLE_METADATA_OPTS)
 	git restore deploy/base/clusterserviceversion.yaml
 	mkdir -p ./bundle/tests/scorecard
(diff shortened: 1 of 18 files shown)


ccojocar:
@saschagrunert All checks finally passed after some reworking. Please could you have a look? Thanks

saschagrunert:
> Currently the spod fails to start on cri-o runtime because unlike containerd or docker, the cri-o runtime expects the local seccomp profile to be prefixed with 'localhost'.

I do not understand this. We test with CRI-O in the CI, so what is the error case?

ccojocar:
It seems to work on the CI because everything is copied into the root folder but we hit this place in our cluster when running on CRI-O: https://github.com/cri-o/cri-o/blob/1e6fd9c520d03d47835d1d4c3209e0f77c38f542/internal/config/seccomp/seccomp.go#L240

```
	// Load local seccomp profiles including their availability validation
	if !strings.HasPrefix(profilePath, k8sV1.SeccompLocalhostProfileNamePrefix) {
		return fmt.Errorf("unknown seccomp profile path: %q", profilePath)
	}
```

As far as I can understand, it expects that the profile is loaded from `localhost/...`.

saschagrunert:
> https://github.com/cri-o/cri-o/blob/1e6fd9c520d03d47835d1d4c3209e0f77c38f542/internal/config/seccomp/seccomp.go#L240

This is the code path for the annotation based seccomp profile, but we're using the fields here, right? So we should land there: https://github.com/cri-o/cri-o/blob/1e6fd9c520d03d47835d1d4c3209e0f77c38f542/internal/config/seccomp/seccomp.go#L296-L302

ccojocar:
I see but I don't get why in our case we hit the `setupFromPath` at least from what I see in the error message.

I was wondering because the fix here has no effect on CI cluster. The profile had to be copied under `localhost` in order to get it working on CI. 

On the `setupFromPath` case the prefix gets stripped off after the check, right?
https://github.com/cri-o/cri-o/blob/1e6fd9c520d03d47835d1d4c3209e0f77c38f542/internal/config/seccomp/seccomp.go#L244

So essential, it doesn't need to be copy under localhost folder, just the localhost prefix needs to be part of the path in the security context.

ccojocar:
It might be because the Kubernetes version in that cluster is 1.22 and the `Field based seccomp profiles` are available in the newer version.

ccojocar:
@saschagrunert if we want to support older version of kubernetes, we will need to prefix the path in the security context. Right? Do you know in which version the switch was made to field base?

saschagrunert:
Hm, we introduced the fields in 1.19 with seccomp going GA. Do we also use CRI-O 1.22 with k8s 1.22?

ccojocar:
> Hm, we introduced the fields in 1.19 with seccomp going GA. Do we also use CRI-O 1.22 with k8s 1.22?
yeah, we use `v1.22.5`.

ccojocar:
We will eventually upgrade, but for now we need to support this.

saschagrunert (approved):

ccojocar:
I should revert the changes in non-root enabler which should not copy the profile under `localhost` but we still need to add the `localhost` prefix into the security context but also check the Kubernetes version. I am thinking that for newer version this is not required.

ccojocar:
@saschagrunert Thanks for approving but I think this is not the final fix. As mentioned above we don't need to copy under localhost folder because the prefix is stripped out.

PR #1290, opened 2022-11-10
Add command argument to the daemon which allows to disable profile recorder controller

ccojocar:
#### What type of PR is this?

/kind feature

This PR adds an command argument to the daemon which allows to disable the profile recorder controller when neither bpf-recorder or log-enricher are enabled. 

```release-note
Add a command argument to the daemon which allows to disable the profile recorder controller.
```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -153,9 +153,10 @@ func main() {
 					Value: false,
 				},
 				&cli.BoolFlag{
-					Name:  recordingFlag,
-					Usage: "Listen for ProfileRecording API recources",
-					Value: false,
+					Name:    recordingFlag,
+					Usage:   "Listen for ProfileRecording API recources",
+					Value:   false,
+					EnvVars: []string{config.EnableRecordingEnvKey},
 				},
 			},
 		},
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -66,6 +66,7 @@ import (
 
 const (
 	jsonFlag           string = "json"
+	recordingFlag      string = "with-recording"
 	selinuxFlag        string = "with-selinux"
 	apparmorFlag       string = "with-apparmor"
 	webhookFlag        string = "webhook"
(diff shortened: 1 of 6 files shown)


jhrozek (commented):
Would you also agree with enabling or disabling the controller with an environment variable? I'm thinking this way because of OLM installs where it's very easy to pass an environment variable to the operator through a `Subscription` object.

ccojocar:
> Would you also agree with enabling or disabling the controller with an environment variable? I'm thinking this way because of OLM installs where it's very easy to pass an environment variable to the operator through a Subscription object.

I'll add an environment variable in addition. What do you think?

ccojocar:
> Also two nits - I think this should be a /kind feature and the switch (and the reason for introducing it) should be mentioned in installation-usage.md. Maybe if you think that the other switches should be mentioned as well, we could have a cleanup ticked that would document them all in one go, but currently our docs are not bad and I'd like to keep everything documented nicely.

I'll do this and also try to add a test.

internal/pkg/config/config.go
```diff
@@ -70,6 +70,9 @@ const (
 	// EnableLogEnricherEnvKey is the environment variable key for enabling the log enricher.
 	EnableLogEnricherEnvKey = "ENABLE_LOG_ENRICHER"
 
+	// EnableRecordingEnvKey is the environment variable key to enabeling profile recording.
```
pjbgf, line 73:
```suggestion
	// EnableRecordingEnvKey is the environment variable key to enabling profile recording.
```

installation-usage.md
```diff
@@ -757,6 +758,16 @@ including the `mknod` syscall:
   - mknod
 ```
 
+#### Disable profile recording
+
+Profile recorder controller along with the corresponding sidecar container is disabled
+when neither `enableBpfRecorder` nor `enableLogEnricher` is set in the SPOD configuration, and
+automatically enabled when either one of them is on.
+
+Also, when running the daemon in standalone mode is possible to switch on the profile recorder
```
pjbgf, line 767:
```suggestion
Also, when running the daemon in standalone mode it is possible to switch on the profile recorder
```

pjbgf (commented):
Tiny nits, apart from that it is a lgtm from me. Thanks for working on this @ccojocar :bow:.

cmd/security-profiles-operator/main.go
```diff
@@ -151,6 +152,12 @@ func main() {
 					Usage: "Listen for AppArmor API resources",
 					Value: false,
 				},
+				&cli.BoolFlag{
+					Name:    recordingFlag,
+					Usage:   "Listen for ProfileRecording API recources",
```
pjbgf, line 157:
```suggestion
					Usage:   "Listen for ProfileRecording API resources",
```

saschagrunert (approved):

PR #1291, opened 2022-11-10
Configurable the daemon resource requirements

ccojocar:
#### What type of PR is this?

/kind feature

It makes configurable the daemon resource requirements. This is useful when the daemon requires different resource requirements due to the cluster size (e.g. number of pods is watching for).

```release-note
Make the daemon resource requirements configurable.

```


jhrozek:
It would be nice to document the new options. I'm on the fence about a test, ideally we should have everything tested (and e.g. we do have tests for webhook label changes), but the patch "just" adds a public kube type, so..*shrug*.

By the way, thanks a lot for doing this performance and scalability work!

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -472,6 +472,11 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 		}
 	}
 
+	// Overwrite the SPOD's default resoruce requirements
```
pjbgf, line 475:
```suggestion
	// Overwrite the SPOD's default resource requirements
```

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -472,6 +472,11 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
...
+	if cfg.Spec.DaemonResourceRequirements != nil {
+		templateSpec.Containers[bindata.ContainerIDDaemon].Resources = *cfg.Spec.DaemonResourceRequirements
+	}
```
pjbgf, line 478: I think we could extend this to the `initContainers` as well. wdyt?

api/spod/v1alpha1/spod_types.go
```diff
@@ -106,6 +106,11 @@ type SPODSpec struct {
 	// namespace to use for pulling the images from SPOD pod from a private registry.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// DaemonResourceRequirements if defined, overwrites the default resoruce requirements
```
pjbgf, line 110:
```suggestion
	// DaemonResourceRequirements if defined, overwrites the default resource requirements.
```

pjbgf (commented):
A few small nits, but overall lgtm from me. Thank you for working on this @ccojocar.

internal/pkg/manager/spod/spod_controller.go
ccojocar: fixed

internal/pkg/manager/spod/spod_controller.go
ccojocar: Do you mean the same resources?

ccojocar:
> It would be nice to document the new options. I'm on the fence about a test, ideally we should have everything tested (and e.g. we do have tests for webhook label changes), but the patch "just" adds a public kube type, so..shrug.


I'll add some docs and tests.

internal/pkg/manager/spod/spod_controller.go
ccojocar: I think the two init containers don't have really performance issues, or need more resources. I would let them as they are for now.

ccojocar:
@jhrozek @pjbgf I think that I addressed all your remarks. I added some documentation and an e2e test.

api/spod/v1alpha1/spod_types.go
```diff
@@ -308,6 +308,11 @@ type SPODSpec struct {
 	// namespace to use for pulling the images from SPOD pod from a private registry.
 	// +optional
 	ImagePullSecrets []corev1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
+
+	// DaemonResourceRequirements if defined, overwrites the default resource requirements
+	// of SPOD daemon.
+	// +optinal
```
JAORMX, line 314: `s/optinal/optional/`

JAORMX (changes requested):

internal/pkg/manager/spod/spod_controller.go
pjbgf: The idea was mostly to provide users a way to set resource requests/limits across all containers we expose. But I agree that our current init containers are less resource intensive, so happy to not expand this for the time being. :+1:

installation-usage.md
```diff
@@ -231,6 +232,16 @@ The `ds/spod` should now be updated by the manager with the new SELinux type, an
             type: unconfined_t
 ```
 
+## Customise the daemon resource requirements
+
+The default resource requirements of the daemon container can be adjusted in the SPOD configuration in the
+`daemonResourceRequirements` filed as follows:
```
pjbgf, line 238: nit:
```suggestion
The default resource requirements of the daemon container can be adjusted by using the field `daemonResourceRequirements`
from the SPOD configuration as follows:
```

saschagrunert (approved):

PR #1295, opened 2022-11-11
Add direclty the file header of generated mocks

ccojocar:
#### What type of PR is this?

/kind cleanup

Adds directly the file header of the generated mocks instead of doing post processing which cased duplicated header on macos.

This allow to run the `make verify` command on macos during development. The current implementation generates a duplicated header which causes a dirty tree.

```release-note
Add directly the file header when generating the mock types.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ issues:
       linters:
         - dupl
         - gofumpt
+        - lll
     - linters:
         - goerr113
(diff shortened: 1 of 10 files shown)


JAORMX (approved):

ccojocar:
@JAORMX Please could you lgtm again this pull request? Thanks

PR #1296, opened 2022-11-11
Fix flaky unit test which check the default operator namespace

ccojocar:
#### What type of PR is this?

/kind failing-test 

This unit test cannot run in parallel because environment variables are global causing random failures. This happens because one test case invalidates the other.

```release-note
Fix flaky unit test which checks default operator namespace.
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -15,6 +15,9 @@ issues:
     - linters:
         - lll
       source: "^//go:generate " # exclude lll issues for long lines with go:generate
+    - linters:
+        - paralleltest
+      path: internal/pkg/config/config_test.go
 linters:
   disable-all: true
   enable:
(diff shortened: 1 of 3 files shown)


pjbgf:
We will need to set lint ignore to get linter happy on the below:
```
Error: internal/pkg/config/config_test.go:23:1: Function TestGetOperatorNamespace missing the call to method parallel (paralleltest)
func TestGetOperatorNamespace(t *testing.T) {
^
Error: internal/pkg/config/config_test.go:42:2: Range statement for test TestGetOperatorNamespace missing the call to method parallel in test Run (paralleltest)
	for _, tt := range tests {
	^
```

JAORMX (approved):
/lgtm

saschagrunert (approved):

saschagrunert (approved):

PR #1306, opened 2022-11-16
make: Add an option to produce a dynamic build

jhrozek:
#### What type of PR is this?
/kind feature
(I guess?)

OCP does not compile the binaries it runs with upstream go, but rather a
fork of go that uses OpenSSL for cryptography:
    https://github.com/golang-fips

This fork uses `dlopen()` on startup to load symbols that correspond to
the crypto functions:
    https://github.com/golang-fips/openssl-fips/blob/b9297ed11fc8890f74822482c46f4c9cb5507251/openssl/openssl.go#L53
    https://github.com/golang-fips/openssl-fips/blob/eeda1baae76c92112220de78c5a498a6ba0abb6b/openssl/goopenssl.h#L66-L72

For some reason, this dlopen initialization does to work well for SPO on
OpenShift clusters running in FIPS mode where the dlopen call just
segfaults:
```
(gdb) frame 0
"libcrypto.so.1.1") at dlopen.c:78
78          return _dlfcn_hook->dlopen (file, mode, DL_CALLER);
(gdb) p _dlfcn_hook
$1 = (struct dlfcn_hook *) 0x0
(gdb) list
73      void *
74      __dlopen (const char *file, int mode DL_CALLER_DECL)
75      {
76      # ifdef SHARED
77        if (!rtld_active ())
78          return _dlfcn_hook->dlopen (file, mode, DL_CALLER);
79      # endif
80
81        struct dlopen_args args;
82        args.file = file;
```

To work around this issue, this patch adds the option to compile SPO as
a dynamically linked executable.

Please see https://issues.redhat.com/browse/OCPBUGS-3431 for more
details.

None

#### Does this PR have test?
No, but I did test this PR on an OCP 4.12 cluster with FIPS enabled.

N/A

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,11 +91,18 @@ GO_PROJECT := sigs.k8s.io/$(PROJECT)
 LDVARS := \
 	-X $(GO_PROJECT)/internal/pkg/version.buildDate=$(BUILD_DATE) \
 	-X $(GO_PROJECT)/internal/pkg/version.version=$(VERSION)
+STATIC_LINK ?= yes
+ifeq ($(STATIC_LINK), yes)
+  EXTLDFLAGS := -extldflags "-static"
+else
+  EXTLDFLAGS :=
+endif
+
 LINKMODE_EXTERNAL ?= yes
 ifeq ($(LINKMODE_EXTERNAL), yes)
-  LDFLAGS := -s -w -linkmode external -extldflags "-static" $(LDVARS)
+  LDFLAGS := -s -w -linkmode external $(EXTLDFLAGS) $(LDVARS)
 else
-  LDFLAGS := -s -w -extldflags "-static" $(LDVARS)
+  LDFLAGS := -s -w $(EXTLDFLAGS) $(LDVARS)
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -91,11 +91,18 @@ GO_PROJECT := sigs.k8s.io/$(PROJECT)
 LDVARS := \
 	-X $(GO_PROJECT)/internal/pkg/version.buildDate=$(BUILD_DATE) \
 	-X $(GO_PROJECT)/internal/pkg/version.version=$(VERSION)
+STATIC_LINK ?= yes
```
JAORMX, line 94: Why not default this to true when building the UBI container?

JAORMX (changes requested):

Makefile
jhrozek: The default is true (same as it is now), did you mean false?

I /thought/ that UBI didn't have the required devel packages, but it seems they do now. I'll try it, thanks!

jhrozek: OK, turns out that it was the static packages that were missing, not devel packages. I was confused!

JAORMX (approved):

PR #1310, opened 2022-11-17
Prefix with localhost the local seccomp profile for cri-o only for older Kubernetes versions

ccojocar:
#### What type of PR is this?

/kind bug

This is a follow up on #1255. See https://github.com/kubernetes-sigs/security-profiles-operator/pull/1255#issuecomment-1318369917

Prefix with localhost the local seccomp profile for cri-o only for older Kubernetes versions.

```release-note
Prefix with localhost the local seccomp profile for cri-o only for older Kubernetes versions.
```


ccojocar:
/assign @saschagrunert 

It would be nice if this can be part of `v0.5.1` release. Thanks

saschagrunert (approved):

PR #1311, opened 2022-11-17
Use the PID when looking up the command in eBPF map

ccojocar:

```release-note
Use the PID when looking up the command in eBPF map.
```


pjbgf:
This is an informal `/lgtm` from me. I think this will make the recording more precise, as it will ignore duplicate command names from other workloads that are not being recorded.

Either way, it would be good to get @saschagrunert :eyes: on it, as he's been closer to the eBPF recording feature.

saschagrunert:
@ccojocar may I ask you to rebase this one again?

ccojocar:
@saschagrunert I rebased. Thanks for looking at this.

saschagrunert (approved):

PR #1326, opened 2022-11-23
Add `replaces` and `containerImage` to operator manifest

saschagrunert:
#### What type of PR is this?

/kind documentation

We now also update the release process to automatically adjust those versions.

None
#### Does this PR have test?
None

None

```release-note
Automatically adding `replaces` and `containerImage` to OperatorHub manifest bundle.
```

### DIFF
--- a/bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
@@ -98,6 +98,7 @@ metadata:
       ]
     capabilities: Basic Install
     categories: Security
+    containerImage: registry.k8s.io/security-profiles-operator/security-profiles-operator:v0.6.0
     olm.skipRange: '>=0.4.1 <0.6.1-dev'
     operatorframework.io/suggested-namespace: security-profiles-operator
     operators.openshift.io/valid-subscription: '["OpenShift Kubernetes Engine", "OpenShift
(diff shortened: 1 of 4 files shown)


bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
```diff
@@ -98,6 +98,7 @@ metadata:
       ]
     capabilities: Basic Install
     categories: Security
+    containerImage: registry.k8s.io/security-profiles-operator/security-profiles-operator:v0.6.0
```
jhrozek, line 101: hmm, shouldn't this point at the `:latest` image we build in CI in between release and only point to the versioned image at the time of a release?

saschagrunert: I thought we're using the local image exclusively in the OLM tests, right?

jhrozek: Yes, we do.
Sorry, I'm confused about this change (this hunk, the `replaces` part is fine). Why are we adding the `containerImage` directive? I couldn't find much about it in OLM docs..

jhrozek: Ah, now I remember!
Thanks!

jhrozek (approved):
/lgtm

PR #1328, opened 2022-11-23
examples: The NS must allow privileged PSS

jhrozek:
#### What type of PR is this?
/kind cleanup

These are the same labels we have in our usual manifests, let's include
them also in the OLM examples.

None

#### Does this PR have test?
N/A

N/A

```release-note
NONE
```

### DIFF
--- a/examples/olm/install-resources.yaml
+++ b/examples/olm/install-resources.yaml
@@ -5,6 +5,9 @@ metadata:
   name: security-profiles-operator
   labels:
     openshift.io/cluster-monitoring: "true"
+    pod-security.kubernetes.io/enforce: privileged
+    pod-security.kubernetes.io/audit: privileged
+    pod-security.kubernetes.io/warn: privileged
 ---
 apiVersion: operators.coreos.com/v1alpha1
 kind: CatalogSource


jhrozek:
hmm let me
/hold
this

Since this is OLM example, it appears that OLM is smart enough to label the namespace correctly once a CSV appears in the namespace. At least on OCP (I need to confirm this on another distro I guess)

jhrozek:
/hold cancel

this appears to be needed outside OCP.
Reading https://master.sdk.operatorframework.io/docs/best-practices/pod-security-standards/ I see:
```
For workloads that need elevated permissions: it is recommended that you ensure the namespace containing your solution is labeled accordingly. You can either update your operator to manage the namespace labels or include the namespace labeling as part of the manual install instructions.
```

I saw the labels added automatically on OCP, because the operator's RBAC includes the use of the `privileged` SCC in comments:
```
git grep resources=securitycontextconstraints
internal/pkg/daemon/seccompprofile/seccompprofile.go:// +kubebuilder:rbac:groups=security.openshift.io,namespace="security-profiles-operator",resources=securitycontextconstraints,verbs=use
internal/pkg/manager/spod/spod_controller.go:// +kubebuilder:rbac:groups=security.openshift.io,namespace="security-profiles-operator",resources=securitycontextconstraints,verbs=use
internal/pkg/webhooks/binding/binding.go:// +kubebuilder:rbac:groups=security.openshift.io,namespace=security-profiles-operator,resources=securitycontextconstraints,verbs=use
```
and OCP has a component called `LabelSyncer` that automatically labels namespaces that contain a CSV if the RBAC rules in that namespace make use of the `privileged` SCC.

PR #1336, opened 2022-11-28
spod: Enabling log enricher via env var should enable the profilerecorder, too

jhrozek:
#### What type of PR is this?
/kind bug

Most of the time when our users enable the log enricher, it is to record
workloads and the profilerecorder controller's `collectLogProfiles()`
understandably doesn't work without log enricher.

However, when the log enricher was enabled through an environment
variable, the variable was not passed down from the controller to the
spod and the profilerecorder did not check the presence of the variable,
only the spod.spec.enableLogEnricher field.

This patch passes the variable down to spod from the manager and makes
the profilerecorder check both the attribute and the env var.

None

#### Does this PR have test?
Not yet. I'm working on an (unrelated) OLM patch set that will enable
testing this use-case.

N/A

```release-note
NONE
```


saschagrunert (commented):
Can we document that behavior?

jhrozek:
> Can we document that behavior?

Done, thanks!
I also added an env var for the BPF recorder enablement.

btw maybe we should 
/hold
this and merge #1337 first so that I can write proper test (it would be easier to test everything with OLM installations than in the go-based e2e tests)

saschagrunert (approved):

PR #1337, opened 2022-11-28
bundle: Suport OLM installation modes properly

jhrozek:
#### What type of PR is this?
/kind bug

Please see the commit mmessages for full information about each change.

But the short version is that our OLM packaging was done incorrectly
(and I wrote it..sorry...). The installation mode was ignored unless
the `RESTRICT_TO_NAMESPACE` variable was set and even then installing
SPO using OLM to another namespace than the default one didn't work.

The patchset uses the `olm.targetNamespace` variable set by OLM as
value of the `WATCH_NAMESPACES` variable, but to keep backwards compatibility,
`RESTRICT_TO_NAMESPACE` still takes precedence.

Next, all installation methods that OLM supports are enabled in the bundle.

OLM also manages Service Accounts when SPO is installed through OLM to allow
for installing into non-default namespaces.

Our examples suggest to use the `AllNamespaces` mode as the default, because
by default we want to watch all namespaces but even more importantly,
unless you want to watch a single namespace only, SPO is not multi-tenant
and installing into all namespaces would ensure that only one CSV exists in
the cluster.

Finally, the OLM tests have been expanded quite a bit to cover all different
installation methods.

None

#### Does this PR have test?
Yes

N/A

```release-note
When using OLM to install the SPO from a bundle, SPO now defaults to installation in all namespaces and watching CRs across all namespaces. Please refer to https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/ to learn how to scope the operator to either watch only a subset of namespaces or install SPO to a different namespace when using OLM.

Note that the other installation methods or the RESTRICT_TO_NAMESPACE environment variables are not affected by this change and work as before.
```

### DIFF
--- a/examples/olm/install-resources.yaml
+++ b/examples/olm/install-resources.yaml
@@ -24,9 +24,6 @@ kind: OperatorGroup
 metadata:
   name: security-profiles-operator
   namespace: security-profiles-operator
-spec:
-  targetNamespaces:
-  - security-profiles-operator
 ---
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
--- a/examples/olm/install-resources.yaml
+++ b/examples/olm/install-resources.yaml
@@ -19,12 +19,6 @@ spec:
   sourceType: grpc
   image: gcr.io/k8s-staging-sp-operator/security-profiles-operator-catalog:latest
 ---
-apiVersion: operators.coreos.com/v1
-kind: OperatorGroup
-metadata:
-  name: security-profiles-operator
-  namespace: security-profiles-operator
----
 apiVersion: operators.coreos.com/v1alpha1
 kind: Subscription
 metadata:
@@ -36,3 +30,9 @@ spec:
   source: security-profiles-operator
   #sourceNamespace: openshift-marketplace on OCP
   sourceNamespace: olm
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: security-profiles-operator
+  namespace: security-profiles-operator
--- a/hack/ci/e2e-olm.sh
+++ b/hack/ci/e2e-olm.sh
@@ -22,6 +22,19 @@ IMG=${REPO}/security-profiles-operator:${GITHUB_SHA}
 BUNDLE_IMG=${REPO}/security-profiles-operator-bundle:v${GITHUB_SHA}
 CATALOG_IMG=${REPO}/security-profiles-operator-catalog:v${GITHUB_SHA}
 
+function sp_in_ns() {
+    ns=$1
+kubectl create -f - << EOF
+apiVersion: security-profiles-operator.x-k8s.io/v1beta1
+kind: SeccompProfile
+metadata:
+  name: log-all
+  namespace: $ns
+spec:
+  defaultAction: "SCMP_ACT_LOG"
+EOF
+}
+
 function build_and_push_spo() {
     make image IMAGE=${IMG}
     podman push --tls-verify=false ${IMG}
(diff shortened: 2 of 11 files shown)


examples/olm/install-resources.yaml
```diff
@@ -36,3 +30,9 @@ spec:
   source: security-profiles-operator
   #sourceNamespace: openshift-marketplace on OCP
   sourceNamespace: olm
+---
+apiVersion: operators.coreos.com/v1
+kind: OperatorGroup
+metadata:
+  name: security-profiles-operator
+  namespace: security-profiles-operator
```
jhrozek, line 38: In case anyone is wondering, I just moved the OG to the end of the file so that it's easier to append to in the tests, as a byproduct we can keep using this example in the test which would hopefully also keep testing the example.

hack/ci/e2e-olm.sh
```diff
@@ -50,30 +63,112 @@ function build_and_push_packages() {
...
+    ns=$1
+
     # Useful in case the CatalogSource is fubar. We retry several times
     # because on transient errors (which are for some reason common even
     # if the catalog is local) the pod gets restarted
     for i in $(seq 1 5); do
         kubectl -nolm wait --for=condition=ready pods -lolm.catalogSource=security-profiles-operator
-        catalog_logs=$(kubectl -nolm logs $(kubectl -nolm get pods --no-headers -lolm.catalogSource=security-profiles-operator | awk '{print $1}') 2>/dev/null)
-        if [[ -n "$catalog_logs" ]]; then
+        install_rv=$?
```
jhrozek, line 166: This change is not strictly related to the tests, but I found (because I was running the tests over and over..) that the wait occasionally times out on the first try, but the `catalog_logs` would usually contain something, so I made the retry loop on `install_rv` instead.

hack/ci/e2e-olm.sh
```diff
@@ -82,30 +177,212 @@ function check_spo_is_running() {
     # (jhrozek): I didn't find a useful condition or status to wait for..
     # ..if only there was a way to check if ANY installedCSV is set..
     sleep 30
-    CSV=$(kubectl -nsecurity-profiles-operator get sub security-profiles-operator-sub -ojsonpath='{.status.installedCSV}')
+    CSV=$(kubectl -n$ns get sub security-profiles-operator-sub -ojsonpath='{.status.installedCSV}')
     # wait for the CSV to be actually installed
-    kubectl -nsecurity-profiles-operator wait --for=jsonpath='{.status.phase}'=Succeeded csv $CSV
+    kubectl -n$ns wait --for=jsonpath='{.status.phase}'=Succeeded csv $CSV
```
jhrozek, line 182: `$ns` instead if `security-profiles-operator` to support the custom namespace test.

hack/ci/e2e-olm.sh
```diff
@@ -82,30 +177,212 @@ function check_spo_is_running() {
...
-    kubectl -nsecurity-profiles-operator wait --for=condition=ready pod -lname=spod
+    kubectl -n$ns wait --for=condition=ready pod -lname=spod || return 1
+
+    return 0
+}
+
+function assert_spo_csv_installed_in_ns() {
+    ns=$1
+
+    [[ $(kubectl get csv -loperators.coreos.com/security-profiles-operator.$ns= -n$ns -oname) ]] || return 1
```
jhrozek, line 217: this probably looks hacky, but I found it easier to list any CSV in the NS that installs SPO rather than parse out the version of the CSV directly..
the `-loperators.coreos.com/security-profiles-operator.$ns=` is an OLM-ism, the label is `-loperators.coreos.com/$NAME_OF_THE_OPERATOR.$NAMESPACE_WHERE_THE_OPERATOR_IS_INSTALLED=`

hack/ci/e2e-olm.sh
```diff
@@ -82,30 +177,212 @@ function check_spo_is_running() {
...
+    smoke_test $method || rv=1
+
+    teardown_spo $method
+done
+
+# deployment into the custom namespace is a bit special
+echo "Testing SPO deployment in custom namespace(s)"
+deploy_spo_in_custom_ns spo-lives-here || rv=1
+check_spo_is_running spo-lives-here || rv=1
+kubectl get csv -A --show-labels
```
jhrozek, line 384: hmm, sorry, this is not needed and is just debugging, I can either remove this command if there's comment for this PR or when I submit a patch for the environment variable installations.

saschagrunert (approved):

hack/ci/e2e-olm.sh
saschagrunert: Let's keep it if it helps debugging.

PR #1339, opened 2022-11-29
 Configure the Selinux type tag when the AppArmor is disabled regardless of EnableSelinux flag

ccojocar:
- Configure the Selinux type tag when the AppArmor is disabled regardless of EnableSelinux flag
- Use the proper init container ID to avoid confusion

#### What type of PR is this?

/kind bug 

This introduced a regression on Flatcar Linux. The SELinux type tag needs to be configured independent of EnableSelinux flag, because the SELinux can be active on the node regardless if the SELinux feature is enabled or not in the operator.
For instance, on Flatcar Linux SELinux type tag needs to be set to 'unconfined_t' instead of 'spc_t'
even though SELinux is disabled in order to get the containers to start.

```release-note
Configure the Selinux type tag when the AppArmor is disabled regardless of EnableSelinux flag.
```

### DIFF
--- a/internal/pkg/manager/spod/spod_controller.go
+++ b/internal/pkg/manager/spod/spod_controller.go
@@ -592,8 +592,12 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 		// Set the logging verbosity
 		templateSpec.InitContainers[i].Env = append(templateSpec.InitContainers[i].Env, verbosityEnv(cfg.Spec.Verbosity))
 
-		// Update the SELinux type tag when is defined in the configuration
-		if cfg.Spec.EnableSelinux != nil && *cfg.Spec.EnableSelinux {
+		// Update the SELinux type tag only when AppArmor is not enabled this is to prevent a crash.
+		// The SELinux type tag needs to be configured independent of EnableSelinux flag, because the
+		// SELinux can be active on the node regardless if the SELinux feature is enabled or not in the operator.
+		// For instance, on Flatcar Linux SELinux type tag needs to be set to 'unconfined_t' instead of 'spc_t'
+		// even though SELinux is disabled in order to get the containers to start.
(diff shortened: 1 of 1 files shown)


ccojocar:
@pjbgf please could you also have a look since was introduced in https://github.com/kubernetes-sigs/security-profiles-operator/commit/025c4d1722a4069531909cb3fdbdbbda41b285b9.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -592,8 +592,12 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
...
 		templateSpec.InitContainers[i].Env = append(templateSpec.InitContainers[i].Env, verbosityEnv(cfg.Spec.Verbosity))
 
-		// Update the SELinux type tag when is defined in the configuration
-		if cfg.Spec.EnableSelinux != nil && *cfg.Spec.EnableSelinux {
+		// Update the SELinux type tag only when AppArmor is not enabled this is to prevent a crash.
+		// The SELinux type tag needs to be configured independent of EnableSelinux flag, because the
+		// SELinux can be active on the node regardless if the SELinux feature is enabled or not in the operator.
+		// For instance, on Flatcar Linux SELinux type tag needs to be set to 'unconfined_t' instead of 'spc_t'
+		// even though SELinux is disabled in order to get the containers to start.
+		if !cfg.Spec.EnableAppArmor {
```
pjbgf, line 600: `EnableAppArmor` is a bool pointer, when not specified this will cause a panic.

pjbgf (approved):
/lgtm

pjbgf:
The failure is orthogonal, and is highlighting we need to bump the ubi-image.

pjbgf:
I am overriding the `ubi-image` failure and merging this. We need to handle the ubi-image fix as a separate PR.

PR #1342, opened 2022-11-30
Refactor the workloadannotator to reduce the code duplication

### LINKED ISSUE #887: Reduce code duplication in workloadannotator
We added support for SELinux profile binding:https://github.com/kubernetes-sigs/security-profiles-operator/pull/854

However, there are some code duplications in
 https://github.com/kubernetes-sigs/security-profiles-operator/blob/de3017ad869867efb9743a1c387398e7741fd3cc/internal/pkg/manager/workloadannotator/workloadannotator.go#L200

Since we will also add profile binding in the future for AppArmor. Under the current method, we will have more duplication.
we should create an interface that all security profiles that support bindings would implement and then pass a variable implementing that interface to the update function.

ccojocar:
#### What type of PR is this?

/kind cleanup

Refactor workloadannotator to reduce code duplication.

Fixes #887

```release-note
Refactor workloadannotator to reduce code duplication.

```


ccojocar:
This refactoring seems to introduce some issue which I can't figure it out. Closing it for now.

PR #1343, opened 2022-12-01
profilerecorder: Don't issue seccomp events for recording of any profile

jhrozek:
#### What type of PR is this?
/kind cleanup

We would issue an event with Seccomp in the event name even when
recording SELinux. Let's use profile-agnostic event names instead.

The alternative would be to have a different even for each profile type. I'm
not sure myself which is better.

None

#### Does this PR have test?
No

N/A

```release-note
NONE
```


saschagrunert (approved):

saschagrunert:
Let's add a release note for that.

jhrozek:
> Let's add a release note for that.

Sure, done.

PR #1347, opened 2022-12-02
OLM: Add separators between CR examples to make sure all CRDs have an example in the bundle

### LINKED ISSUE #885: Not all CRs have an example annotation in the bundle
#### What happened:
I noticed this while working on the operatorhub submission: Not all CRs have an example annotation defined in the CSV. This will get also caught in scorecard tests, although I don't think it's such a huge deal.

To make sure the CSV is well-formed, all CRs should have an example annotation

`less bundle/manifests/security-profiles-operator.clusterserviceversion.yaml` then look for the `alm-examples`

#### Environment:

- Kernel (e.g. `uname -a`):
- Others:

jhrozek:
#### What type of PR is this?
/kind cleanup

The bundle normally contains an annotation called `alm-examples` that is
displayed in UIs to guide users in creating CRs. Some of our CRs did not
have an example, because there were no separators between the examples
and OLM had trouble parsing them.

In addition, adds a separator between the examples and the deployment
manifest.

Fixes #885

#### Does this PR have test?
Will be tested by the OLM tests.

see inline

```release-note
NONE
```


Makefile
```diff
@@ -479,15 +479,16 @@ BUNDLE_IMG ?= $(PROJECT)-bundle:v$(VERSION)
...
 OLM_EXAMPLES := \
 	examples/apparmorprofile.yaml \
 	examples/config.yaml \
 	examples/profilerecording-seccomp-bpf.yaml \
 	examples/profilebinding.yaml \
 	examples/rawselinuxprofile.yaml \
 	examples/seccompprofile.yaml \
-	examples/selinuxprofile.yaml
+	examples/selinuxprofile.yaml \
+	deploy/separator.yaml
```
jhrozek, line 491: so this is clearly a hack, but I didn't figure out how could I ask `kustomize` to produce `operator.yaml` with separators on the first line.

Anyone has an idea?

saschagrunert: I don't let's
/hold 
and feel free to unhold if there is no other solution in sight.

saschagrunert (approved):

PR #1353, opened 2022-12-07
Parse ebpf event in one step

saschagrunert:
#### What type of PR is this?

/kind cleanup

We now make ebpf events 4 bytes smaller and parse them in one rush.

None
#### Does this PR have test?
None

None

```release-note
None
```


JAORMX (commented):
Shouldn't we instead be trying to derive the number to use from `/proc/sys/kernel/pid_max`? Seems to me that using u32 might be error prone 😕

saschagrunert:
> Shouldn't we instead be trying to derive the number to use from `/proc/sys/kernel/pid_max`? Seems to me that using u32 might be error prone confused

The maximum amount of pids on 64 bit systems is `2^22`, which cannot be larger than `2^32-1` (`u32` max).

JAORMX:
> > Shouldn't we instead be trying to derive the number to use from `/proc/sys/kernel/pid_max`? Seems to me that using u32 might be error prone confused
> 
> The maximum amount of pids on 64 bit systems is `2^22`, which cannot be larger than `2^32-1` (`u32` max).

That makes sense. Thanks for replying!

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -38,7 +38,7 @@ struct {
 } events SEC(".maps");
 
 struct event_t {
-    u64 pid;
+    u32 pid;
```
jhrozek, line 41: IIRC `pid_t` is a 32bit type so this change is correct, but I wonder (just for my curiosity, really), where does `u32` come from? It took me a bit to realize I'm not looking at `uint32_t` and I was wondering why we were not using `pid_t` in the first place here.

Is the type a BPF type?

saschagrunert: Yes, `vmlinux.h` provides typedefs for them: https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/internal/pkg/daemon/bpfrecorder/vmlinux/x86/6.0.7-301.fc37.x86_64.vmlinux.h

jhrozek (approved):
/lgtm

PR #1371, opened 2022-12-12
manager: Own the ServiceMonitor

jhrozek:
#### What type of PR is this?
/kind bug

The ServiceMonitor is created by the main operator deployment, but in
case the operator is uninstalled the service monitor is still around
which might trigger alerts esp in case the DS is removed.

Note: I'm unsure about the approach. I'll point some questions inline.

None

#### Does this PR have test?
Yes

see inline

```release-note
NONE
```

### DIFF
--- a/internal/pkg/manager/spod/spod_controller.go
+++ b/internal/pkg/manager/spod/spod_controller.go
@@ -101,7 +101,7 @@ func (r *ReconcileSPOd) Healthz(*http.Request) error {
 // Operand:
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets,verbs=get;list;watch;create;update;patch
-// +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets/finalizers,verbs=delete;get;update;patch
+// +kubebuilder:rbac:groups=apps,resources=deployments/finalizers;daemonsets/finalizers,verbs=delete;get;update;patch
 // +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=cert-manager.io,resources=issuers;certificates,verbs=get;list;watch;create;update;patch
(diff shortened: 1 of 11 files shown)


internal/pkg/manager/spod/spod_controller.go
```diff
@@ -101,7 +101,7 @@ func (r *ReconcileSPOd) Healthz(*http.Request) error {
 // Operand:
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets,verbs=get;list;watch;create;update;patch
-// +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets/finalizers,verbs=delete;get;update;patch
+// +kubebuilder:rbac:groups=apps,resources=deployments/finalizers;daemonsets/finalizers,verbs=delete;get;update;patch
```
jhrozek, line 104: I think this was a bug in the first place, I don't think we need to delete deploys at the moment and the other verbs were already pretty much in the line above

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -172,7 +172,10 @@ func (r *ReconcileSPOd) Reconcile(_ context.Context, req reconcile.Request) (rec
 	webhook := bindata.GetWebhook(r.log, r.namespace, spod.Spec.WebhookOpts, image,
 		pullPolicy, caInjectType, spod.Spec.Tolerations, spod.Spec.ImagePullSecrets)
 	metricsService := bindata.GetMetricsService(r.namespace, caInjectType)
-	serviceMonitor := bindata.ServiceMonitor(caInjectType)
+	serviceMonitor, err := bindata.ServiceMonitor(foundDeployment, r.scheme, caInjectType)
```
jhrozek, line 175: Here's my main question: Does this ownership structure make sense? Or should we maybe make the default spod owned by the operator and the DS and the serviceMonitor both by the spod? (Or spod->ds->serviceMonitor) ?

JAORMX: If we're gonna have a service monitor for a DS, and the DS is created via the SPOd objects we might want to consider dedicating a service monitor for the DS itself instead of having a deployment+ds bundle as we do today 😕 

The main question is, do we want to support multiple spod deployments at all?

PR #1378, opened 2022-12-13
profilerecorder: gracefully skip containers that did not record anything

jhrozek:
#### What type of PR is this?
/kind bug

This is mostly concerning SELinux because even containers that don't
seemingly do anything at least record some syscalls, e.g. exec.

But with SELinux, if a container only performs actions that are allowed
by the container policy in the first place, no AVCs would be generated.
In that case, the call to get AVCs would return with a rather generic
error, which would then be retried by the profilerecording reconciler.
As a consequence, if a pod had multiple containers and the one with no
AVCs was collected before others, the profilerecorder would never reach
the other containers.

Instead, let's return an error matching the situation and handle the
absence of AVCs gracefully, by emptying whatever there might be in the
grpc cache and moving on to the next container.

None

#### Does this PR have test?
Yes

N/A

```release-note
NONE
```

### DIFF
--- a/internal/pkg/daemon/bpfrecorder/bpfrecorder.go
+++ b/internal/pkg/daemon/bpfrecorder/bpfrecorder.go
@@ -704,8 +704,10 @@ func (b *BpfRecorder) findProfileForContainerID(id string) (string, error) {
 
 			for p := range pods.Items {
 				pod := &pods.Items[p]
-				for c := range pod.Status.ContainerStatuses {
-					containerStatus := pod.Status.ContainerStatuses[c]
+				//nolint:gocritic // We explicitly do not want to append to the same slice
+				statuses := append(pod.Status.InitContainerStatuses, pod.Status.ContainerStatuses...)
+				for c := range statuses {
+					containerStatus := statuses[c]
 					fullContainerID := containerStatus.ContainerID
 					containerName := containerStatus.Name
 
--- a/internal/pkg/daemon/enricher/grpc.go
+++ b/internal/pkg/daemon/enricher/grpc.go
@@ -22,10 +22,19 @@ import (
 	"fmt"
 	"runtime"
 
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	api "sigs.k8s.io/security-profiles-operator/api/grpc/enricher"
+)
+
+const (
+	// ErrorNoSyscalls is returned when no syscalls are recorded for a profile.
+	ErrorNoSyscalls = "no syscalls recorded for profile"
+	// ErrorNoAvcs is returned when no AVCs are recorded for a profile.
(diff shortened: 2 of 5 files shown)


jhrozek:
Looks like the OLM tests are broken right now:
```
+ operator-sdk olm install --version v0.18.2 --timeout 6m
time="2022-12-13T14:07:45Z" level=info msg="Fetching CRDs for version \"v0.18.2\""
time="2022-12-13T14:07:45Z" level=info msg="Fetching resources for resolved version \"v0.18.2\""
2022/12/13 14:07:45 protocol error: received *http.http2GoAwayFrame before a SETTINGS frame
time="2022-12-13T14:07:45Z" level=fatal msg="Failed to install OLM version \"v0.18.2\": failed to get resources: failed to fetch CRDs: request failed: failed GET 'https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.18.2/crds.yaml': Get \"[https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.18.2/crds.yaml\](https://github.com/operator-framework/operator-lifecycle-manager/releases/download/v0.18.2/crds.yaml/)": connection error: PROTOCOL_ERROR"
```

jhrozek:
olm tests and the regular e2e tests now passed, but the fedora e2e failed twice with the same symptom. Going to look into that.
I could swear that the tests passed for me locally earlier in vagrant..

ccojocar (approved):

jhrozek:
Rebased atop current main

ccojocar (approved):

jhrozek:
Rebased atop the flatcar test fixes again

jhrozek:
hmm please hold off with a follow up review before I triage the test failures..

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
```diff
@@ -721,7 +721,9 @@ func (b *BpfRecorder) findContainerID(id string) error {
 
 			for p := range pods.Items {
 				pod := &pods.Items[p]
-				for c := range pod.Status.ContainerStatuses {
+				//nolint:gocritic // We explicitly do not want to append to the same slice
+				statuses := append(pod.Status.ContainerStatuses, pod.Status.InitContainerStatuses...)
+				for c := range statuses {
```
jhrozek, line 726: FWIW, this was the isse, we generated longer statuses and iterated over them, but inside the for loop we would have still assigned `containerStatus := pod.Status.ContainerStatuses[c]`.

jhrozek: The upcoming patch version changes to `containerStatus := statuses[c]`

jhrozek:
@ccojocar Hi, I'm sorry for the super long delay in getting back to this PR. I got pulled into other directions at work. Could I please get another `lgtm`? I added the reason the patch was initally failing into the PR as a comment, because of course you can't remember the PR anymore :-)
There are no other changes.

ccojocar (approved):

PR #1387, opened 2022-12-19
Exposing spo image repo & tag in helm charts to make it configurable

### LINKED ISSUE #1334: Helm: Make all images configurable values

Add configuration values to the SPO Helm Chart to configure all deployable images.

#### Why is this needed:

This enables users to have better control over what images/tags they deploy when using the official SPO helm chart, instead of having to create their own. 

#### User story covered

As an infrastructure SRE, Kirsti would like to deploy the SPO using the official Helm chart, whilst still having the power to define what version of each image to be deployed.

rahulroshan-kachchap:
#### What type of PR is this?
kind feature

This enables users to have better control over what images/tags they deploy when using the official SPO helm chart, instead of having to create their own.

Fixes #1334 

#### Does this PR have test?
Tested the above changes with spoimage repository set to gcr.io/k8s-staging-sp-operator/security-profiles-operator and spoimage tag set to latest.
Image pulled successfully. 

```
Added repository & tag in values.yaml to make SPO repo & image values configurable in helm charts.
```


ccojocar (approved):

rahulroshan-kachchap:
Hi @pjbgf , what is the expected date for approving this PR?

deploy/helm/values.yaml
```diff
@@ -1,6 +1,11 @@
 # Default values for security-profiles-operator.
 replicaCount: 3
 
+
+spoimage:
```
JAORMX, line 5: let's use camel-case for this. e.g. we should use `spoImage` instead to be consistent with other keys.

deploy/helm/values.yaml
```diff
@@ -1,6 +1,11 @@
...
+  repository: gcr.io/k8s-staging-sp-operator/security-profiles-operator
```
JAORMX, line 6: In order to better integrate with automatic dependency update tools, I'd suggest splitting the repository and registry keys as shown in the following example: https://docs.renovatebot.com/modules/manager/helm-values/

This would look as follows:

```yaml
spoImage:
  registry: gcr.io
  repository: k8s-staging-sp-operator/security-profiles-operator
  tag: latest
```

In case someone would set up an explicit image tag in their values file, this would then be automatically picked up and receive updates. It's also consistent to what other helm charts are doing in general.

JAORMX (changes requested):
Thanks a lot for your contribution! I apologize for taking long to review this. I have a couple of comments on this, and after they're addressed we're ready to merge! I don't have a strong opinion on my comment about splitting the registry and repository keys, but I would sure appreciate changing the main key to use camel case. Thanks again!

rahulroshan-kachchap:
Raising a separate PR , messed with commits here.

PR #1388, opened 2022-12-19
Install the default log enricher profile only when log enricher is enabled

ccojocar:
#### What type of PR is this?

/kind bug

Install the default log enricher seccomp profile only when log enricher is enabled in the spod configuration.

```release-note
Install the default log enricher sccomp profile only when log enricher is enabled in the spod configuration.

```


JAORMX (commented):
LGTM, I'll wait for CI to pass before merging. Thanks!

ccojocar:
Thanks @JAORMX ! It seems that all the tests passed. Please could you have a look over this, and approve it if everything is fine.

PR #1425, opened 2023-01-20
Add an option which enables the memory optimization inside of spod daemon

### LINKED ISSUE #1385: Memory requirement for profile recorder controller increases significantly with the cluster size
#### What happened:

It seems that the memory required by the profile recorder controller increases significantly with the cluster size. I had to configure 4096Mi for a cluster with 155 nodes and about 100+ pods on each node, in order to get the controller started, otherwise it gets stuck and the spo container is killed with OOM.

It is not clear to me, what's causing this issue but it seems to be deep down in the controller runtime even though there is a filter applied only for pods running on the node local to the daemon. 

https://github.com/kubernetes-sigs/security-profiles-operator/blob/9108c0ac8571a6cb2fb64dcf3055aeca8bb5acf7/internal/pkg/daemon/profilerecorder/profilerecorder.go#L156

The memory required by the profile recorder controller should not increased significantly with the cluster size.

Enable eBPF profile recorder when the spod is deployed in a large cluster(e.g. 100 nodes with more than 100 pods on each node).

Kubernetes Version: v1.22.10

#### Environment:

- Kernel (e.g. `uname -a`):
- Others:

### LINKED ISSUE #1386: spod crashes when profile recording in enabled in a large cluster
#### What happened:

security profiles operator is deployed in a fairly large cluster with 100+ nodes and 100+ pods per node. I had to increase the memory to 4096Mi in order to get the spod daemon started. It seems that even though there is enough memory the spod container crashes. See the attached log. 

[spod.txt](https://github.com/kubernetes-sigs/security-profiles-operator/files/10251054/spod.txt)

It's not clear to me what exactly is causing this crash. 

All containers of spod daemon should run without crashing.

Enable eBPF profile recorder in a large cluster (e.g. 100+ nodes, 100+ pods per node), and then increase the memory requirements of the deamonset such that it can start without being killed with OOM. After this, you should start seeing the 
spo container part of spod daemonset crashing. 

Kubernetes Version: v1.22.8

#### Environment:

- Kernel (e.g. `uname -a`):
- Others:

ccojocar:
#### What type of PR is this?

/kind bug

Adds an option to enable memory optimization inside of spod daemon.

The controller running inside of spod daemon process is watching all pods available in the cluster when profile recording
is enabled. It will perform some pre-filtering before the reconciliation to select only the pods running on local
node as well as pods annotated for recording, but this operation takes place after all pods objects are loaded
into the cache memory of the informer. This can lead to very high memory usage in large clusters with 1000s of pods, resulting in spod daemon running out of memory or crashing.

In order to prevent this situation, the spod daemon can be configured to only load into the cache memory the pods explicitly
labeled for profile recording.

Fixes #1385
Fixes #1386

#### Does this PR have test?

yes

```release-note
Add an option to enable memory optimization inside of spod daemon.
```

### DIFF
--- a/api/spod/v1alpha1/spod_types.go
+++ b/api/spod/v1alpha1/spod_types.go
@@ -258,6 +258,11 @@ type SPODSpec struct {
 	// EnableProfiling tells the operator whether or not to enable profiling
 	// support for this SPOD instance.
 	EnableProfiling bool `json:"enableProfiling,omitempty"`
+	// EnableMemoryOptimization enables memory optimization in the controller
+	// running inside of SPOD instance and watching for pods in the cluster.
+	// This will make the controller loading in the cache memory only the pods
+	// labled explicitly for profile recording with 'spo.x-k8s.io/enable-recording=true'.
+	EnableMemoryOptimization bool `json:"enableMemoryOptimization,omitempty"`
 	// tells the operator whether or not to enable SELinux support for this
 	// SPOD instance.
 	EnableSelinux *bool `json:"enableSelinux,omitempty"`
--- a/api/spod/v1alpha1/spod_types.go
+++ b/api/spod/v1alpha1/spod_types.go
@@ -261,7 +261,7 @@ type SPODSpec struct {
 	// EnableMemoryOptimization enables memory optimization in the controller
 	// running inside of SPOD instance and watching for pods in the cluster.
 	// This will make the controller loading in the cache memory only the pods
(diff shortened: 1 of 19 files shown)


JAORMX:
very nice! It looks good to me. I'll merge this once CI passes.

saschagrunert (approved):

saschagrunert:
Thank you!

PR #1438, opened 2023-01-27
Make the kubelet root directory configurable

ccojocar:
#### What type of PR is this?

/kind feature

This makes the kubelet root directory configurable. kubelet has a command option which allows to
configure its root directory. This is the place form where the configuration is loaded, including
the seccomp profiles. 

This is useful in a cluster which doesn't use the default `/var/lib/kubelet` path. 

#### Does this PR have test?

Yes

```release-note
Make the kubelet root directory configurable via KUBELET_DIR environment variable.
```

### DIFF
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -10,6 +10,7 @@
   - [Installation using OLM using upstream catalog and bundle](#installation-using-olm-using-upstream-catalog-and-bundle)
   - [Installation using helm](#installation-using-helm)
   - [Installation on AKS](#installation-on-aks)
+- [Configure a custom kubelet root directory](#configure-a-custom-kubelet-root-directory)
 - [Set logging verbosity](#set-logging-verbosity)
 - [Pull images from private registry](#pull-images-from-private-registry)
 - [Configure the SELinux type](#configure-the-selinux-type)
@@ -188,6 +189,12 @@ $ kubectl -nsecurity-profiles-operator get spod spod
 NAME   STATE
 spod   RUNNING
 ```
+
+## Configure a custom kubelet root directory
+
+You can configure a custom kubelet root directory in case your cluster is not using the default `/var/lib/kubelet` path.
+You can achieve this by setting the environment variable `KUBELET_DIR` in the operator deployment. This environment variable will
+be then set in the manager container as well as it will be propagated into the containers part of spod daemonset.
 
 ## Set logging verbosity
 
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -47,14 +47,6 @@ const (
 	// DefaultKubeletPath specifies the default kubelet path.
 	DefaultKubeletPath = "/var/lib/kubelet"
 
-	// KubeletSeccompRootPath specifies the path where all kubelet seccomp
-	// profiles are stored.
-	KubeletSeccompRootPath = "/var/lib/kubelet/seccomp"
-
-	// ProfilesRootPath specifies the path where the operator stores seccomp
-	// profiles.
-	ProfilesRootPath = KubeletSeccompRootPath + "/operator"
-
 	// NodeNameEnvKey is the default environment variable key for retrieving
 	// the name of the current node.
 	NodeNameEnvKey = "NODE_NAME"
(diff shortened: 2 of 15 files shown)


installation-usage.md
```diff
@@ -189,6 +190,12 @@ NAME   STATE
 spod   RUNNING
 ```
 
+## Configure a custom kubelet root directory
+
+You can configure a custom kubelet root directory in case your cluster is not using the default `/var/lib/kubelet` path.
+You can achieve this by setting the environment variable `KUBELET_DIR` in the operator deployment. This environment variable will
+be then set in the manager container as well as it will be propagated into the containers part of spod daemonset.
```
saschagrunert, line 197: Would it be better / possible to have this as part of the SPOD instance?

ccojocar: The challenge is that the kubelet path is required in the non-root enabler sidecar which currently doesn't read any SPOD CRD before copying the profiles into the kubelet directory. 

It seems to be used inside of nodestatus controller as well.

The idea is to have this environment variable propagated everywhere that all components used the same custom kubelet directory including both manager and daemon instances.

saschagrunert: Alright, make sense. Any further thoughts on that @kubernetes-sigs/security-profiles-operator-maintainers ?

internal/pkg/config/config.go
```diff
@@ -155,3 +153,25 @@ func TryToGetOperatorNamespace() (string, error) {
...
+	if kubeletDir == "" {
+		return DefaultKubeletPath
+	}
+	return kubeletDir
+}
+
+// KubeletSeccompRootPath specifies the path where all kubelet seccomp
+// profiles are stored.
+func KubeletSeccompRootPath() string {
+	return KubeletDir() + "/seccomp"
```
pjbgf, line 170:
```suggestion
	return path.Join(KubeletDir(), "seccomp")
```

internal/pkg/config/config.go
```diff
@@ -155,3 +153,25 @@ func TryToGetOperatorNamespace() (string, error) {
...
+}
+
+// ProfilesRootPath specifies the path where the operator stores seccomp
+// profiles.
+func ProfilesRootPath() string {
+	return KubeletSeccompRootPath() + "/operator"
```
pjbgf, line 176:
```suggestion
	return path.Join(KubeletSeccompRootPath(), "operator")
```

installation-usage.md
pjbgf: Relying on env-vars would make it a lot easier to enable support for non symmetrical configuration across node pools (e.g. node pools with different OS and configuration). +1 from me on the approach.

pjbgf (commented):
Two nits, apart from that it looks good to me.

internal/pkg/config/config.go
ccojocar: Fixed

ccojocar:
> Two nits, apart from that it looks good to me.

@pjbgf Thanks for review! I addressed your comments.

pjbgf (approved):
/lgtm

ccojocar:
@saschagrunert it seems that are some recent changes in the `llvm.sh` script which is causing the verify step to fail. The script is calling now the `sudo` command when is adding the llvm package key. This command doesn't seem to be available in the golang base image. The hack is to add the key before running the script in order to avoid running into sudo command later.

The fix is in this commit https://github.com/kubernetes-sigs/security-profiles-operator/pull/1438/commits/79329751d8cd03a171e2358ba52333b938b0c4cc

ccojocar:
It seems that they fixed the `llvm.sh` script in the meantime and the `sudo` was removed.

ccojocar:
The llvm.sh seems to be broken:
```
Hit:3 http://deb.debian.org/debian bullseye-updates InRelease
Get:4 https://apt.llvm.org/bullseye llvm-toolchain-bullseye-14 InRelease [6836 B]
Err:4 https://apt.llvm.org/bullseye llvm-toolchain-bullseye-14 InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 15CF4D18AF4F7421
Reading package lists...
W: GPG error: https://apt.llvm.org/bullseye llvm-toolchain-bullseye-14 InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 15CF4D18AF4F7421
E: The repository 'http://apt.llvm.org/bullseye llvm-toolchain-bullseye-14 InRelease' is not signed.
```

saschagrunert (approved):

PR #1453, opened 2023-02-03
Update golangci-lint version to v1.51.0

ccojocar:
#### What type of PR is this?

/kind bug

```release-note
Update golangci-lint to version 1.51.0.
```


ccojocar:
I am not sure if this is the only place where needs to be updated.

saschagrunert:
@ccojocar thanks! dependencies.yaml has to be updated for this as well.

ccojocar:
@saschagrunert I updated also in `dependencies.yaml`. It seems that locally passed. Let's see the build.

saschagrunert (approved):

PR #1476, opened 2023-02-13
Allow a custom kubelet directory for each node or pool of nodes

### LINKED ISSUE #1446: Configure kubelet directory path differently for each worker nodes pool

Add an option which allows to configure the kubelet directory differently for each worker nodes pool. 

A possible solution would be to request that all worker nodes with a different kubelet location are labeled for instance with `kubelet.kubernetes.io/directory-location` that has the kubelet directory path in the value. As an example for a node which is configured to use the kubelet directory path `/mnt/resource/kubelet`, the node should be labeled with:
```
kubelet.kubernetes.io/directory-location: mnt_resource_kubelet
```

The operator will list all nodes and identify all the distinct location labels. For each label, the operator will create a different daemonset object with a custom selector that match the value of the location label. It will also update the KUBELET_DIR environment variable of the daemonset to point to the same kubelet directory as the label value. 

The custom selector will ensure that the pods of each daemonset will be correctly scheduled on nodes that have the same kubelet directory as configured in their `KUBELET_DIR` environment variable.

#### Why is this needed:

The kubelet directory can be configured differently for each worker node and more typically for each worker nodes pool. The spo at the moment only allows to configure a custom kubelet directory for entire cluster. There are situations when the kubelet directory is different across the nodes in a cluster.

ccojocar:
#### What type of PR is this?

/kind feature

This PR allows to configure a custom kubelet root directory for each node or pool of nodes. This is done 
by applying the `kubelet.kubernetes.io/directory-location` label on the node objects that have a custom path. 
The value of the label is the custom kubelet root directory path, where the `/` is replaced by `-`.

The non-root enabler init container running inside the daemon pod, it will read the labels on the node is running
on. If the custom kubelet directory label is available, it will parse the value, and it will persist it into the operator 
configuration folder `/var/lib/security-profiles-operator`, such that it can be later read by the other containers. 

The `security-profiles-operator` daemon container will get now the kubelet root directory path from different 
locations in the following order:
* `/var/lib/security-profiles-operator/kubelet-config.json` file if exists
* `KUBELET_DIR` environment variable if exists
* use default Kubernetes value `/var/lib/kubelet`

Fixes #1446

```release-note
Allow to configure a custom kubelet root directory for each node or a pool of nodes.

```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -530,7 +530,20 @@ func runNonRootEnabler(ctx *cli.Context, info *version.Info) error {
 	const component = "non-root-enabler"
 	printInfo(component, info)
 	runtime := ctx.String("runtime")
-	return nonrootenabler.New().Run(ctrl.Log.WithName(component), runtime)
+
+	cfg, err := ctrl.GetConfig()
+	if err != nil {
+		return fmt.Errorf("getting config: %w", err)
+	}
+	mgr, err := ctrl.NewManager(cfg, ctrl.Options{})
+	if err != nil {
+		return fmt.Errorf("creating manager: %w", err)
+	}
+	kubeletDir, err := util.GetKubeletDirFromNodeLabel(ctx.Context, mgr.GetAPIReader())
+	if err != nil {
+		kubeletDir = config.KubeletDir()
+	}
+	return nonrootenabler.New().Run(ctrl.Log.WithName(component), runtime, kubeletDir)
 }
 
 func runWebhook(ctx *cli.Context, info *version.Info) error {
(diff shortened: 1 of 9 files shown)


JAORMX:
/hold

I'd like to discuss this further in the GH issue.

JAORMX:
@ccojocar commented here https://github.com/kubernetes-sigs/security-profiles-operator/issues/1446#issuecomment-1429253175

ccojocar:
> commented here https://github.com/kubernetes-sigs/security-profiles-operator/issues/1446#issuecomment-1429253175

@JAORMX I added more details in this comment https://github.com/kubernetes-sigs/security-profiles-operator/issues/1446#issuecomment-1429336681. I agree with your suggestion since initially I proposed something similar, but I think this could be a mid-term approach which requires more refactoring. 

We are currently blocked to roll out the operator in our production clusters by this issue, this is why I provided here more pragmatic solution which works.

saschagrunert (approved):
Yeah let's use this as mid-term solution and focus on having multiple SPOD configurations as a dedicated feature.

JAORMX:
Sounds good.

/hold cancel

ccojocar:
@JAORMX Do you want to create a follow up issue, or should I create it? Thanks

JAORMX:
@ccojocar please create it if you have a chance.

PR #1488, opened 2023-02-17
Make the priority class name configurable in the SPOD configuration

ccojocar:
#### What type of PR is this?

/kind feature

Make the priority class name configurable in the SPOD configuration, and it sets the default value to `system-node-critical`.

This is useful when the spod daemon pod remains in `Pending` state because there isn't enough capacity on the related not to be scheduled.

```release-note
Make the priority class name configurable in th SPOD configuration.
```


saschagrunert (approved):
Makes sense, thank you for the enhancement!

Issue #1482, opened 2023-02-16
Provide SPO headless CLI


saschagrunert:
From the meeting notes of Feb 15:

- Having a CLI experience to record / install profiles
- SPO CLI takes care of application execution (or maybe wrap around runc)
- Goal: Provide a user interface, especially for constrained environments

PR #1490, opened 2023-02-20
Add `spoc`, the Security Profiles Operator CLI

saschagrunert:
#### What type of PR is this?

/kind feature

We discussed in our last sync meeting that it would be helpful to have a slimmer SPO CLI for faster execution in edge scenarios. For that we would need a new binary, which can be used to directly interact with features like the ebpf recorder.

Refers to https://github.com/kubernetes-sigs/security-profiles-operator/issues/1482

#### Does this PR have test?

None

None

```release-note
Added `spoc` binary, the Security Profiles Operator CLI. This new binary is also part of the default container images.
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ LABEL name="Security Profiles Operator" \
 
 COPY --from=build /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=make /work/result/security-profiles-operator /
+COPY --from=make /work/result/spoc /
 
 USER 65535:65535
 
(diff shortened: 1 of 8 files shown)


saschagrunert:

JAORMX (approved):
I love the name!!

JAORMX:
/lgtm cancel
Let's let other maintainers look at this first

ccojocar (approved):
lgtm

How do you envision the workflow of using the CLI? Does one need to run the spoc container on the same host next to another container for which the recording is performed?  Is additionally required the existing recorder container which communicates over gRCP with spoc container, or does it get merged into spoc CLI container?

saschagrunert:
> How do you envision the workflow of using the CLI? Does one need to run the spoc container on the same host next to another container for which the recording is performed?

No, my main idea was to provide 2 distinct sets of functionality:

- Doing something without having a need for kubernetes at all, means that there is also no operator deployed
- Having sort of client interface to the operator

Means we can start with the first one and trying to integrate bpf recording of executables. Later on we can use the same binary for other features.

PR #1495, opened 2023-02-22
selinux: Allow using other container-selinux policy templates than container

jhrozek:
#### What type of PR is this?
/kind bug

The container-selinux package (at least on RHEL/Fedora) ships a number
of policy templates. One of them is called simply `container` is special
in the sense that using it also creates a `type process`, the
others do not.

The effect is that if the other templates such as `net_container` should
be used, the CIL policy must also include the raw container type.

This patch amends the generation of the CIL policy from the SPO-specific
format so that the `container` type is always used.

None

#### Does this PR have test?
Yes

N/A

```release-note
Fixed a bug where SELinux policies inheriting from another template than container
would not load correctly.
```

### DIFF
--- a/internal/pkg/translator/obj2cil.go
+++ b/internal/pkg/translator/obj2cil.go
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	typePermissive = "(typepermissive process)"
+	typePermissive         = "(typepermissive process)"
+	systemContainerInherit = "container"
 )
 
 func Object2CIL(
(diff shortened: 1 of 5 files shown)


saschagrunert (approved):
/hold

for @JAORMX eyes on this one

jhrozek:
New patch version attempts to resolve the verify failure

JAORMX (approved):

saschagrunert:
Looks like we're hitting the rate limiter more often:
```
Github rate-limiter failed the request. Either authenticate or wait a couple of minutes.
```

saschagrunert (approved):

PR #1534, opened 2023-03-09
Add `spoc run` command

saschagrunert:
#### What type of PR is this?

/kind feature

This allows to run a seccomp profile for a custom command or binary.

Refers to #1482 
#### Does this PR have test?

Yes
None

```release-note
Added `spoc run` command for experimental testing of seccomp profiles.
```


saschagrunert:

JAORMX (approved):

JAORMX:
/lgtm cancel

Let's wait for another maintainer to review.

internal/pkg/cli/runner/impl.go
```diff
@@ -0,0 +1,80 @@
...
+
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate -header ../../../../hack/boilerplate/boilerplate.generatego.linux.txt
+//counterfeiter:generate . impl
+type impl interface {
+	ReadFile(string) ([]byte, error)
+	YamlUnmarshal([]byte, interface{}) error
+	JSONMarshal(any) ([]byte, error)
+	JSONUnmarshal([]byte, any) error
+	SetupSeccomp(*specs.LinuxSeccomp) (*configs.Seccomp, error)
+	InitSeccomp(*configs.Seccomp) (int, error)
```
pjbgf, line 45: It would be great if we could have a more generic API, in which we could send different types of security profile enforcer (seccomp, apparmor, selinux, etc..). But as this is internal, we can do that at a later stage.

internal/pkg/cli/runner/runner.go
```diff
@@ -0,0 +1,91 @@
...
+	log.Print("Setting up seccomp")
+	libConfig, err := r.SetupSeccomp(runtimeSpecConfig)
+	if err != nil {
+		return fmt.Errorf("convert profile: %w", err)
+	}
+
+	log.Print("Load seccomp profile")
+	if _, err := r.InitSeccomp(libConfig); err != nil {
+		return fmt.Errorf("init profile: %w", err)
+	}
```
pjbgf, line 82: More for a follow-up: I am not sure the difference between `Init` and `Setup`. For other enforcers, maybe things like `Init` and `Load`, or `Init` and `Enforce`.

Assuming that the former initialises the enforcer and the latter actually applies it to the target application/process.

pjbgf:
Great idea and implementation! Who needs ChatGPT when @saschagrunert is around! :mechanical_arm: :rocket: 

Adding a hold in case other folks may want to comment before the merge.

/lgtm
/hold

ccojocar (approved):
Great work! 

I am wondering if is not more appropriate to name this command "validate" instead of "run" since this seems more inline with the purpose of the command.

saschagrunert:
Awesome, yeah we can follow-up on the other topics since I'm also planning to push more docs and testing around the features. I just updated the PR to propagate the correct exit code via `spoc run`.

saschagrunert:
/unhold

Thanks for all the reviews!

PR #1543, opened 2023-03-14
Fix memory optimisation feature after upgrade to controller-runtime v0.14.5

ccojocar:
#### What type of PR is this?

/kind bug

It seems that after upgrading to `controller-runtime` `v0.14.5`, the `BuilderWithOptins` function has introduced a bug when combining 
the options provided by the cluster with the custom options, which breaks the memory optimization feature.

`BuilderWithOptions` function creates a new scheme object which is configured in the cache object, but not into the cluster object 
where the original scheme object is maintained. This means that the registered schemes in the cluster scheme are not reflected 
into the cache scheme. This prevents the controller from starting with this error messages:

```
E0314 10:44:24.541264 1302480 source.go:146] controller-runtime/source "msg"="kind must be registered to the Scheme"
"error"="no kind is registered for the type v1beta1.SeccompProfile in scheme \"pkg/runtime/scheme.go:100\""
```

The issue was introduced by https://github.com/kubernetes-sigs/security-profiles-operator/commit/892ee33e52eb60679fd768523b02296ba710bd67.

It seems that our e2e test for memory optimization didn't catch the issue https://github.com/kubernetes-sigs/security-profiles-operator/blob/98d0e094e59cfb9fe5400cc63512a06e76885dbe/test/tc_memoptm_test.go#L19.

I think because we just test the start-up but in this case the spod pods got restarted after 1-2 minutes and keep restarting because of the registration issue.
#### Does this PR have test?

Yes

```release-note
Fix memory optimization feature after upgrading to controller-runtime v0.14.5.
```


saschagrunert (approved):
Thank you!

JAORMX (approved):

saschagrunert:
```
cmd/security-profiles-operator/main.go:36: File is not `gci`-ed with --skip-generated -s standard,default,prefix(sigs.k8s.io/security-profiles-operator) (gci)
	"k8s.io/apimachinery/pkg/labels"
cmd/security-profiles-operator/main.go:43: File is not `gci`-ed with --skip-generated -s standard,default,prefix(sigs.k8s.io/security-profiles-operator) (gci)
	"k8s.io/client-go/rest"
```

ccojocar:
@saschagrunert All tests passed, please could you lgtm this again? Thanks

saschagrunert (approved):

PR #1550, opened 2023-03-16
Create an architecture diagram with mermaid

ccojocar:

```release-note

```


doc/architecture.mmd
```diff
@@ -0,0 +1,64 @@
...
+        FS2[(FileSystem)]
+        P2[Pod]
+        D2[Daemon]
+        R2[Recorder]
+        D2 -.->|install profile|FS2
+        P2 -.-> |use profile|FS2
+        D2 --> |gRPC: record profile| R2
+        R2 --> |record profile| P2
+        D2 -->|get/watch| P2
+    end
```
saschagrunert, line 24: Do we want to use only 1 node for simplicity? The same would apply to everything `*2`.

saschagrunert (commented):
Great start! Do we want to add it to the install docs as well?

doc/architecture.mmd
```diff
@@ -0,0 +1,64 @@
...
+        P1 -.-> |use profile|FS1
+        D1 --> |gRPC: record profile| R1
+        R1 --> |record profile| P1
+        D1 -->|get/watch| P1
+    end
+    subgraph Node2
+        FS2[(FileSystem)]
+        P2[Pod]
+        D2[Daemon]
+        R2[Recorder]
```
saschagrunert, line 18: If we make it that detailed that we want to add the recorder, then we have to add the others as well:

- Enricher
- Metrics
- Reconciler for Seccomp, SELinux and AppArmor

doc/architecture.mmd
```diff
@@ -0,0 +1,64 @@
...
+    D1 -.->|watch|Config
+    D2 -.->|watch|Config
+    D1-.->|get/update|Recording
+    D2-.->|get/update|Recording
+
+
+    Profile[/SeccompProfile CR<br>SelinuxProfile CR <br> ApparmorProfile CR/]
+    D1 -->|create/update/watch|Profile
+    D2 -->|create/update/watch|Profile
+    U -->|create/update/get|Profile
```
saschagrunert, line 57: Deletion is a valid use case as well.

doc/architecture.mmd
```diff
@@ -0,0 +1,64 @@
...
+        D2 -->|get/watch| P2
+    end
+    
+    Recording[/ProfileRecording<br> CR/]
+    Config[/SPOD CR/]
+    W[Webhook]
+    M[Manager]
+    D[Daemonset]
+
+    U -->|create|Config
```
saschagrunert, line 32:
```suggestion
    U -->|update|Config
```

doc/architecture.mmd
```diff
@@ -0,0 +1,64 @@
...
+    U -->|deploy|M
+
+
+    W -.->|watch| Recording
+    W -->|add annotations<br>profile binding| P1
+    W -->|add annotations<br>profile binding| P2
+
+    M -.->|watch|Config
+    M -->|create/update|D
+    M -->|create|W
```
saschagrunert, line 43: Hm, the manager creates the webhook and the user deploys it as well? There seems to be something wrong here.

saschagrunert:
Profile bindings seems to be missing from the CRDs.

doc/architecture.mmd
ccojocar: This is when the webhook is statically deployed.

doc/architecture.mmd
ccojocar: We can do that, just wanted to emphasis the multiplicity but are too many details.

doc/architecture.mmd
ccojocar: added it

doc/architecture.mmd
tuxerrante: should the name be fixed in architecture.md?

doc/architecture.mmd
tuxerrante, line 1: should the name be fixed in architecture.md?

ccojocar: It's not a markdown, but mermaid extension - https://mermaid.js.org/syntax/flowchart.html. 

We can embed the svg in the installation markdown.

doc/architecture.mmd
tuxerrante: Can we add also a textual description of what if does?

ccojocar: I updated the text that is created only when is not present.

ccojocar:
@saschagrunert @tuxerrante I think that I addressed all your comments. Please have a look agin. Thanks

PR #1568, opened 2023-03-23
Fix: namespace resource should not be defined in the helm chart

### LINKED ISSUE #1567: Namespace resource should be deleted from Helm static resources
#### What happened:
Installing the 0.7.0 release on a AKS cluster through Azure pipelines goes in error since the namespace resource already exists.
The other way around is also not permitted, since if you don't have a namespace you're not allowed to create a new one since this is a "user space" installation. Also Helm is not intended to create namespaces.

> Helm 3 follows the behavior of other Kubernetes tooling and returns an error if the namespace does not exist.

https://v3-1-0.helm.sh/docs/faq/

Instead I get:
> Error: rendered manifests contain a resource that already exists. Unable to continue with install: Namespace "dev-pau-spo" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "[app.kubernetes.io/managed-by](http://app.kubernetes.io/managed-by)": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "spo"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "dev-pau-spo"

Warning: This would also delete the entire namespace when the release is deleted.
Warning: Namespace labeling seems needed by the project, I've already documented this [in the Helm installation readme](https://github.com/tuxerrante/security-profiles-operator/blob/main/installation-usage.md#installation-using-helm).

This should work in an existing namespace:
```sh
helm upgrade --namespace dev-pau-spo   --allow-unreleased   --values /azp/agent/_work/10/s/config/values.yaml spo   /azp/agent/_work/10/a/chart/security-profiles-operator-0.7.0.tgz
```

config/values.yaml
```yaml
replicaCount: 3

# This should be an apparmor-enabled compiled image
spoImage:
  registry: MYPRIVATEREPO
  repository: pau/spo-apparmor
  tag: 0.7.0

enableAppArmor: true

enableSelinux: false
enableLogEnricher: false
enableBpfRecorder: false
enableProfiling: false
verbosity: 0

nameOverride: ""
fullnameOverride: ""

serviceAccount:
  create: true
  annotations: {}
  name: ""

podAnnotations: {}

podSecurityContext: {}

resources:
  limits:
    memory: 128Mi
  requests:
    cpu: 250m
    memory: 50Mi

autoscaling:
  enabled: false
  minReplicas: 1
  maxReplicas: 100

nodeSelector:
  kubernetes.io/os: linux
  nodegroup: apps

tolerations: 
- key: pool
  operator: Equal
  value: apps
  effect: NoSchedule

affinity: {}
```

Helm install 0.7.0 release in a existing namespace

#### Environment:
AKS 1.25.5
Helm v3.9.0

tuxerrante:
#### What type of PR is this?
/kind bug

Fixes #1567 

#### Does this PR have test?
N/A
You can test trying to install it to a cluster with an existing ns, like shown in the issue.

NONE

```release-note
Fixed a bug that prevents helm install to work when installing on a cluster where the namespace already exists.
```

### DIFF
--- a/deploy/overlays/helm/delete-ns.yaml
+++ b/deploy/overlays/helm/delete-ns.yaml
@@ -1,0 +1,5 @@
+$patch: delete
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: security-profiles-operator
--- a/deploy/helm/templates/static-resources.yaml
+++ b/deploy/helm/templates/static-resources.yaml
@@ -1,19 +1,4 @@
 apiVersion: v1
-kind: Namespace
-metadata:
-  annotations:
-    meta.helm.sh/release-name: security-profiles-operator
-    meta.helm.sh/release-namespace: '{{ .Release.Namespace }}'
-  labels:
-    app: security-profiles-operator
-    app.kubernetes.io/managed-by: Helm
-    helm.sh/chart: security-profiles-operator
-    pod-security.kubernetes.io/audit: privileged
-    pod-security.kubernetes.io/enforce: privileged
-    pod-security.kubernetes.io/warn: privileged
-  name: '{{ .Release.Namespace }}'
----
-apiVersion: v1
 kind: ServiceAccount
 metadata:
   annotations:
--- a/deploy/overlays/helm/kustomization.yaml
+++ b/deploy/overlays/helm/kustomization.yaml
@@ -25,18 +25,6 @@ patches:
   target:
     kind: (ClusterRoleBinding|RoleBinding)
 
-  # Replace the namespace name into the namespace resource.
-  # Remove unused openshift label
-- patch: |
-    - op: replace
-      path: /metadata/name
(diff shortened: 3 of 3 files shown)


saschagrunert (approved):

deploy/overlays/helm/delete-ns.yaml
```diff
@@ -0,0 +1,5 @@
+$patch: delete
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: security-profiles-operator
```
saschagrunert, line 5: Let's add a newline at the end of the file otherwise editors will add whitespace

tuxerrante: sure

saschagrunert (approved):
/ok-to-test

PR #1583, opened 2023-03-29
Bug/resource limits

tuxerrante:
#### What type of PR is this?
/kind bug

There are multiple resource requests defined for CPU but no limits. 
There should be either no requests and no limits or if there is a request there should be also a respective limit.
In the current situation the webhook deployment fails to start when installed through Helm with the following errors:

> Error creating: Pod "security-profiles-operator-webhook-5d555d9d79-tdh5j" is invalid: spec.containers[0].resources.requests: Invalid value: "250m": must be less than or equal to cpu limit

None

#### Does this PR have test?
N/A
Deploying though Helm, the webhook won't start.

```release-note
NONE
```

### DIFF
--- a/bundle/manifests/security-profiles-operator.x-k8s.io_apparmorprofiles.yaml
+++ b/bundle/manifests/security-profiles-operator.x-k8s.io_apparmorprofiles.yaml
@@ -3,6 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.8.0
+  creationTimestamp: null
   labels:
     app: security-profiles-operator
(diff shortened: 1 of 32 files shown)


bundle/manifests/security-profiles-operator.x-k8s.io_apparmorprofiles.yaml
```diff
@@ -3,7 +3,6 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.8.0
-  creationTimestamp: null
```
saschagrunert, line 6: This is generated, we need those to make the CI happy.

saschagrunert (approved):

PR #1587, opened 2023-03-30
Use mergo for syscall merge

saschagrunert:
#### What type of PR is this?

/kind bug

The merging itself is not that intelligent any more, but we preserve possible args and error return values. It also makes the merge more simple, while we still sort the syscall names and actions.

None
#### Does this PR have test?
None

None

```release-note
Fixed overriding args and error return values when merging profiles.
```


saschagrunert:
I'll try something different.

ccojocar (approved):

PR #1600, opened 2023-04-04
Auto-select selinuxd image based on nodeInfo.OSImage

### LINKED ISSUE #1576: Automatic selinuxd image selection
SPO to be able to automatically select the appropriate selinuxd image for the nodes where the SPOD DS is running on.

#### Why is this needed:
Different OSs have different libsemanage versions and the version of libsemanage inside the selinuxd container must match the version of libsemanage on the host.

#### User story covered
As an infrastructure SRE, Kirsti wants to make sure that no security profiles are in an errored state

#### High level design
First, selinuxd images  must exist for the target OS. The list of possible (approved) selinuxd images must be available to SPO.

Next, because the image must be known before the DS is created, the selection must be performed by the operator, probably by the spod controller. A new workload that examines the target nodes should probably be created. 

#### More detailed design proposal

1. the list of approved images would be created as a deploy time manifest, similarly to how we create the seccomp profile for the SPO. The format will be a configMap where the key is the `CPE_NAME` key in the `/etc/os-release` file and the value is a pullspec of selinuxd.
2. The spod controller will be adjusted so that if selinuxd is enabled, a `Job` will be created with the configMap mentioned above. This job will be privileged, but the only thing it will do is to read `/etc/os-release` and write the appropriate selinuxd pullspec into the spod status.
3. this pullspec will be then used by the DS

#### Questions

- What should the semantic of `RELATED_IMAGE_SELINUXD` be? IMO if that variable is set, then the selection above should not be performed and the related image should be used.
- What if the selection fails? Is there another option than failure? (I don't think so)

jhrozek:
#### What type of PR is this?
/kind feature

Cluster nodes might be upgraded between major OS releases where the
libselinux major version is incompatible. This means that the selinuxd
image must be based on images matching the OS release and the libselinux
major version.

But SPO must also be able to select the images based on what OS are the
cluster nodes running.

To implement that, this patch adds a new key to the
security-profiles-operator-profile configmap that maps regexes matching
nodeInfo.OSImage to selinuxd images. The operator then selects the right
selinuxd image, falling back to using the `RELATED_IMAGE_SELINUXD` environment
variable.

Fixes #1576

#### Does this PR have test?
some unit tests

I'll hold this patch before I can install a RHEL-9 based cluster - I've
been having issues with that. But I wanted to get some early feedback.

```release-note
SPO now autoselects the appropriate selinuxd image based on mapping in the
security-profiles-operator-profile. If none of the entries match, SPO falls back
to the image provided by `RELATED_IMAGE_SELINUXD`.
```

### DIFF
--- a/deploy/overlays/webhook/role.yaml
+++ b/deploy/overlays/webhook/role.yaml
@@ -1,3 +1,3 @@
 ---
-- path: /rules/2
+- path: /rules/3
   op: remove
--- a/bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
+++ b/bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
@@ -98,6 +98,17 @@ data:
         }
       ]
     }
+  selinuxd-image-mapping.json: |
+    [
+        {
+            "regex":"(.*)(CoreOS).*([\\d+])\\.8[\\d+]\\.(.*)",
+            "image":"quay.io/security-profiles-operator/selinuxd-el8:latest"
+        },
+        {
+            "regex":"(.*)(CoreOS).*([\\d+])\\.9[\\d+]\\.(.*)",
+            "image":"quay.io/security-profiles-operator/selinuxd-el9:latest"
+        }
+    ]
   selinuxd.cil: |
     (block selinuxd
         (blockinherit container)
--- a/bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
@@ -308,6 +308,14 @@ spec:
       - rules:
         - apiGroups:
           - ""
+          resourceNames:
+          - security-profiles-operator-profile
+          resources:
+          - configmaps
+          verbs:
+          - get
+        - apiGroups:
+          - ""
           resources:
           - nodes
           verbs:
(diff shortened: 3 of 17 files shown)


jhrozek:
/hold
for RHEL-9 testing (the installer is having issues it seems)

deploy/overlays/webhook/role.yaml
```diff
@@ -1,3 +1,3 @@
 ---
-- path: /rules/2
+- path: /rules/3
```
jhrozek, line 2: In case someone is going wtf is this change..because I changed the spo role by adding a new permission, this overlay would have removed a permission at the wrong index:
```
diff --git a/deploy/webhook-operator.yaml b/deploy/webhook-operator.yaml
index cad32631..ed2ae5f8 100644
--- a/deploy/webhook-operator.yaml
+++ b/deploy/webhook-operator.yaml
@@ -2171,9 +2171,9 @@ rules:
   - list
   - watch
 - apiGroups:
-  - ""
+  - admissionregistration.k8s.io
   resources:
-  - services
+  - mutatingwebhookconfigurations
   verbs:
   - create
```

jhrozek:
/hold cancel
I finally got hold of a RHEL-9 cluster and the right image was selected and a smoke test passed (a SELinux policy was installed)

saschagrunert (approved):

PR #1642, opened 2023-04-21
Make the selinuxd auto-selection logic use environment variables

jhrozek:
#### What type of PR is this?
/kind cleanup

This patch doesn't change the functionality in any way whatsoever, but
makes it easier for tools that reason about bundle or helm manifests
to do their job.

I hit such issue with a tool we use in OpenShift that was not able to detect
that I forgot to rewrite the quay.io upstream images in the configMap to
our downstream images.

We used to put the mapping between selinuxd images and OS directly into
a configmap. This is problematic for the aforementioned tools as a
container pullspec is hard to find in operator manifests. To make their
job easier, the configmap now points to an environment variable name
that is defined in the operator deployment manifest.

None

#### Does this PR have test?
Adjusts existing tests

N/A

```release-note
NONE
```


jhrozek:

PR #1712, opened 2023-05-26
Recorded profiles can optionally be disabled after creation

### LINKED ISSUE #1270: Add a new flag to not install recorded profiles automatically
A profile recording should get a flag that would cause the profile to not be applied after recording.

#### Why is this needed:
we probably don't want to install profiles automatically when recording in CI

#### User story covered
As an application developer, Ming would like to be able to automatically generate initial security profiles that are specific to the application.

Acceptance criteria:

- set a flag in the recording, record a profile
- verify it's not installed (a new state should be added)

jhrozek:
#### What type of PR is this?
/kind feature
/kind api-change

In many cases it's useful to just record profiles but use them at a
later stage, e.g. record in CI, deploy later by the end user.

This patch implements a new option `disableProfileAfterRecording` for
the `ProfileRecording` CR that creates the recorded profiles using a
new state `Disabled`. Such disabled profiles are not reconciled, but
can be merged if the profile recording is set to do that.

Even though this PR adds a new attribute to the CRs, the changes are
backwards compatible as the option is to disable the profile (not
explicitly enable) and the default, both on the CRD level and the Go
structure level is false.

Fixes #1270

#### Does this PR have test?
yes

Even though this PR changes the API, I think that the change is backwards-compatible.

```release-note
The ProfileRecorder CR gets a new attribute disableProfileAfterRecording that
can be used to avoid installing profiles after recording.
```


saschagrunert (approved):
Awesome, great feature! Thank you!

PR #1843, opened 2023-08-22
Fix for selinux policy constantly being processed

### LINKED ISSUE #1842: SelinuxProfile never finishes reconcilation and causes semodule to use 100% cpu
#### What happened:
When applying an SelinuxProfile to a K8s cluster running the security-profiles-operator, the semodule process on the node uses 100% cpu. This appears to be caused by the operator constantly detecting a drift between the applied policy on the node and what is stored in the CRD. For each detection of a drift, the operator attempts to install the policy again. However, since the operator does not currently care about order in the Selinux policy it changes each time which triggers another drift. Since the policy is constantly changing, semodule runs constantly and attempts to install the selinux policy.

The operator to only install / change the policy if the upstream CRD has changed and therefore semodule only runs for actual changed policy.

Apply an SelinuxProfile CRD with selinux in enforcement mode.

no

#### Environment:
RHEL8

novaesis:
#### What type of PR is this?
/kind bug

Sorts the allow rules in the selinux policy when doing the write to the node and doing the comparison between CRD and node policy which allows for consistent comparisons.

Fixes #1842 

#### Does this PR have test?
N/A

```release-note
Fixed selinux policy constantly being processed.
```

### DIFF
--- a/internal/pkg/util/maps.go
+++ b/internal/pkg/util/maps.go
@@ -1,8 +1,24 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
 package util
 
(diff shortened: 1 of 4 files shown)


jhrozek (approved):
/lgtm
really nice work, thank you. Let's see if the tests pass!

internal/pkg/util/maps.go
```diff
@@ -0,0 +1,12 @@
+package util
```
saschagrunert, line 1: CI complains about:

```
1 files have incorrect boilerplate headers:
internal/pkg/util/maps.go
```

novaesis:
@saschagrunert fixed the issue the CI was complaint about. The other two build issues appear not related to this PR.

saschagrunert (approved):
Thank you!

Issue #1868, opened 2023-09-06
Create an "Enforce on all Pods" option for Seccomp Profile Bindings


CoreyCook8:
#### What would you like to be added:

Logic would be  added to the [Handle function](https://github.com/kubernetes-sigs/security-profiles-operator/blob/33e7640bce0816232e79f1d7096514103dc8b301/internal/pkg/webhooks/binding/binding.go#L112) to look for a container name `*` for example and then apply the seccomp profile to the pod.

This could could be scaled back through webhookOptions if there are exceptions within a namespace.

#### Why is this needed:

Profile Bindings are good for enforcing specific seccomp profiles on specific containers.

Applying a default seccomp profile is not realistic with the way they are set up currently. 

#### User story covered

As a developer, I would like to define a default seccomp profile to be applied to each pod in my namespace so that I can ensure the pods running in my cluster are secure without having to make profile bindings for each container.

PR #1869, opened 2023-09-06
Add a default profile option for profileBindings

CoreyCook8:
#### What type of PR is this?
/kind feature

This PR introduces an "Enforce for all pods" option for ProfileBinding. 

Container profile binding will have precedence here, but this will allow developers to create a default seccomp or selinux profile to be defaulted to all pods in a namespace. 

If there exists a default profile binding and there has been no match between a container and profile binding for a given pod, the default profile will be applied to the pod level securityContext. 

Fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/1868

```release-note
Adds functionality to the profile binding functionality to establish a default seccomp/selinux profile for a given namespace.
Specific image bindings have priority over the default profiles allowing more tailored profiles for specific images while allowing customization of a default profile applied to all pods without having to specify specific images strings.
```

### DIFF
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -631,6 +631,21 @@ spec:
   image: nginx:1.19.1
 ```
 
+You can enable a default profile binding by using the string "*" as the image name.
+This will only apply a profile binding if no other profile binding matches a container in the pod.
+
+```yaml
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: ProfileBinding
+metadata:
+  name: nginx-binding
+spec:
+  profileRef:
+    kind: SeccompProfile
+    name: profile-complain
+  image: *
+```
+
 If the Pod is already running, it will need to be restarted in order to pick up
 the profile binding. Once the binding is created and the Pod is created or
 recreated, the SeccompProfile should be applied to the container whose image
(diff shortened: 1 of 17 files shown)


installation-usage.md
```diff
@@ -631,6 +631,21 @@ spec:
...
+```yaml
+apiVersion: security-profiles-operator.x-k8s.io/v1alpha1
+kind: ProfileBinding
+metadata:
+  name: nginx-binding
+spec:
+  profileRef:
+    kind: SeccompProfile
+    name: profile-complain
+  image: *
```
saschagrunert, line 646: Hm, how about allowing regular expressions here in general? This would end-users allow to select a registry for example.

saschagrunert (commented):
Thank you for the contribution! One general suggestion and I assume we need a new e2e test for this.

installation-usage.md
CoreyCook8: I had considered that, the implementation for the container storage is a sync.Map which I think would be difficult to work with a regular expression. I also didn't want to add too much complexity here. So I figured start with this case since it is simple rather than introducing a lot of complexity to the solution.

I can work on the E2E test!

saschagrunert: Sounds good to me!

CoreyCook8:
> /retest

Why do you tell me I have these powers and then not listen to me!! 😭

saschagrunert (approved):

saschagrunert:
@CoreyCook8 I triggered the build for you. Can you add a release note to the release-notes block in the body of the PR description?

CoreyCook8:
Is something actually breaking here? Or is it just a flakey test?

CoreyCook8:
Bump on this, is there any work / Updates I need to add to this? Or is it just the test being flakey?

saschagrunert:
@CoreyCook8 may I ask you to rebase please?

Issue #1878, opened 2023-09-13
File Descriptor memory leak on selinux socket comms


CoreyCook8:
#### What happened:

SPOd pods were OOMing every few hours. After an investigation, it showed that file descriptors for the selinux socket were being leaked.

#### What you expected to happen:

File descriptors to not leak.

#### How to reproduce it (as minimally and precisely as possible):

Run SPOD. Check the metrics for process_open_fds and see that they are continuously increasing. 
You can check the nodes and see the open fds, check the security-profiles-operator daemon process and see that the fds are pointing to selinuxd.sock.

#### Environment:

- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Others:

PR #1879, opened 2023-09-13
Fix fileDescriptor memory leak

CoreyCook8:

This PR creates only one http client and reuses it rather than making a new client every time. 

fixes https://github.com/kubernetes-sigs/security-profiles-operator/issues/1878

```release-note

```


saschagrunert (approved):
Thank you!

PR #1956, opened 2023-11-03
Avoid listing pods in the recording webhook handler which leads sometimes to a webhook timeout issue

ccojocar:
#### What type of PR is this?

/kind bug

Avoid listing pods in the recording webhook handler, because this operation is sometimes really slow leading to a webhook timeout problem.

It seems that listing pods in a cluster with many pods is quite slow, even though the number of pods in the namespace from where they are listed is low. 

This looks like a performance problem in the Kubernetes API server. A list pods API call can take sometime over 30 seconds which is the maximum timeout for a webhook handler in Kubernetes.

The recording webhook doesn't need to list all pods being recorded anyhow in order to maintain the status in the ProfileRecording CR. Every time a Pod is being created/updated or deleted, the webhook handler is called with the Pod object, so it can update the status accordingly.

#### Does this PR have test?

N/A

#### Special notes for your reviewer:p

None

```release-note
NONE

```

### DIFF
--- a/internal/pkg/webhooks/recording/recording.go
+++ b/internal/pkg/webhooks/recording/recording.go
@@ -353,7 +353,6 @@ func (p *podSeccompRecorder) setFinalizers(
 	selector labels.Selector,
 	podLabels labels.Set,
 ) error {
-
 	if op == admissionv1.Delete {
 		if controllerutil.ContainsFinalizer(profileRecording, finalizer) {
 			controllerutil.RemoveFinalizer(profileRecording, finalizer)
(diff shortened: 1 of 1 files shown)


ccojocar:
@saschagrunert Have you managed to fix the vagrant?

saschagrunert:
@ccojocar may I ask you to rebase please?

ccojocar:
I rebased now. Thanks for pining.

saschagrunert (approved):
Thanks!

PR #2019, opened 2023-12-14
OCPBUGS-25373: Message is not a required field in the CRDs

yuumasato:
#### What type of PR is this?

/kind bug
/kind regression

- Removes `message` from the required list of CRD fields
  - This fixes errors when upgrading SPO from v0.8.0 to v0.8.1.
  - Related: https://github.com/k8s-operatorhub/community-operators/pull/3580

None

#### Does this PR have test?
N/A

```release-note
Remove "message" from the required list of CRD fields to fix issues when upgrading SPO to v0.8.1
```

### DIFF
--- a/bundle/manifests/security-profiles-operator.x-k8s.io_rawselinuxprofiles.yaml
+++ b/bundle/manifests/security-profiles-operator.x-k8s.io_rawselinuxprofiles.yaml
@@ -123,7 +123,6 @@ spec:
                       type: string
                   required:
                   - lastTransitionTime
-                  - message
                   - reason
                   - status
                   - type
--- a/bundle/manifests/security-profiles-operator.x-k8s.io_seccompprofiles.yaml
+++ b/bundle/manifests/security-profiles-operator.x-k8s.io_seccompprofiles.yaml
@@ -264,7 +264,6 @@ spec:
                       type: string
                   required:
                   - lastTransitionTime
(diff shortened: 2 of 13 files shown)


bundle/manifests/security-profiles-operator.x-k8s.io_rawselinuxprofiles.yaml
```diff
@@ -123,7 +123,6 @@ spec:
                       type: string
                   required:
                   - lastTransitionTime
-                  - message
```
saschagrunert, line 126: Those fields are auto-generated. I think we have to change the go code to fix that up.

saschagrunert (changes requested):

bundle/manifests/security-profiles-operator.x-k8s.io_rawselinuxprofiles.yaml
yuumasato: Is there a way to override the "+required"?
https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

I'm afraid the fix might mean reverting to using self defined `Condition`.

yuumasato:
I'm closing this in favor of https://github.com/kubernetes-sigs/security-profiles-operator/pull/2023

PR #1917, opened 2023-10-12
Adding first iteration on AppArmor profile generation to the recorder

0xmilkmix:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This is a PR adding features for AppArmor profiles generation to the recorder:
- Modifications to the eBPF recorder to trace executions, file operations, sockets operations and capabilities
- Modifications to the AppArmor CRD to add an `abstract` representation, allowing to generate simpler profiles (like [Bane](https://github.com/genuinetools/bane)
- Modifications to the `spoc` CLI to generate AppArmor profiles (CRD and raw) from the command line

This PR does not include the integration with the recorder controller for now. I wanted to first discussed the idea and first iteration with you. Would it be possible to have collaborations so that the code from this PR could then be integrated by you in the controller for in-cluster recording?

Also, note that using `open`/`openat` tracing has some limitations. Namely, symlinks will not be resolved and this could lead to misconfigured AppArmor profiles. A LSM hook such as `security_open` would be beneficial here.

Other known limitations in the submitted version:
- No support for recording sub-processes
- ~Sometime the recorded process could end and events would still be in the ring buffer, generating an incomplete profile~
  - fixed

#### Does this PR have test?

N/A

#### Does this PR introduce a user-facing change?

Yes

```release-note
The `spoc` cli tool now features `apparmor` and `raw-apparmor` types to generate CRDs and raw apparmor profiles.
```

PR #2066, opened 2024-01-20
Update auto generated files

pjbgf:
This is a fix for the make verify which breaks every turn of year due to the Year in the copyright.


pjbgf:
This fixes the issues with `make verify-bundle` which are currently blocking all the dependency bumps.

ccojocar:
It was updated in https://github.com/kubernetes-sigs/security-profiles-operator/pull/1917.

ccojocar:
I disable temporary the vulnerability scanning for buld-image. I don't understand why is failing.

Issue #891, opened 2022-04-08
The spod daemonset can't work,the log is that fatal error: timer when must be positive.


neblen:
#### What happened:
The spod daemonset can't work.I deploy spo0.4.1 on ubuntu k8s,but the spod can't work correctly.
#### What you expected to happen:
The spod daemonset work correctly.
#### Anything else we need to know?:
The error log is as follows.
```
I0418 05:30:48.045731 1 logr.go:252] "msg"="Set logging verbosity to 0"
I0418 05:30:48.045731 1 logr.go:252] "msg"="Profiling support enabled: false"
fatal error: timer when must be positive
goroutine 4 [running]:
runtime.throw({0x1b3a37a, 0x0})
runtime/panic.go:1198 +0x71 fp=0xc000059ee8 sp=0xc000059eb8 pc=0x438b31
runtime.modtimer(0xb418cff40fe67fd0, 0xc00004a000, 0xc00004a000, 0x1, {0x0, 0x0}, 0x1a43301)
runtime/time.go:425 +0x4a7 fp=0xc000059f30 sp=0xc000059ee8 pc=0x458407
runtime.resettimer(...)
runtime/time.go:539
time.resetTimer(...)
runtime/time.go:229
runtime.scavengeSleep(0x1e8480)
runtime/mgcscavenge.go:237 +0x73 fp=0xc000059f80 sp=0xc000059f30 pc=0x423a53
runtime.bgscavenge()
runtime/mgcscavenge.go:366 +0x205 fp=0xc000059fe0 sp=0xc000059f80 pc=0x423cc5
runtime.goexit()
runtime/asm_amd64.s:1581 +0x1 fp=0xc000059fe8 sp=0xc000059fe0 pc=0x46c9c1
created by runtime.gcenable
runtime/mgc.go:182 +0x65
goroutine 1 [runnable]:
flag.sortFlags(0x0)
flag/flag.go:353 +0x51
flag.(*FlagSet).Visit(0xc000082d80, 0xc000461b48)
flag/flag.go:407 +0x25
github.com/urfave/cli/v2.normalizeFlags({0xc000082c80, 0x3, 0xc0000b2d80}, 0xc000327200)
github.com/urfave/cli/v2@v2.3.0/context.go:174 +0xec
github.com/urfave/cli/v2.(*Command).parseFlags(0xc0000b2d80, {0x1e12ab8, 0xc0002425b8}, 0xb0)
[...]

PR #2121, opened 2024-02-12
Update security-profiles-operator-profile_v1_configmap.yaml with clock_gettime syscall

CoreyCook8:
Add clock_gettime to address issue referenced here:  https://github.com/kubernetes-sigs/security-profiles-operator/issues/891

Issue caused one spod pod to be in CBLO. 
Debugged by manually setting the default_action to SCMP_ACT_LOG and then checking the node logs.
Seccomp was being violated with the `clock_gettime` call. Adding that call here to fix that issue.

#### What type of PR is this?

/kind bug

```release-note
Fixed issue with crashing SPOD daemon by allowing `clock_gettime` syscall.
```

### DIFF
--- a/bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
+++ b/bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
@@ -20,6 +20,7 @@ data:
             "capget",
             "capset",
             "chdir",
+            "clock_gettime",
             "clone",
             "clone3",
             "close",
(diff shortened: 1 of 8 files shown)


saschagrunert (approved):
Thank you!

JAORMX (approved):
/lgtm

kecebon9:
Hi Team,
I’ve recently encountered an issue in a cluster with over 20 nodes, where the problem only affects one specific DaemonSet. I’ve verified that all nodes are using the same OS, kernel, and architecture versions, and that the date/time is synchronized across all of them. I also tried to reproduce the issue locally in a kind cluster, but I couldn’t replicate it.

The issue appears to be resolved by adding the clock_gettime syscall, as referenced in this PR. Any ideas on what the root cause might be? We’re curious why this issue suddenly occurred and why it only affects one node.

Thanks

PR #2145, opened 2024-03-04
fix: dynamic clusters sometimes encounter a finalizer mismatch

jlowe64:
#### What type of PR is this?

/kind bug

This PR attempts to fix an issue with dynamically scaling clusters, such as ones using Karpenter, where the number of nodes can remain the same while the contents of the profile's finalizers may differ. This will result in not being able to manage the profile without interacting with the finalizers directly.

None

#### Does this PR have test?

N/A

NONE

```release-note
NONE
```


saschagrunert:
Thank you for the PR @jlowe64! Do you mind signing the CLA?

jlowe64:
> Thank you for the PR @jlowe64! Do you mind signing the CLA?

I did this earlier, I was just waiting on someone from my team to tell me how our company handles open source contributions! Thank you.

Also, could you please hold off on approving the automated tests (I believe you have to approve them), I need to push to my fork cause I'm doing testing on a linux workstation and need to commit in order to run tests. I'll poke you on here when I want the tests to be run. I don't want to waste your resources when I am fixing issues from the tests!

jlowe64:
I'm not ready to merge this, I need to look at it some more.

internal/pkg/util/nodeutils.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, err
+	}
+
+	// Extract node names
+	nodeNames := make([]string, 0, len(nodeList.Items))
+	for _, item := range nodeList.Items {
+		var node map[string]interface{}
+		err := runtime.DefaultUnstructuredConverter.FromUnstructured(item.Object, &node)
+		if err != nil {
+			return nil, err
```
ccojocar, line 68: nit: I would wrap these errors with some context.

internal/pkg/util/nodeutils.go
```diff
@@ -0,0 +1,106 @@
...
+		nodeStatus := &nodeStatusList.Items[i]
+		if !StringInSlice(currentNodeNames, nodeStatus.NodeName) {
+			// We've found a node that doesn't exist anymore
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func StringInSlice(list []string, str string) bool {
```
ccojocar, line 99: I think there is an `util.Contains` function already for this.

internal/pkg/util/nodeutils.go
```diff
@@ -0,0 +1,106 @@
...
+func GetNodeList(ctx context.Context) ([]string, error) {
+	dynamicClient, err := GetDynamicClient()
+	if err != nil {
+		return nil, err
+	}
+	// Specify the resource (nodes) and namespace
+	nodeResource := schema.GroupVersionResource{Group: "", Version: "v1", Resource: "nodes"}
+
+	// List the nodes (using the dynamic client)
+	nodeList, err := dynamicClient.Resource(nodeResource).Namespace("").List(ctx, metav1.ListOptions{})
```
ccojocar, line 57: How does this call perform for big clusters with hundreds of nodes? Do we need to use any limit here when fetching the node objects?

internal/pkg/util/nodeutils.go
jlowe64: No prob, I've added context!

internal/pkg/util/nodeutils.go
jlowe64: @ccojocar I am doing something slightly different in that I'm not looking for an equality of strings, but rather if a substring is contained within a string. I looked at your function and I've matched the naming convention of what you have for `util.Contains`, by calling it `util.ContainsSubstring`.

internal/pkg/util/nodeutils.go
jlowe64: @ccojocar The largest cluster I've got access to is about 150 nodes (although during peak it could grow bigger). I have no access to multi-hundreds of nodes clusters in a test infrastructure. My test cluster is about 50~ nodes. Since I'm trying to determine if all nodes match the finalizers on an object (not just in number) it might be better to skip the check if we see performance issues on multi-hundred node clusters. 

Off the top of my head I can't think of an example of doing a check like this in a performant way to gain some inspiration, do you know of any such check done by another project?

internal/pkg/util/nodeutils.go
```diff
@@ -0,0 +1,106 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
saschagrunert, line 2:
```suggestion
Copyright 2024 The Kubernetes Authors.
```

ccojocar (approved):
Thanks, it looks good!

saschagrunert (approved):
Is this still `WIP` @jlowe64 ?

jlowe64:
> Is this still `WIP` @jlowe64 ?

I have just removed the WIP tag.

PR #2141, opened 2024-03-01
Add End-to-End Tests for Spoc

mhils:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

I have a series of (work-in-progress) BPF recorder improvements sitting at https://github.com/kubernetes-sigs/security-profiles-operator/compare/main...mhils:security-profiles-operator:wip that I'd like to upstream. I'd like to kick this off with end-to-end tests for spoc so that we have more confidence in further PRs. :)

#### Which issue(s) this PR fixes:

refs (but not fixes) https://github.com/kubernetes-sigs/security-profiles-operator/issues/718

#### Does this PR have test?

Yes.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #2155, opened 2024-03-08
BPF Recorder Cleanup + Child Process Tracing

mhils:
#### What type of PR is this?

/kind cleanup

This PR cleans up the BPF C code to remove some duplication. 

It also adds support for tracing child processes in AppArmor profiles. I have more AppArmor improvements ready for a separate PR, but this particular one is a bit hard to take out of the existing commit chain. Hope that's ok. :)

None

#### Does this PR have test?

Yes, in #2141.

The best way to review this is probably to go through the individual commits, I've tried to keep them fairly targeted on doing one specific thing. :)

```release-note
NONE
```


internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -112,11 +113,9 @@ static __always_inline int save_args(u32 event_id, const char * filename,
     return 0;
 }
 
-static __always_inline int load_args(u32 event_id, const char ** filename,
-                                     u64 * flags)
+static __always_inline int load_args(const char ** filename, u64 * flags)
 {
-    u32 tgid = bpf_get_current_pid_tgid();
-    u64 id = ((u64)event_id << 32) | tgid;
+    u32 id = bpf_get_current_pid_tgid();
```
ccojocar, line 118: Is this state map just used internally in eBPF to keep track of arguments when the syscall enters and exits the trace point? Is the event id not anymore needed? I suppose was not parsed in the Go userspace code since I don't see any change in there.

ccojocar (approved):
Nice clean-up, thanks for this!

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
mhils: > Is this state map just used internally in eBPF to keep track of arguments when the syscall enters and exits the trace point?

Correct. This is just temporary storage so that we have the filename in the exit hook.

> Is the event id not anymore needed?

From chatting with @0xmilkmix (who wrote the first iteration): Initially this code used the PID as storage key only, but that was causing race conditions with concurrent syscalls from different threads. A first attempt to fix this was to use PID + event id (=syscall name) as a the key. Only later on @0xmilkmix discovered pid_tgid, which uniquely identifies threads. With the proper fix in place, the event id is not needed anymore. :)

PR #2201, opened 2024-04-09
Add `spoc convert` Command

mhils:
#### What type of PR is this?

/kind bug
/kind feature

This command allows users to transform security profiles from a custom resource definition YAML to the raw profile as consumed by the respective security module. My primary use case here is testing AppArmor rules without dealing with Kubernetes, but it works for seccomp just as well.

None

#### Does this PR have test?

Yes.

This PR first consolidates the code paths in `artifact` and `merger` to read YAML files (fixing ab bug in the artifact impl, see https://github.com/kubernetes-sigs/security-profiles-operator/pull/2136#discussion_r1504981477), and then reuses that for the convert functionality. It's probably easiest to review commits individually.

Happy to bikeshed on the name of the command if you feel `convert` is too generic. I hold no strong opinions. :)

```release-note
Added a `spoc convert` command to transform security profile YAML definitions to their raw representation.
```


saschagrunert:
/ok-to-test

Thank you @mhils :pray: !

internal/pkg/artifact/artifact.go
```diff
@@ -337,41 +337,33 @@ func (a *Artifact) Pull(
 		return nil, fmt.Errorf("read profile: %w", err)
 	}
 
-	a.logger.Info("Trying to unmarshal seccomp profile")
-	seccompProfile := &seccompprofileapi.SeccompProfile{}
-	err = a.YamlUnmarshal(content, seccompProfile)
-	if err == nil {
+	profile, err := a.ReadProfile(content)
+	if err != nil {
+		return nil, fmt.Errorf("%w: last err: %w", ErrDecodeYAML, err)
```
saschagrunert, line 342: Would `errors.Join` work here?

internal/pkg/artifact/profile_reader.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+
+	apparmorprofileapi "sigs.k8s.io/security-profiles-operator/api/apparmorprofile/v1alpha1"
+	seccompprofileapi "sigs.k8s.io/security-profiles-operator/api/seccompprofile/v1beta1"
+	selinuxprofileapi "sigs.k8s.io/security-profiles-operator/api/selinuxprofile/v1alpha2"
+)
+
+func ReadProfile(content []byte) (client.Object, error) {
```
saschagrunert, line 30: Please add a doc string.

internal/pkg/artifact/artifact.go
mhils: Yes, great call! 👍

saschagrunert (approved):

PR #2260, opened 2024-05-14
Refactor BPF recorder to enable simultaneous profile recording

### LINKED ISSUE #2255: Proposal: Allow `spoc` to simultaneously record different profile types.

spoc currently has `--type` and `--output-file` options for the profile that should be recorded. This means we cannot easily record Seccomp and AppArmor profiles at the same time: it's either `--type seccomp` or `--type apparmor`. It would be useful if we could record both AppArmor and Seccomp profiles at the same time. This requires some changes to the CLI interface:

1. **Option A:** Add dedicated `--apparmor`, `--apparmor-raw`, `--seccomp`, ... options. A user could run
   ```
   spoc --apparmor-raw apparmor-profile.yaml --seccomp seccomp-profile.yml ...
   ```
   to record both Seccomp and AppArmor profiles simultaneously. This would break the existing command line interface.
2. **Option B:** Make `--type` support new "combined" types such as `apparmor,seccomp` or simply a new meta type `all`. A user could run 
   ```
   spoc --type all --output-file profile.yaml ...
   ```
   to record both Seccomp and AppArmor CRD profiles simultaneously. We're limited to a single (combined) outfile.

Any thoughts/opinions? :)

mhils:
#### What type of PR is this?

/kind bug
/kind cleanup
/kind feature

This PR:

1. Separates the BPF recorder from Seccomp and AppArmor logic and puts those parts into `bpfrecorder_seccomp.go` and `bpfrecorder_apparmor.go` respectively.
2. Updates the BPF recorder to use LSM and tracepoint/sched hooks instead of hooking syscalls. This increases recorder coverage (for example, we previously missed some lockfiles, which caused incomplete profiles) and enables us to use `bpf_d_path()` to get absolute file paths in all cases. It also fixes tracking of child-child-processes. This makes the AppArmor recorder work flawlessly with several real-world workloads.
3. Adds `--type all` for the recorder to simultaneously record AppArmor and Seccomp profiles.

Fixes #2255

#### Does this PR have test?

Yes.

The good news is that we remove significantly more lines than we add, the bad news is that this PR is still quite large. This thing emerged from a rather iterative integration hell, let me know if we need to split it up further. :)

```release-note
`spoc` can now record Seccomp and AppArmor profiles simultaneously.
The AppArmor recorder is now significantly more robust
```


ccojocar (approved):
Thanks, this is a great refactoring! I didn't spot any issues. 😄

mhils:
Thank you for the super quick review, @ccojocar - you folks are fantastic! 😃 🍰 

CI has unfortunately uncovered a new issue I wasn't aware of before: Modern kernels all have BPF LSM support, but Debian/Ubuntu has not enabled it by default (see https://github.com/lockc-project/lockc/issues/159 and  https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2054810). This means the AppArmor recorder won't work on these systems properly as these hooks are silently ignored (Exhibit A: `test / e2e-spoc` fails in CI). Unfortunately there are no good workarounds. We do really want `bpf_d_path` to get absolute paths, the previous implementation was pretty broken. But the non-LSM hooks [allowlisted to use bpf_d_path](https://github.com/torvalds/linux/blob/a5131c3fdf2608f1c15f3809e201cf540eb28489/kernel/trace/bpf_trace.c#L937-L964) are not comprehensive enough to cover our needs. So we have to use LSM hooks for recording.

I've updated `AppArmorRecorder.Load` to error at runtime if we detect that LSM hooks are disabled. This unfortunately means we cannot run all the spoc tests in CI currently, but I've made it so that we still cover the AppArmor parts that do not rely on file paths.

ccojocar:
> I've updated AppArmorRecorder.Load to error at runtime if we detect that LSM hooks are disabled. This unfortunately means we cannot run all the spoc tests in CI currently, but I've made it so that we still cover the AppArmor parts that do not rely on file paths.

Isn't possible to enable it into the integration tests for Ubuntu? Does it require to recompile the kernel with the LSM hook flag enabled? Not sure what kernel version is used by the integration tests, but it would be great to have a test for this to avoid future breakage. Maybe you can deploy a dedicated VM image.

ccojocar (approved):

saschagrunert:
The profile merging test seems to fail now:

```
TestSuite/TestSecurityProfilesOperator/cluster-wide:_profile_merging
```

mhils:
> Isn't possible to enable it into the integration tests for Ubuntu? Does it require to recompile the kernel with the LSM hook flag enabled? Not sure what kernel version is used by the integration tests, but it would be great to have a test for this to avoid future breakage. Maybe you can deploy a dedicated VM image.

My understanding is that it requires a recent Kernel (which we have in CI), but also [boot-time configuration](https://eunomia.dev/tutorials/19-lsm-connect/#verifying-bpf-lsm-availability) which is not set on the GitHub-hosted runners. The libbpf are running custom kernels via QEMU - that would be an alternative, but also comes with quite a lot of CI infrastructure that we would need to maintained (the libbpf folks have their own repo just for the GitHub Actions - https://github.com/libbpf/ci/tree/main). 

I have updated `test/spoc/e2e_spoc_test.go` so that we always test AppArmor socket and capability recording, even if BPF LSM support is missing on the host. This ensures that the AppArmor subsystem is working, which I think should cover most things that could possibly break. I feel it's probably ok to accept the risk for now. Eventually https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2054810 should be fixed and we get this fully covered in CI again without spending any effort on it.



Will take a look at the remaining CI failures later today. :)

mhils:
CI failures fixed in 8d28fc6. 😃

mhils:
I have GHA tests passing for the same code at https://github.com/mhils/security-profiles-operator/pull/4, so I assume the previous failures were just flaky tests. :)

ccojocar:
> My understanding is that it requires a recent Kernel (which we have in CI), but also [boot-time configuration](https://eunomia.dev/tutorials/19-lsm-connect/#verifying-bpf-lsm-availability) which is not set on the GitHub-hosted runners. The libbpf folks are running custom kernels via QEMU to work around this. That would be an alternative, but also comes with quite a lot of CI infrastructure that we would need to maintained (they have their own repo just for the GitHub Actions - https://github.com/libbpf/ci/tree/main).

We use the `ubuntu-22.04` as an action image and on top of it runs vagrant with different images such as ubunut, fedora and flatcar where we run the e2e tests. This provides us some flexibility since the LSM needs to be enabled on vagrant image. I think we can add a dedicated e2e job for debian with LSM enabled (LSM should be enabled by default in Debian as I understand from [ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/2054810), or a custom ubuntu image with LSM enabled.

In case the e2e tests are taking too long, we can think to phase out the flatcar e2e test. I am not sure how much is used these days.

I would add this e2e test as a follow up pull request. I guess you can follow the same approach we have for [fedora.](https://github.com/kubernetes-sigs/security-profiles-operator/blob/8324a7ec7d76197ce06fbcfcc5eb77c339ce52f1/.github/workflows/test.yml#L63).

@saschagrunert What do you think?

ccojocar (approved):
Thanks for this great contribution!

saschagrunert (approved):

PR #2296, opened 2024-06-09
Integrate the eBPF based AppArmor recorder into the API

ccojocar:
#### What type of PR is this?

/kind feature

It integrates the eBPF based AppArmor profile recorder into the API. This will allow to record AppArmor profiles for workloads directly into the cluster in the same fashion like seccomp profiles.

#### Does this PR have test?

Yes

I will follow up with some integration tests in a separate pull request.

```release-note

Add the eBPF based AppArmor profile recorder into the API.
```

cc @mhils @@milkmix_


cmd/security-profiles-operator/main.go
```diff
@@ -506,7 +506,7 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
 func runBPFRecorder(_ *cli.Context, info *version.Info) error {
 	const component = "bpf-recorder"
 	printInfo(component, info)
-	return bpfrecorder.New("", ctrl.Log.WithName(component), true, false).Run()
+	return bpfrecorder.New("", ctrl.Log.WithName(component), true, true).Run()
```
mhils, line 509: All in favor of this, but do we have any (rough) estimations on additional memory usage? The number of syscalls is naturally limited, the number of accessed file paths not so much. 🫣

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -54,17 +54,24 @@ var appArmorHooks = []string{
 
 var pathWithPid *regexp.Regexp = regexp.MustCompile(`^/proc/\d+/`)
 
+// mntnsID is a unique identifier for a group of processes usually running in a container
+// Note: on a host running concurrent containers, there will be multiple process running with
+// the same PID but they are assigned to different mntns since  they run in different containers.
+// Therefore, in order to have unique apparmor profiles, each profile should be recorded using
+// mntns + pid as a key identifier.
```
mhils, line 61: I'm not sure I understand the mntns + pid part. 

My understanding is that right now (with this PR) we record by mntns and disregard the pid, which makes sense to me: Child processes within a container effectively inherit the main process' security profile (and contribute to it during recording). Do we have more advanced plans here? :)

(Nit: mntns + pid isn't strictly unique, we would need pidns + pid :))

mhils (commented):
🚀 🚀🚀

cmd/security-profiles-operator/main.go
ccojocar: This is just enabling the recorder into the operator. If no `ProfileRecording` CR is created, there will not be any recording triggered. 

I still need to add some integration tests and do more testing.

Also in the past we did some memory optimisation for operator itself when it comes to big clusters with a lot of nodes. It used to run out of memory. If I understand correctly, I think you are referring to the memory usage of the eBPF recorder when a recording is performed. 

Do you have any numbers for apparmor profiles when they are recorded form the CLI?

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
ccojocar: We introduced mount namespace originally because we had concurrent recording for two different containers running into a cluster. For instance, two different ProfileRecording CRs are creating in simultaneously which triggers two parallel recordings. 

Since the instrumentation is shared, the pidid might be the same for these two containers, but the mount ns is unique per container as far as I am aware of. The combination of two, allowed us to distinguish between the events coming from these two recording when they are sent up by the kernel.

ccojocar:
@mhils please could you review the changes and let me know if you find any issues? Thanks a lot!

cmd/security-profiles-operator/main.go
mhils: > If no `ProfileRecording` CR is created, there will not be any recording triggered.

Ok. If we only persist things for the containers for which a `ProfileRecording` CR is created, I am much less worried!

> Do you have any numbers for apparmor profiles when they are recorded form the CLI?

This is highly specific on the workload. For the workloads I tested so far it's fairly negligible (in the order of 1kb/container), but it's a very different story if we have a container that continuously writes temporary files with random names. That would just grow without bounds at the moment I think. There really isn't a great way to fix that AFAIK, so we probably just want to accept the risk. :)

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
mhils: > We introduced mount namespace originally because we had concurrent recording for two different containers running into a cluster. For instance, two different ProfileRecording CRs are creating in simultaneously which triggers two parallel recordings.

All makes sense. 👍 

> The combination of two, allowed us to distinguish between the events coming from these two recording when they are sent up by the kernel.

Where are we combining mntns and pid? I thought we only use mntns to distinguish containers. Put differently, it's 

```
recordedCapabilities     map[mntnsID][]int
```
and not 
```
recordedCapabilities     map[mntnsIDAndPid][]int
```

Why does the docstring read "each profile should be recorded using mntns + pid as a key identifier." and not just "each profile should be recorded using mntns as a key identifier."?

(The implementation makes sense to me, I'm only curious why the pid is explicitly mentioned in the docstring :))

internal/pkg/version/version_test.go
```diff
@@ -25,6 +25,11 @@ import (
 func TestVersionText(t *testing.T) {
 	t.Parallel()
 
+	// buildDate and version are typically empty during test because they are populated
+	// by the build system. Let's set some default value to avoid test failures.
+	// buildDate = "2024-06-09T17:39:57Z"
+	// version = "1.0.0"
```
mhils, line 31: Leftover?

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
```diff
@@ -312,10 +312,65 @@ func (b *BpfRecorder) SyscallsForProfile(
...
+		return nil, errors.New("no apparmor profiles recording running")
+	}
+	b.logger.Info("Getting apparmor profile for profile " + r.Name)
+
+	mntns, err := b.getMntnsForProfileWithRetry(r.Name)
+	if err != nil {
+		return nil, err
+	}
+	b.loadUnloadMutex.RLock()
+	apparmor := b.AppArmor.GetAppArmorProcessed(mntns)
```
mhils, line 354: For Seccomp we have `PopSyscalls` (which also removes the recorded events for the specific mntns). Should we do `PopAppArmorProcessed` as well?

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
ccojocar: comment updated

internal/pkg/version/version_test.go
ccojocar: Yeah, I cleaned up.

internal/pkg/daemon/bpfrecorder/bpfrecorder.go
ccojocar: I think this is not necessary for apparmor because the events are streamed from eBPF into the user space while for seccomp the intercepted syscalls are stored directly into the map by the eBPF program. This map is shared and needs to be cleaned when the syscalls are read form the user space. This is the reason for `PopSyscalls`.

ccojocar:
@saschagrunert @mhils This is ready for final review. All tests finally passed. Thanks

installation-usage.md
```diff
@@ -1800,6 +1802,15 @@ you can restrict Pod's access with the annotation:
 When [AppArmor becomes GA](https://github.com/kubernetes/enhancements/pull/3298) a new field within SecurityContext will be created to replace the annotations above.
 For up-to-date information on how to use AppArmor in Kubernetes, refer to the [official documentation](https://kubernetes.io/docs/tutorials/security/apparmor/).
 
+### Record a AppArmor profile
+
+The operator is able to record AppArmor profiles for a workload using the build-in eBPF recorder. For more details, please
+refer to the [eBPF based recording](#ebpf-based-recording)) section in `seccomp` documentation.
```
saschagrunert, line 1808:
```suggestion
refer to the [eBPF based recording](#ebpf-based-recording) section.
```

saschagrunert (commented):
Great work! Just a documentation nit, otherwise LGTM.

installation-usage.md
ccojocar: Thanks, I fixed it.

test/spoc/e2e_spoc_test.go
```diff
@@ -154,13 +150,11 @@ func recordAppArmorTest(t *testing.T) {
 				break
 			}
 		}
-
 		// Run binary...
-		cmd2 := exec.Command(demobinary, "--net-tcp")
+		cmd2 := exec.Command(demobinary, "--net-tcp", "--sleep", "30")
```
mhils, line 154: Why the added sleep here?

internal/pkg/cli/recorder/recorder.go
```diff
@@ -95,6 +95,22 @@ func (r *Recorder) Run() error {
 		ch := make(chan os.Signal, 1)
 		r.Notify(ch, os.Interrupt)
 		log.Print(WaitForSigIntMessage)
+
+		// searching the mntns of the process started outside of spoc
+		cmd := r.options.commandOptions.Command()
+		if err := util.Retry(func() (err error) {
```
mhils, line 101: For seccomp, it currently works as follows if `--no-proc` is passed:

1. We record syscalls for all mount namespaces by binaries [with the target program name](https://github.com/kubernetes-sigs/security-profiles-operator/blob/32fe817cea4b121d248df110f99eba267af7f9dc/internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c#L126-L131).
2. We merge them below in processSeccomp.

Is there a reason why we don't want to do the same for AppArmor, i.e. merge all recorded profiles?

mhils (commented):
🚀

test/spoc/e2e_spoc_test.go
ccojocar: It was not enough when running the tests. Sometimes needed to run a bit longer.

mhils: Is this because we now have a race with `ProcessIDByName`? I tried hard to make things work for short-lived processes, so if there is a different reason I'd be happy to dig into things.

internal/pkg/cli/recorder/recorder.go
ccojocar: My understanding of this flag is that the process is not started by the spoc CLI but externally and spoc will start recording the profile for it as soon as the PID comes live.

I think in the previous implementation,  it was recording everything for all process running into the system until it got interrupted. 

In this implementation, it will look up first the PID by command name, and then the namespace where that PID is running. This is required to retrieve the recorded information from the maps by namespace.

Now we create a apparmor profile for an entire namespace where the process is running. This is also the case for seccomp. This is not ideal, if other process run in parallel in that namespace when the CLI is recording. The profiles will contain more stuff than expected. I think for CLI this is acceptable, since people will most likely use this feature to record profiles for a single process running into a wider system, and that process will have its own namesapce unless they will start it in a dedicated namesapce. 

I would argue that this feature is probably not the best option, if you want to  record a profile for an entire container  (e.g. form outside). In that case, I would use the in-cluster option.

ccojocar:
@saschagrunert @mhils if you are happy please could you approve? I'll follow up with e2e tests and other small fixes if needed since this got already quite large. Thanks a lot!

test/spoc/e2e_spoc_test.go
ccojocar: Yes, that's the reason. It takes a bit of time to find the pid into the list. It is less than 30 seconds, but this is large enough to keep the tests consistently running.

saschagrunert (approved):

internal/pkg/cli/recorder/recorder.go
mhils: > I think in the previous implementation, it was recording everything for all process running into the system until it got interrupted.

Not quite. The ebpf code itself [checks](https://github.com/kubernetes-sigs/security-profiles-operator/blob/32fe817cea4b121d248df110f99eba267af7f9dc/internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c#L126-L131) the program name and only records events for matching processes. So yes, you would get all events for "curl", no matter from which namespace, but you would not get events from the rest of the system. This is why the seccomp implementation [takes all namespaces into account](https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/internal/pkg/cli/recorder/recorder.go#L183) (mntns is 0) with `--no-proc`. I think we could do the same for AppArmor and avoid the PID race.

PR #2299, opened 2024-06-10
Make selinuxd images configurable

mikroskeem:
#### What type of PR is this?

/kind feature

Makes selinuxd images configurable (in Helm chart)

One reason would be mirroring images on own registry, and pulling images from there.

However, while deploying SPO on Fedora CoreOS I found out that default image is not working for my deployment. This makes images configurable in more convenient manner.

See [discussion on Slack](https://kubernetes.slack.com/archives/C013FQNB0A2/p1718011082988209)

None

#### Does this PR have test?

N/A

```release-note
Make selinuxd images configurable in Helm chart
```

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -6,6 +6,20 @@ spoImage:
   repository: k8s-staging-sp-operator/security-profiles-operator
   tag: latest
 
+selinuxdImage:
+  default:
+    registry: quay.io
+    repository: security-profiles-operator/selinuxd
+    tag: latest
+  el8:
+    registry: quay.io
+    repository: security-profiles-operator/selinuxd-el8
+    tag: latest
+  el9:
+    registry: quay.io
+    repository: security-profiles-operator/selinuxd-el9
+    tag: latest
+
 enableSelinux: false
 enableLogEnricher: false
 enableAppArmor: false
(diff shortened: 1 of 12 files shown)


deploy/helm/values.yaml
```diff
@@ -6,6 +6,20 @@ spoImage:
   repository: k8s-staging-sp-operator/security-profiles-operator
   tag: latest
 
+selinuxdImage:
+  default:
+    registry: quay.io
+    repository: security-profiles-operator/selinuxd
+    tag: latest
```
saschagrunert, line 13: We can also add security-profiles-operator/selinuxd-fedora :)

saschagrunert (changes requested):

saschagrunert:
Thank you!

/ok-to-test

deploy/helm/values.yaml
mikroskeem: There's no environment variable which reads this however

mikroskeem: Perhaps this will do? fd182ac7b5c7b3d6907eab5280d5015d8d8657e7

saschagrunert: Yep, thanks!

saschagrunert (approved):

saschagrunert (approved):

PR #2333, opened 2024-07-02
`spoc record --no-proc`: include all namespaces

mhils:
#### What type of PR is this?

/kind cleanup

This PR builds on #2296: Since we already filter by program name when running `spoc record`, we don't need to limit ourselves to a single mntns. This way we don't have to apply somewhat brittle heuristics to find the correct namespace. One result of this is that we can remove the `--sleep` in the test!

None

#### Does this PR have test?

Yes.

```release-note
NONE
```


ccojocar (approved):

saschagrunert (approved):
/ok-to-test

saschagrunert:
/lgtm cancel

```
internal/pkg/cli/recorder/recorder.go:307:3: type assertion must be checked (forcetypeassert)
		spec = profile.(*apparmorprofileapi.AppArmorProfile).Spec
```

ccojocar (approved):

PR #2357, opened 2024-07-13
Replace the variance in the apparmor file paths with apparmor variables

ccojocar:
#### What type of PR is this?

/kind bug

It replaces the variance such as task ID and container ID from file paths recorded in the apparmor profile.

The task ID is replaced with the apparmor variable `@{tid}` and the container ID just with `*` which should
match any container ID in the path.

#### Does this PR have test?

Yes

```release-note
It replaces the variance such as task ID and container ID from files paths recorded in apparmor profile.
```

cc @mhils


mhils (commented):
LGTM from my end! 😃  Any particular reason why you moved the regex into the method? This is likely negligible here, but caching the `MustCompile` result does not hurt typically. :)

saschagrunert (approved):
Thanks!

ccojocar:
> 😃 Any particular reason why you moved the regex into the method? This is likely negligible here, but caching the MustCompile result does not hurt typically. :)

I reduced the scope since was not globally used. I think is enough to defined it locally.

Issue #2388, opened 2024-07-31
Make the `Policy` field optional and eventually deprecate/remove it from the AppArmorProfile CR


ccojocar:
#### What would you like to be added:

The [Policy](https://github.com/kubernetes-sigs/security-profiles-operator/blob/d755d59f66eba5a052313889c2915464579976a1/api/apparmorprofile/v1alpha1/apparmorprofile_types.go#L71) field was originally introduced to store the raw apparmor profile installed by the operator. Later, a new field [Abstract](https://github.com/kubernetes-sigs/security-profiles-operator/blob/d755d59f66eba5a052313889c2915464579976a1/api/apparmorprofile/v1alpha1/apparmorprofile_types.go#L72) was additionally added. The later defines a detailed schema of the apparmor profile. This looks like a duplication in the API which can lead to complications and out of sync data.

At the moment, the `Policy` is a [requied](https://github.com/kubernetes-sigs/security-profiles-operator/blob/d755d59f66eba5a052313889c2915464579976a1/deploy/operator.yaml#L2184) field in the AppArmorProfile CR definition. This make the `Policy` field mandatory, otherwise the API validation will fail before reaching the operator.

The plan is to make this optional, and use the abstract as a source for apparmor profiles. The raw profile can be generated on the fly by the operator when the `Policy` field is empty. 

Eventually, we should deprecate and remove the policy field. Not sure, what is the impact of directly removing it. This can break the cluster where apparmor profile is installed via the `Policy` field. 

#### Why is this needed:

Improve the ApparmorPolicy API by removing data duplication.

ccojocar:
@mhils Was this completed? What's still left to be done? Thanks

mhils:
This can be closed, but I don't have permissions to do so. :)

PR #2428, opened 2024-08-30
spoc: only include abstract policy in recorded apparmor profiles

mhils:
#### What type of PR is this?

/kind bug

This PR unbreaks some spoc-related workflows. Profiles recorded by spoc should only have an abstract profile. `spoc convert` can be used to render the raw AppArmor template.

refs #2388

#### Does this PR have test?

N/A

```release-note
AppArmor profiles recorded by spoc now include the abstract profile only, which ensures that the raw profile does not diverge.
```


saschagrunert (approved):

saschagrunert:
/ok-to-test

Thank you!

PR #2554, opened 2024-11-13
AppArmor: Add read permission for executables

mhils:
#### What type of PR is this?

/kind bug

Depending on how processes are spawned, `ix` is not enough and we also need to explicitly permit reading the executable.

None

#### Does this PR have test?

N/A

```release-note
Fix a bug where recorded AppArmor profiles would prevent executables from spawning.
```

### DIFF
--- a/examples/apparmorprofile-sleep.yaml
+++ b/examples/apparmorprofile-sleep.yaml
@@ -6,7 +6,7 @@ policy: |2
 
     # Executable rules
 
-    /bin/busybox ix,
+    /bin/busybox ixr,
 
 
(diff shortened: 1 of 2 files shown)


ccojocar (approved):

ccojocar:
@mhils If this doesn't pass could you please rebase it on main branch, we removed the image scanner for now because it seems that trivy rate limited their database image and is now causing a lot of timeout errors. 

https://github.com/aquasecurity/trivy-action/issues/389

saschagrunert:
Please rebase to fix the CI.

PR #2572, opened 2024-11-19
Improve the apparmor read-write file permissions to allow any file in a directory if already two files are allowed

ccojocar:
#### What type of PR is this?

/kind bug

It improves the apparmor read-write file permissions to allow any files if at least two files are allowed. There
are binaries like nginx which typically creates files with random name on every start in the `/etc/nginx/conf.d/` directory. The random named files cannot be captured in advance and allowed in the apparmor profile, therefore any file needs to be allowed.

#### Does this PR have test?

Yes

```release-note
NONE

```


saschagrunert:
@ccojocar do you mind a rebase?

ccojocar:
> do you mind a rebase?

Thanks, done it!

saschagrunert (approved):

Issue #2576, opened 2024-11-20
Apparmor recorder allows extra capabilities in the profile which doesn't seem to be required


ccojocar:
#### What happened:

The apparmor recorder apparently allows extra capabilities in the profile such as sys_admin and setpcap for nginx container when recording an apparmor profile. It seems that these capabilities are not required.

#### What you expected to happen:

Allow in the Apparmor profile only the minimum required capabilities. 

#### How to reproduce it (as minimally and precisely as possible):

Record an apparmor profile for nginx container image, and check the recorded capabilities in the corresponding CR.

#### Environment:

- Cloud provider or hardware configuration: GKE 
- OS (e.g: `cat /etc/os-release`): COS
- Kernel (e.g. `uname -a`):
- Others:

ccojocar:
cc @mhils

jindijamie:
Other than extra capabilities, it also includes extra paths. When I record the pod that runs sleep only. I got the following paths in recording
```
filesystem:
      readOnlyPaths:
      - /usr/bin/runc
      writeOnlyPaths:
      - /proc/@{pid}/attr/apparmor/exec
```

It looks like the recording records things happen when the container runtime starts the container as well.

PR #2587, opened 2024-11-29
Filter out unnecessary files from apparmor profile

ccojocar:
#### What type of PR is this?

/kind bug

Filter out unnecessary files from apparmor profile. These files are not required for the execution but they
are rather a side effect of recording the whole resources of a container runtime execution.

partial fix for #2576

#### Does this PR have test?

Yes

```release-note
Cleanup uncecessary files from a recorded apparmor profile.

```

cc @mhils


saschagrunert (approved):
Thanks!

PR #2598, opened 2024-12-04
Add a flag in ApparmorProfile CRD to switch to switch the profile into complain mode

### LINKED ISSUE #2559: Extend the AppArmor CRD with a profile flag which should allow to switch the profile mode

The AppArmor CRD should include a flag which defines the mode of the apparmor profile. This will allow to easily switch  a profile from complain to audit or enforcement mode. 

The list of supported profile flags is:
```
PROFILE FLAGS = 'complain' | 'audit' | 'enforce' | 'mediate_deleted' |
           'attach_disconnected' | 'chroot_relative'
```
source: https://manpages.ubuntu.com/manpages/bionic/en/man5/apparmor.d.5.html

Also some details are available in https://documentation.suse.com/sles/15-SP6/html/SLES-all/cha-apparmor-profiles.html#sec-apparmor-profiles-flags.

#### Why is this needed:

This is required in order to easily switch a profile from audit to enforcement or complain mode.

ccojocar:
#### What type of PR is this?

/kind feature

Adds a flag into `ApparmorProfile CRD` which allows to switch the profile mode from `enfoce` mode  to `complain` mode.

It the current implementation the `Profile Mode` flag is not specified in the apparmor profile file, which
places the profile by default into enforce mode. In this mode if a profile action is not allowed in the rules,
it will automatically terminate the container execution with an error. This is the expected behavior when running
productively but sometimes during the development is useful to put the profile in `complain` mode. In this mode,
if the profile rules don't grant permissions to an action, that action will be allowed, but the violation will
be logged with a tag of `ALLOWED unconfined`.

Fixes #2559

#### Does this PR have test?

Yes

```release-note
Add the complainMode flag into the ApparmorProfile CRD which allows to switch the apparmor profile into complain mode.

```

cc @mhils


saschagrunert (approved):

saschagrunert:
@ccojocar you'd have to rebase to fix the macos CI.

PR #2599, opened 2024-12-04
Remove file from known read files prefixes

ccojocar:
#### What type of PR is this?

/kind bug

The `/sys/kernel/mm/transparent_hugepage/hpage_pmd_size` file is not
present in the base apparmor pofile form COS
OS(https://cloud.google.com/container-optimized-os/docs) used by GKE. It
needs to be explicitly recorded.

```release-note
NONE
```

cc @mhils

### DIFF
--- a/internal/pkg/daemon/bpfrecorder/apparmor_filters.go
+++ b/internal/pkg/daemon/bpfrecorder/apparmor_filters.go
@@ -44,7 +44,6 @@ var knownReadPrefixes = []string{
 	"/etc/locale/",
 	"/etc/locale.alias/",
 	"/etc/localtime",
-	"/sys/kernel/mm/transparent_hugepage/hpage_pmd_size",
 	"/usr/share/locale/",
 	"/usr/lib/locale/",
 	"/etc/locale.alias",


saschagrunert (approved):

mhils:
Considering that it's in https://gitlab.com/apparmor/apparmor/-/blob/master/profiles/apparmor.d/abstractions/base#L110 and will eventually make its way downstream, should we consider to allowlist it [here](https://github.com/kubernetes-sigs/security-profiles-operator/blob/c20c5ead253583978dd392e5575583c7665caf55/internal/pkg/daemon/apparmorprofile/crd2armor/crd2armor.go#L31) instead?

ccojocar:
> Considering that it's in https://gitlab.com/apparmor/apparmor/-/blob/53e322b755d63573b3fbee6875b84692cfc19098/profiles/apparmor.d/abstractions/base#L113 and will eventually make its way downstream, should we consider to allowlist it [here](https://github.com/kubernetes-sigs/security-profiles-operator/blob/c20c5ead253583978dd392e5575583c7665caf55/internal/pkg/daemon/apparmorprofile/crd2armor/crd2armor.go#L31) instead?

I think is much more involved to enable it in COS base profile and to propagate it on all nods of GKE clusters. I don't see any harm in having it explicitly in the profile. This will enable some cases.

PR #2608, opened 2024-12-09
AppArmor Recorder: Add `mkdir` hook

mhils:
#### What type of PR is this?

/kind bug

This PR makes sure we properly record `mkdir` hooks.

None

#### Does this PR have test?

Yes.

```release-note
Improve AppArmor BPF recorder to capture mkdir syscalls.
```


internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -257,6 +257,61 @@ int BPF_PROG(bprm_check_security, struct linux_binprm * bprm)
...
+        pathlen + 0 < sizeof(event->data) &&
+        pathlen + 1 < sizeof(event->data) &&
+        pathlen + 2 < sizeof(event->data)
+    ) {
+        event->data[pathlen - 1] = '/';
+        event->data[pathlen - 0] = '*';
+        event->data[pathlen + 1] = '*';
+        event->data[pathlen + 2] = '\0';
+    } else {
+        bpf_printk("failed to fixup directory entry, not enough space: %s", event->data);
```
ccojocar, line 295: This case is realistic only when the path exceeds the `PATH_MAX` size `4096` characters. Right?

I am wondering if it makes sense to overwrite the last last directory in the path with `**` in that case, but this is almost unrealistically to occur given the length of the path.

nit: maybe we should mention this in the error message.

ccojocar (approved):

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
mhils: > This case is realistic only when the path exceeds the PATH_MAX size 4096 characters. Right?

Yes, correct - I'd say we look at workarounds iff we ever hit this. 
I've expanded the error message and added a comment. :)

ccojocar: Perfect, thanks!

mhils:
It looks like `bpf_d_path` is not universally alllowlisted for `lsm/path_mkdir`. Closing this in favor of #2611, which addresses this already for other syscalls.

PR #2611, opened 2024-12-10
Profile Recorder: Add `bpf_d_path` implementation and hook mkdir, mknod, unlink.

mhils:
#### What type of PR is this?

/kind feature

This PR adds a custom `bpf_d_path` implementation that can be used in eBPF hooks that are not allowlisted to actually use `bpf_d_path`. We then use this to hook `mkdir`, `mknod`, and `unlink` to create more comprehensive AppArmor profiles.

None

#### Does this PR have test?

Yes

supersedes #2608

```release-note
Profile recorder: Hook `mkdir`, `mknod`, and `unlink` for better AppArmor profiles
```


ccojocar:
It seems that the entire overlay filesystem is added to the profile https://github.com/kubernetes-sigs/security-profiles-operator/actions/runs/12261338471/job/34208338061?pr=2611#step:8:170.  

We need to filter out all these file permissions since is not required in the apaprmor profile.

ccojocar:
I think we need to filter out the bootstraping and init phases of the container. 

As far as I know the `nsenter` (https://github.com/opencontainers/runc/tree/main/libcontainer/nsenter) gets executed before `runc init` which executes itself and then starts the `container init process`. I believe that we should not record the resources before container init process.  

I am think maybe is possible to filter out all resource associated with runc command. There is already this filter for syscalls https://github.com/kubernetes-sigs/security-profiles-operator/blob/3cb35b9bc6c75a8452b5fbdc0c2fe8147040040d/internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c#L124.

mhils:
Split up into 
https://github.com/kubernetes-sigs/security-profiles-operator/pull/2663
https://github.com/kubernetes-sigs/security-profiles-operator/pull/2667
https://github.com/kubernetes-sigs/security-profiles-operator/pull/2668

PR #2623, opened 2024-12-13
BPF Recorder: Exclude Container Initialization from Recorded Profile

### LINKED ISSUE #2576: Apparmor recorder allows extra capabilities in the profile which doesn't seem to be required

mhils:
#### What type of PR is this?

/kind feature

This PR makes us 1) detect when containers are (about to be) initialized, and 2) removes that phase from the recorded profile.

Fixes #2576

#### Does this PR have test?

Covered by existing E2E tests.

99% sure that some kernels won't support `sched_prepare_exec`, but let's see if it works in the first place before we revise.

```release-note
The BPF profile recorder now excludes unnecessary permissions exercised during container init.
```


internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -322,6 +326,43 @@ int BPF_KPROBE(cap_capable)
...
+    }
+}
+
+SEC("tracepoint/sched/sched_prepare_exec")
+int sched_prepare_exec(struct trace_event_raw_sched_process_exec * ctx)
+{
+    bpf_printk("sched_prepare_exec");
+    struct task_struct * task = (struct task_struct *)bpf_get_current_task();
+    u32 mntns2 = BPF_CORE_READ(task, nsproxy, mnt_ns, ns.inum);
+    //clear_mntns(mntns2);
```
ccojocar, line 361: Why is this commented out?

ccojocar: Is it because you are trying the alternative solution?

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -203,6 +203,27 @@ func (b *AppArmorRecorder) handleCapabilityEvent(capEvent *bpfEvent) {
 	b.recordedCapabilities[mid] = append(b.recordedCapabilities[mid], requestedCap)
 }
 
+// Delete all data recorded for a particular mount namespace.
+//
+// The recorder triggers this after container initialization to make sure that
+// permissions needed for setup are not included in the final profile.
+func (b *AppArmorRecorder) clearMntns(event *bpfEvent) {
+	// Everything we captured so far is container initialization.
```
ccojocar, line 211: nit: This comment is not so relevant if this function gets called from a different context.

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -203,6 +203,27 @@ func (b *AppArmorRecorder) handleCapabilityEvent(capEvent *bpfEvent) {
...
+	// Let's get rid of it.
+	b.lockRecordedFiles.Lock()
+	defer b.lockRecordedFiles.Unlock()
+	b.lockRecordedCapabilities.Lock()
+	defer b.lockRecordedCapabilities.Unlock()
+	b.lockRecordedSocketsUse.Lock()
+	defer b.lockRecordedSocketsUse.Unlock()
+
+	mntns := mntnsID(event.Mntns)
+	log.Printf("Clearing mntns: %d\n", mntns)
```
ccojocar, line 221: Does the syscalls maps doesn't need any clean-up? Is it already cleaned in the eBPF code?

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -93,6 +96,7 @@ typedef struct __attribute__((__packed__)) event_data {
 const volatile char filter_name[MAX_COMM_LEN] = {};
 const volatile u32 exclude_mntns = 0;
 
+static const char RUNC_DONE[] = "runc:[2:INIT]";
```
ccojocar, line 99: This will lock us only to runc. Other alternatives like crun which I think is used by crio-o, it will not be covered. It is worth to mention this somewhere into the docs when this gets merged.

ccojocar (approved):

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
ccojocar: @saschagrunert do we have tests to cover crun for seccomp? I think this won't affect the number of syscalls recorded in that case.

saschagrunert: Yes the e2e tests for seccomp should use runc and crun.

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
mhils: Yes, correct. `sched_prepare_exec` is very new (https://github.com/torvalds/linux/commit/c82389947d90c8b0dd206d9ef082191eb58df8fc), so I am experimenting with `sched_process_exit` instead. That doesn't fully work yet, I'm working on fixing it.

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
mhils: Already cleaned in the eBPF code.

ccojocar (approved):

ccojocar:
It is an impressive reduction in permissions in both apapmror and secocmp. 🤞 that this works!

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
```diff
@@ -51,6 +53,9 @@ char LICENSE[] SEC("license") = "Dual BSD/GPL";
 #define unlikely(x) __builtin_expect((x), 0)
 #endif
 
+#define trace_hook(...)
```
ccojocar, line 56: nit: this look like an empty macro, I would remove it if it is not used.

ccojocar (approved):

internal/pkg/daemon/bpfrecorder/bpf/recorder.bpf.c
mhils: The intent here is that one can quickly toggle between no-op and bpf_printk, which makes debugging BPF much nicer. Does that make sense, any better suggestions?

ccojocar (approved):

PR #2632, opened 2024-12-17
Support both runtimes runc and crun for apparmor e2e test

ccojocar:
#### What type of PR is this?

/kind bug

- Extend the apparmor E2E tests to validate the profile recording and installation against both runtimes crun and runc, the same like seccomp E2E tests.
- Switch the seccomp E2E test also on Debian (in order to run all e2e scripts (spoc, seccomp and apparmor) on the same Linux distribution.

```release-note
NONE
```


saschagrunert (approved):
Thanks!

PR #2658, opened 2025-01-07
Cleanup the apparmor recorder maps after processing

ccojocar:
#### What type of PR is this?

/kind bug

Cleans up the apparmor recorder maps after processing.

#### Does this PR have test?

Yes

```release-note
NONE
```

### DIFF
--- a/hack/ci/e2e-apparmor.sh
+++ b/hack/ci/e2e-apparmor.sh
@@ -25,7 +25,7 @@ APPARMOR_REFERENCE_TMP_PROFILE_FILE="/tmp/apparmorprofile-sleep-reference.yaml"
 APPARMOR_PROFILE_FILE_COMPLAIN_MODE="examples/apparmorprofile-sleep-complain-mode.yaml"
 SLEEP_INTERVAL_RECORDING="30"     # 30s sleep interval during recording.
 SLEEP_INTERVAL_VERIFICATION="300" # 5min to make sure that the enforcement check finds a running  PID.
-RUNTIMES=(runc crun)
+RUNTIMES=(crun runc)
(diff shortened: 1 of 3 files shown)


ccojocar:
cc @mhils

mhils (commented):
```
func (b *AppArmorRecorder) Unload() {
}
```

in `bpfrecorder_apparmor.go` needs to be fixed as well.

ccojocar:
> in bpfrecorder_apparmor.go needs to be fixed as well.

It seems that that closing the module will destroy all the links and also unload the programs from the kernel.  https://github.com/kubernetes-sigs/security-profiles-operator/blob/b2742f128d71f98dbff58bf4788999163bd0c08f/internal/pkg/daemon/bpfrecorder/bpfrecorder.go#L887

This is what actually happens in module close https://github.com/kubernetes-sigs/security-profiles-operator/blob/b2742f128d71f98dbff58bf4788999163bd0c08f/vendor/github.com/aquasecurity/libbpfgo/module.go#L183

I'm not sure if we need to do anything extra in the AppArmorRecorder `Unload`.

mhils:
I just saw that the seccomp recorder is cleaning itself up in Unload and the AppArmor recorder isn't, which kind of fits the theme of this PR. Not sure if the userspace datastructures survive. :)

ccojocar:
> I just saw that the seccomp recorder is cleaning itself up in Unload and the AppArmor recorder isn't, which kind of fits the theme of this PR. Not sure if the userspace datastructures survive. :)

Got it! I cleaned up all the internal maps in the Unload.

ccojocar:
Close in favour of #/2675.

PR #2675, opened 2025-01-15
BPF Recorder: load bpf program on startup to win race with container startup

mhils:
#### What type of PR is this?

/kind flake

This commit makes it so that the BPF program is loaded (and relocated) on profile recorder startup. When we then want to start the recording, we only need to attach the programs. This fixes the flakiness in #2667 and #2668 where program loading is taking so long that we miss container startup.

None

#### Does this PR have test?

N/A

```release-note
NONE
```

### DIFF
--- a/examples/apparmorprofile-sleep-crun.yaml
+++ b/examples/apparmorprofile-sleep-crun.yaml
@@ -16,7 +16,6 @@ spec:
         - mknod
         - setgid
         - setpcap
-        - setuid
         - sys_admin
     executable:
       allowedExecutables:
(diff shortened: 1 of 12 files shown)


ccojocar (approved):

ccojocar:
@mhils I think you need to regenerate the mocks with `make update-mocks`.

ccojocar:
Also there are some linting warnings which needs to be fixed. Just run `make verify`.

mhils:
> Also there are some linting warnings which needs to be fixed. Just run `make verify`.

Thanks. Tests still need fixing altogether, I wanted to push this early to see if it deflakes the apparmor and seccomp e2e tests.

ccojocar (approved):

PR #2686, opened 2025-01-22
Add more metrics for AppArmor Profile

### LINKED ISSUE #2665: More Metrics for AppArmor Profile

- apparmor_profile_audit_total is defined but the function IncAppArmorProfileAudit is never called, the underlying implementation should be added.
- apparmor_profile_error_total counts errors on apparmor profile installation, the only field is error reason, we want an extra field on the name of the profile so that we can monitoring which profile goes wrong.
- We also want something like apparmor_profile_denial_total that counts the operations being denied.

#### Why is this needed:
To monitoring potential failures because of missing entries in the profiles and profile installation failures.

ccojocar:
#### What type of PR is this?

/kind feature

- Implement the apparmor_profile_audit_total metric
- Add the profile name to apparmor_profile_error_total metric
- Use a metric label name instead of reason inside of apparmor_profile_audit_total
- Add a metric for apparmor profile denail

Fixes #2665

#### Does this PR have test?

N/A

```release-note
Add more metrics for AppArmor profile.

```


ccojocar:
@saschagrunert Please could you have a quick look? Thanks a lot!

saschagrunert (approved):

PR #2694, opened 2025-01-24
Changed AppArmorProfile and SecCompProfile scope

### LINKED ISSUE #2582: ApparmorProfile CRD should be a cluster-wide resource

The  ApparmorProfile CRD should be cluster-wide scope, and not namespaced. I would argue that also SeccompProfile should be the same.

The Profile CRDs  are currently namespaced and there are a number of issues with this:
- apparmor profile during the installation, it gets loaded into the Linux kernel of each ndoe, and it is cluster-wide available, any pod can reference it into the security context. 
- There isn't any mechanism in the security context to allow name-spacing the secruity profiles.
- It is possible currently to overwrite a profile from one namesapce with another profile created in a different namesapce if both profiles have the same name. This can be used to preform cross namespaces attacks. For instance a less permissive profile gets overwritten by a profile with more privileges from a different namespace even though the profile creator doesn't have permissions to do so in the original namesspace.   

#### Why is this needed:

This will enhance the security and make it more transparent to RBAC policies that actually this profiles are cluster-wide.

anshuman-agarwala:
#### What type of PR is this?

/kind api-change

Fixes #2582

```release-note

```

### DIFF
--- a/bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
+++ b/bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
@@ -1,5 +1,142 @@
 apiVersion: v1
 data:
+  bpf-recorder.json: |
+    {
+      "defaultAction": "SCMP_ACT_ERRNO",
+      "architectures": [
+        "SCMP_ARCH_X86_64",
+        "SCMP_ARCH_X86",
+        "SCMP_ARCH_X32",
+        "SCMP_ARCH_AARCH64"
+      ],
+      "syscalls": [
+        {
+          "names": [
+            "accept4",
+            "arch_prctl",
+            "bind",
+            "bpf",
+            "brk",
+            "capget",
+            "capset",
+            "chdir",
+            "clone3",
+            "close",
+            "connect",
+            "dup3",
+            "epoll_create1",
+            "epoll_ctl",
+            "epoll_pwait",
+            "epoll_wait",
+            "eventfd2",
+            "execve",
+            "faccessat2",
+            "fchownat",
+            "fcntl",
+            "fstat",
+            "fstatfs",
+            "futex",
+            "getcwd",
+            "getdents64",
+            "getpeername",
+            "getpid",
+            "getppid",
+            "getrandom",
+            "getrlimit",
+            "getsockname",
+            "getsockopt",
+            "gettid",
+            "ioctl",
+            "listen",
+            "lseek",
+            "madvise",
+            "memfd_create",
+            "mmap",
+            "mprotect",
+            "mremap",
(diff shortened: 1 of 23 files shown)


bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 data:
-  security-profiles-operator.json: |
+  bpf-recorder.json: |
```
ccojocar, line 3: Please could you update with latest policies from master?

ccojocar (commented):
I think the RBAC needs to be also updated.

bundle/manifests/security-profiles-operator-profile_v1_configmap.yaml
anshuman-agarwala: Rebased with master, is there anything else I need to do?

anshuman-agarwala:
> I think the RBAC needs to be also updated.

I've looked through all the Role manifests, not sure which ones need updating.
Also for the E2E errors, when I deploy the operator locally I get the following:
```
E0201 14:55:54.892359   10551 controller.go:332] "Reconciler error" err="cannot ensure node status: cannot create node status for log-enricher-trace: creating node status: an empty namespace may not be set during creation" controller="profile" controllerGroup="security-profiles-operator.x-k8s.io" controllerKind="SeccompProfile" SeccompProfile="log-enricher-trace" namespace="" name="log-enricher-trace" reconcileID="e956f8ea-57c7-4211-be7a-88d40edea885"
```
Cannot find the source of this error, I know it's due to the scope change since it's trying to set the namespace for a clustered resource but I cannot figure out where.

ccojocar:
> Cannot find the source of this error, I know it's due to the scope change since it's trying to set the namespace for a clustered resource but I cannot figure out where.

I think the NodeStatus is namespaced, and it gets a reference to the security base profile here 
https://github.com/kubernetes-sigs/security-profiles-operator/blob/efc45d4032dd1d237006d3399a09007695c6a102/internal/pkg/nodestatus/nodestatus.go#L56. 

It is created for both seccomp and apparmor, for instance here https://github.com/kubernetes-sigs/security-profiles-operator/blob/efc45d4032dd1d237006d3399a09007695c6a102/internal/pkg/daemon/seccompprofile/seccompprofile.go#L463. 

I believe you need to set up its namesapce explicitly in the operator namesapce since now the security profiles are cluster scoped.

ccojocar:
Closing this in favour of https://github.com/kubernetes-sigs/security-profiles-operator/pull/2735.

PR #2646, opened 2024-12-23
Harden the spo and bpf-recorder containers with custom apparmor profiles

ccojocar:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This pull request harden the security-profiles-operator and bpf-recorder containers as part of spod daemonset with custom apparmor profiles when apparmor is enabled.
These two containers run in privileged mode when the apparmor is activated.

#### Which issue(s) this PR fixes:

Fixes #65

#### Does this PR have test?

Yes

#### Does this PR introduce a user-facing change?

```release-note
Harden the security-profiles-operator and bpf-recorder containers with custom apparmor profiles when apparmor is enabled.

```

PR #2696, opened 2025-01-24
fix apparmor profile for spod

mhils:
#### What type of PR is this?

/kind bug

This PR fixes the apparmor profile for the bpf recorder itself so that recording works again after #2646. The template injection hack is not meant to be a permanent solution, but a temporary stopgap to make things work without reverts.

Most of the missing permissions were found looking at `journalctl`, but for `/` AppArmor did not emit any DENIED messages despite actually denying permissions. This is a bit worrying in general, but what to do.

None

#### Does this PR have test?

N/A

```release-note
NONE
```

### DIFF
--- a/deploy/base/profiles/bpfrecorder-apparmor.yaml
+++ b/deploy/base/profiles/bpfrecorder-apparmor.yaml
@@ -23,21 +23,15 @@ spec:
       - /proc/sys/net/core/somaxconn
       - /sys/devices/kprobe/type
       - /sys/devices/system/cpu/online
-      - /sys/fs/bpf/**
+      - /sys/fs/bpf/
       - /sys/kernel/btf/vmlinux
-      - /sys/kernel/debug/tracing/events/raw_syscalls/sys_enter/id
-      - /sys/kernel/debug/tracing/events/sched/sched_process_exec/id
-      - /sys/kernel/debug/tracing/events/sched/sched_process_exit/id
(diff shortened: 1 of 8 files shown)


saschagrunert (approved):

ccojocar (approved):

ccojocar:
The ptrace stuff is requried in addition to #2691. Did we find out what introduced the ptrace thingy? I think was not required before.

ccojocar:
@mhils this one needs a rebase.

mhils:
> Did we find out what introduced the ptrace thingy? I think was not required before.

Based on https://gitlab.com/apparmor/apparmor/-/wikis/TechnicalDoc_Proc_and_ptrace#list-of-proc-files-that-may-cause-ptrace-trace-requests, AppArmor wants ptrace permissions if `/proc/<pid>/ns/` is read (which is in both profiles). So the ptrace permission is not only about a `ptrace()` syscall, but also for certain file access.

/proc/1/ns is read by go-apparmor here: https://github.com/pjbgf/go-apparmor/blob/1375e5e7aa892edd0ad2a1cebef2297ac2e3872d/pkg/hostop/mount_linux.go#L86. So my explanation for why this did not show up before is that the previous cluster where we recorded the profile did not have apparmor recording enabled.

ccojocar (approved):

PR #2713, opened 2025-02-04
AppArmor: Do not deny `mount`

mhils:
#### What type of PR is this?

/kind bug

`crd2armor` previously inserted a `deny mount` AppArmor rule by default, but with `deny` taking precedence over everything else this breaks every setup that does require mount. This PR unbreaks it and adds a bit of a hack to permit `mount` by default for sys_rawio containers until we have proper mount recording.

None

#### Does this PR have test?

Tested locally.

```release-note
Permit AppArmor profiles with `cap_sys_rawio` to call `(u)mount`.
```


saschagrunert (approved):
/ok-to-test

ccojocar (approved):

PR #2735, opened 2025-02-20
Change the security profiles CRDs scope to be cluster wide

### LINKED ISSUE #2582: ApparmorProfile CRD should be a cluster-wide resource

ccojocar:
#### What type of PR is this?

/kind feature

Change the scope of security profiles CRDs to be cluster wide.

Fixes #2582

#### Does this PR have test?

Yes

```release-note
Change the scope of security profiles CRDs to be cluster wide.
```


saschagrunert (approved):

ccojocar:
@saschagrunert Please could you have another look? All tests now passed. Thanks

saschagrunert (approved):

PR #2782, opened 2025-03-25
Add AppArmorProfile to security-profiles-operator role

Edan-Hamilton:
#### What type of PR is this?

/kind bug

Fixes a bug where the operator is unable to list AppArmorProfiles.

None

#### Does this PR have test?

N/A

```release-note
NONE
```


ccojocar:
@Edan-Hamilton  I think you need to regenerate the deployment manifests, there is a drift in the configuration.

PR #2784, opened 2025-03-25
Add extra field to Abstract AppArmorProfile spec

### LINKED ISSUE #2678: Breaking changes to AppArmorProfile spec
The recent decision to remove `policy` from AppArmorProfiles breaks existing deployed resources. Changes to the schema should be rolled out as a new API Version so that existing resources are unaffected.

Further, it is unclear how complex policies are to be implemented now that the `policy` field is not available. It does not appear that `abstract` can handle exclusion, there is no clear way to limit signals/ptrace, and while I suspect it supports wildcards there is no documentation so even that is not certain. Much of this is easy to solve, but removing `policy` while `abstract` is still clearly so early in development seems preemptive.

Edan-Hamilton:
#### What type of PR is this?

/kind api-change

Adds an "extra" field to the Abstract spec, which allows including additional AppArmor parameters. This is important for handling edge-cases not (yet) covered by the spec.

Fixes #2678

#### Does this PR have test?

N/A

```release-note
Added "extra" field to Abstract AppArmorProfiles
```


saschagrunert:
CI does not seem to be happy about this:
```
Error: api/apparmorprofile/v1alpha1/apparmorprofile_types.go:85:1: File is not properly formatted (gci)
	Extra string `json:"extra,omitempty"`
^
```

ccojocar:
@Edan-Hamilton Please could you fix the build failure? I think you need to regenerate the deployment manifests with `make deployments`. Thanks

PR #2807, opened 2025-04-09
Fix the logging at the given verbosity

ngopalak-redhat:
#### What type of PR is this?
/kind bug

Sets the configured verbosity to the logr.
Currently logging with verbosity:
```
e.logger.V(config.VerboseLevel).Info("Got line: " + line)
```
does not produce any log in the output when configured as described here https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/installation-usage.md#set-logging-verbosity

The fix addresses this issue by setting the verbosity given by the user. The default logging level is not changed.

```release-note

```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -283,7 +283,8 @@ func initialize(ctx *cli.Context) error {
 }
 
 func initLogging(ctx *cli.Context) error {
-	ctrl.SetLogger(textlogger.NewLogger(textlogger.NewConfig()))
+	logConfig := textlogger.NewConfig()
+	ctrl.SetLogger(textlogger.NewLogger(logConfig))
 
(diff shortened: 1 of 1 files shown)


saschagrunert:
Thank you!

saschagrunert:
@ngopalak-redhat you need to fix the linter CI:
```
level=debug msg="[gocritic] All checks are enabled"
cmd/security-profiles-operator/main.go:298:38: integer-format: fmt.Sprintf can be replaced with faster strconv.FormatUint (perfsprint)
	if err := logConfig.Verbosity().Set(fmt.Sprintf("%d", level)); err != nil {
	                                    ^
cmd/security-profiles-operator/main.go:301:2: expressions should not be cuddled with blocks (wsl)
	ctrl.Log.Info(fmt.Sprintf("Set logging verbosity to %d", level))
	^
```

PR #2808, opened 2025-04-09
Fix the logging at the given verbosity

ngopalak-redhat:
#### What type of PR is this?
/kind bug

Sets the configured verbosity to the logr.
Currently logging with verbosity:
```
e.logger.V(config.VerboseLevel).Info("Got line: " + line)
```
does not produce any log in the output when configured as described here https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/installation-usage.md#set-logging-verbosity

The fix addresses this issue by setting the verbosity given by the user. The default logging level is not changed.
NONE

```release-note
NONE
```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -298,7 +298,7 @@ func initLogging(ctx *cli.Context) error {
 	ctrl.SetLogger(ctrl.Log.V(int(level)))
 
 	if err := logConfig.Verbosity().Set(strconv.FormatUint(uint64(level), 10)); err != nil {
-		return fmt.Errorf("setting the verbosity flag to level %d : %w", level, err)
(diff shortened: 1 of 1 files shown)


cmd/security-profiles-operator/main.go
```diff
@@ -294,6 +296,11 @@ func initLogging(ctx *cli.Context) error {
 	}
 
 	ctrl.SetLogger(ctrl.Log.V(int(level)))
+
+	if err := logConfig.Verbosity().Set(strconv.FormatUint(uint64(level), 10)); err != nil {
+		return fmt.Errorf("setting the verbosity flag to level %d : %w", level, err)
```
saschagrunert, line 301:
```suggestion
		return fmt.Errorf("setting the verbosity flag to level %d: %w", level, err)
```

saschagrunert (approved):

PR #2826, opened 2025-04-22
Apply the seccomp and selinux profiles only there are actual changes

ccojocar:
#### What type of PR is this?

/kind bug

Applies the changes to Seccomp and Selinux profiles only when there are actual changes in the CRs.

This is a follow up for #2822 which fixes Apparmor profile.

None

#### Does this PR have test?

Yes

```release-note
Applies the changes to Seccomp and Apparmor profiles only whent here are effective changes in the CRs.

```


saschagrunert (approved):

ccojocar:
It seems that this change triggers some test failure which I need to debug.

ccojocar:
@saschagrunert it somehow passed now. It just needs a lgtm. Thanks

saschagrunert (approved):

PR #2835, opened 2025-05-05
2817: Add In-Pod Activity Log recorder in Audit JSON lines format

### LINKED ISSUE #2817: Add In-Pod Activity Audit Recorder
Introduce a new audit logging recorder to capture process execution activity within Pods/Containers.

This new recorder should:

1. *Output Format*: Generate audit records as JSON lines, structurally similar to the existing [Kubernetes API server audit logs](https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/) for easier integration with standard log collection systems.

2. *Log Information*: Each record should detail executed process, including:

    - Executable name (e.g., `/bin/bash`, `redis`, etc)
    - Command-line arguments (`Cmdline`)
    - UID under which the process ran.
    - syscalls (if configured)

*Implementation approach*:

Since this operator by design tails the generated audit logs, the idea here is to configure `SeccompProfile` such that only required amount of log is generated. For example:
```
apiVersion: security-profiles-operator.x-k8s.io/v1beta1
kind: SeccompProfile
metadata:
  namespace: default
  name: profile1
spec:
  defaultAction: SCMP_ACT_ALLOW
  sys calls:
  - action: SCMP_ACT_LOG
    names:
    - execve
    - clone
```
This will allow the new recorder to receive log lines when processes are forked or commands are executed by a user inside the pod/container. 

The recorder can then aggregate the above mentioned log information within a time window and produce audit log.

Also provide an option to log all syscalls made within the window if detailed logging is needed at the cost of performance.

*Filtering*: Allow filtering of audit records based on the executable name to help manage log volume in environments with many nodes or applications with many executables.

*Configuration*: Aspects like the aggregation time window and syscall/executable filters should be configurable.

A sample JSON line produced can look like this in the log:
```
{
	"version": "spo/v1_alpha",
	"auditID": "a6029022-4ff0-4c54-97ed-4099d0ca1923",
	"executable": "/bin/dash",
	"cmdLine": "chmod 777 docker-entrypoint.sh",
	"pid": 70714,
	"uid": 0,
	"gid": 0,
	"resource": {
		"pod": "my-pod-debug-v9mfp",
		"namespace": "default",
		"container": "nginx",
	},
	"node": {
		"name": "ci-ln-31bf5rk-72292-t4x4k-worker-c-ppx72",
		"sourceIPs": ["172.31.22.88"],
	}
	"syscalls": [
		"execve"
	]
	"timestamp": "2025-04-11T08:36:48.679291Z"
}
```
Note that the deployed workloads will have to use this Seccomp profile.

#### Why is this needed:

Standard Kubernetes audit logs capture API interactions but lack visibility into activities inside pods/containers. This includes commands run via `kubectl exec`, `kubectl debug`, scripts executed within a container's lifecycle, background processes, and the specific system calls made by applications.

There are hacks/wrappers available on the kubectl client that help record the commands executed.  Also current version of this operator has log enricher which generates syscall information. These are not sufficient as audit logging end-users need user activity in JSON lines format that can interleave into the Kubernetes audit logs.

The feature will drive the adoption of this operator in production systems that need audit of the SRE activity.

ngopalak-redhat:
#### What type of PR is this?
/kind feature

Adds Audit JSON ability to security profiles operator. Outputs JSON lines for:
- User activity done inside "kubectl exec"
- Commands executed by the container itself

Usage is described in installation.md file of this PR.

Note:
- The feature currently only works if Seccomp profile can be enabled for this pod and the pod is not using privileged containers.
- Currently supports writing the JSON lines log at 1 minute interval to standard output. It can be made configurable later

#### Does this PR have test?
Yes. Its added in the PR itself and run on the Fedora and Ubuntu OS

This is the first implementation for the feature. Once this is merged we can add features to make it more configurable. Sample output is available here: https://gist.github.com/ngopalak-redhat/f5552937e8d261493aa0bd8cb43640ee

```release-note
Add In-Pod Activity Log recorder in audit JSON lines format
```


installation-usage.md
```diff
@@ -847,6 +848,98 @@ Note that a security profile that is in use by existing pods cannot be
...
+   
+   Create the pod using kubectl apply:
+   ```shell
+   kubectl apply -f my-pod.yaml
+   ```
+5. **Monitor the Audit Logs:**
+   
+   To monitor the audit log tail:
+   ```shell
+   oc -n security-profiles-operator logs --since=1m --selector name=spod -c json-enricher --max-log-requests 6 -f
```
ccojocar, line 938: nit: kubectl

internal/pkg/manager/spod/bindata/spod.go
```diff
@@ -615,6 +617,58 @@ semodule -i /opt/spo-profiles/selinuxrecording.cil
...
+							},
+							{
+								Name:      "host-syslog-volume",
+								MountPath: filepath.Dir(config.SyslogLogPath),
+								ReadOnly:  true,
+							},
+						},
+						SecurityContext: &corev1.SecurityContext{
+							ReadOnlyRootFilesystem: &truly,
+							Privileged:             &truly,
```
ccojocar, line 638: Does this container need to run as `privileged`?

installation-usage.md
```diff
@@ -847,6 +848,98 @@ Note that a security profile that is in use by existing pods cannot be
...
+- Tracking when someone uses commands like kubectl exec to get into a running container and run commands or scripts.
+- Monitoring activities in debug containers where users might run various tools.
+
+Currently, this feature is hardcoded to generate log for a process at 60 second interval. Also the JSON lines are output 
+to the standard output stream.
+
+To start using this feature, you need to have the Security Profiles Operator installed in your Kubernetes cluster. 
+Once it's installed, you can enable the JSON log enricher with this command:
+```sh
+oc -n security-profiles-operator patch spod spod --type=merge -p '{"spec":{"enableJsonEnricher":true,"verbosity":0}}'
```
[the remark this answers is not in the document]
ngopalak-redhat: Thanks fixed

internal/pkg/manager/spod/bindata/spod.go
ngopalak-redhat: Fixed and tested

ccojocar:
@ngopalak-redhat It's a lgtm form my side. Thanks for contributing this. I can approve it if is ready to be merged.

installation-usage.md
```diff
@@ -847,6 +848,98 @@ Note that a security profile that is in use by existing pods cannot be
...
+- **Command Line Arguments (cmdline)**: The extra instructions given when the program was started (e.g., ls -l /home).
+- **User and Group IDs (uid/gid)**: The identification numbers of the system user who ran the program.
+- **System Calls (syscalls)**: A list of system calls (syscalls) that the process made
+
+This log format is very similar to how Kubernetes itself records audit logs. This is useful for:
+- Seeing what users and automated processes are doing inside a pod.
+- Tracking when someone uses commands like kubectl exec to get into a running container and run commands or scripts.
+- Monitoring activities in debug containers where users might run various tools.
+
+Currently, this feature is hardcoded to generate log for a process at 60 second interval. Also the JSON lines are output
```
haircommander, line 866: nit: `to generate logs for a process at an interval of 60 seconds`

deploy/helm/values.yaml
```diff
@@ -26,6 +26,7 @@ selinuxdImage:
 
 enableSelinux: false
 enableLogEnricher: false
+enableJsonEnricher: false
```
haircommander, line 29: if there's going to be a configurable log interval period in the future, it almost makes sense to me to have this be a pointer to a jsonEnricher object that has an interval embedded. if the object is nil, no enrichment. if the interval is unset (0) then it defaults to 60 seconds, otherwise it uses the value of the interval

internal/pkg/daemon/enricher/impl.go
```diff
@@ -165,3 +172,41 @@ func (d *defaultImpl) Stat(name string) (os.FileInfo, error) {
...
+
+	file, err := os.Open(filepath.Clean(cmdline))
+	if err != nil {
+		return "", fmt.Errorf("%w: %w", ErrProcessNotFound, err)
+	}
+
+	defer func() {
+		cerr := file.Close()
+		if err == nil {
+			err = cerr
```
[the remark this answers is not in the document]
haircommander: follow on: I like using a different name like `retErr` instead of `err` because weird scoping bugs can result in unexpected values. then, always set `err` and have `retErr` be set by the return statement, and reference it in the `defer`

ngopalak-redhat:
@haircommander / @ccojocar  I have addressed the review comments. Once this is merged, I will create a issue/PR for:
1. Making the audit log interval configurable.
2. Add an option to redirect the log output to a file rather than the standard output. Some observability systems like files instead of standard output
3. Raise a feature requirement to get human user information into the log output. Currently system user name is logged. This requires several changes which I'll explain in the feature.
All these are enhancements.

api/spod/v1alpha1/spod_types.go
```diff
@@ -276,6 +284,12 @@ type SPODSpec struct {
 	// tells the operator whether or not to enable log enrichment support for this
 	// SPOD instance.
 	EnableLogEnricher bool `json:"enableLogEnricher,omitempty"`
+	// tells the operator whether or not to enable audit JSON enrichment support for this
+	// SPOD instance.
+	EnableJsonEnricher bool `json:"enableJsonEnricher,omitempty"`
+	// Defines options specific to the JsonEnricher
+	// functionality of the SecurityProfilesOperator
+	JsonEnricherOpts JsonEnricherOptions `json:"jsonEnricherOptions,omitempty"`
```
haircommander, line 292: IMO this should be a pointer to JsonEnricherOptions and then we don't need the enablement bool
I also think JSON would be better? though that may have it serialize oddly?

ngopalak-redhat: - I made `JsonEnricherOpts` as a pointer
- `EnableJsonEnricher` bool is required to launch this container. Hence its retained
- "JSON" in the variable name is not camel cased and from reading several docs its recommended to keep CamelCase variables. Esp when serialization is involved.

ngopalak-redhat: @haircommander `EnableJsonEnricher` works similar to `EnableLogEnricher`. There is a e2e test that's using it: `https://github.com/kubernetes-sigs/security-profiles-operator/blob/ecdd64fc47e88dd5fe0faf5937517c2b9d34dd83/test/suite_test.go#L782` . Hence its required

ccojocar (approved):

PR #2867, opened 2025-06-02
JSON Audit logging flush interval configuration

ngopalak-redhat:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
The PR introduces an option to set the audit log interval for the JSON Enricher. Within this interval the syscalls are collated per process id.

#### Does this PR have test?
Yes

#### Does this PR introduce a user-facing change?

```release-note
Adds an option (`--audit-log-interval-seconds`) to set the audit log interval for the JSON Log Enricher
```

PR #2871, opened 2025-06-05
Add file system support for JSON Log Enricher Audit logging

ngopalak-redhat:
#### What type of PR is this?
/kind feature

This PR introduces file system as a destination for the JSON Log Enricher. Detailed usage instructions are in installation-usage.md in this PR.

- Integrates lumberjack for log rotation (path, size, age, backups), similar to kube-apiserver's audit logging.

- Security Profiles Operator ConfigMap is used for volume mount input, giving users the freedom to specify any volume type for audit logs, just like kube-apiserver's audit backend.

This addresses a key operational need for manageable audit logging maintaining the current JSON lines format as-is.

This PR advances the JSON Log Enricher by aligning its audit logging capabilities with kube-apiserver's configuration.

Currently, audit logs default to standard output (stdout), which often requires distributed logging systems to continuously tail that output. By enabling a file system as the logging backend, logs can be stored directly and persistently based on requirements, eliminating the need for constant tailing.

This change simplifies audit log management, making it easier for customers to gather and centralize these audit records, and enhances SPO's integration into existing audit logging infrastructures.

#### Does this PR have test?
Yes, both e2e and unit tests are added

Reference: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apiserver/pkg/server/options/audit.go#L509
Addressed @ccojocar comment: https://github.com/kubernetes-sigs/security-profiles-operator/pull/2867#discussion_r2124026232

```release-note
Add file system support for JSON Log Enricher Audit logging
```


api/spod/v1alpha1/spod_types.go
```diff
@@ -246,6 +246,28 @@ type JsonEnricherOptions struct {
 	// within this interval are grouped together. The default is 60 seconds.
 	// Increasing this interval will reduce the rate at which logs are written.
 	AuditLogIntervalSeconds int32 `json:"auditLogIntervalSeconds,omitempty"`
+
+	// This field specifies the path for the accumulated audit log data.
+	// The audit log will be written to this file in JSON format if a file path
+	// is provided. If left unspecified, the output will be directed to
+	// standard output (stdout).
+	// The internal implementation uses natefinch/lumberjack v2
```
haircommander, line 254: I don'tt think this makes sense in api documentation

ccojocar (approved):

PR #2907, opened 2025-07-04
Fix the injection of the request UID for JSON Log enricher in case of kubectl node debug

ngopalak-redhat:
#### What type of PR is this?
/kind bug

When a user initiated a node debugging session using `kubectl debug node/<node-name>`, the `requestUID` – essential for tracking the end-user – was not injected into the debug pod's environment by JSON log enricher webhook. This made it difficult to attribute activity in node-debug pods back to the originating user. 

Other cases like `kubectl exec` and `kubectl debug pod` are working, the issue is only with node debugging pod.

To address this, a recent client-side change ([kubernetes/kubernetes#131791](https://github.com/kubernetes/kubernetes/pull/131791)) introduced `app.kubernetes.io/managed-by: "kubectl-debug"` label for these node debugging pods. While this is a step forward, not all Kubernetes API clients (like `oc debug` for OpenShift) utilize this specific label.

This PR introduces a capability: **users can now configure a custom label in the webhook configuration to identify node debugging pods**. This ensures that the requestUID is reliably injected, regardless of the client used to initiate the node debug session. If the custom label is not added it will use  `app.kubernetes.io/managed-by: "kubectl-debug"` .

**Note** that this change has no impact on the exec into a container or pod debug using ephemeral containers.

The execmetadata webhook will now distinguish between different node debug pod types and inject the `SPO_EXEC_REQUEST_UID` environment variable accordingly.

#### Does this PR have test?
Manual tests as kubectl having the node debug pod label is not released.

```release-note
Adds support for kubectl node debugging for JSON Log enricher
```

### DIFF
--- a/installation-usage.md
+++ b/installation-usage.md
@@ -501,7 +501,7 @@ To view the resulting `MutatingWebhookConfiguration`, call:
 $ kubectl get MutatingWebhookConfiguration spo-mutating-webhook-configuration -oyaml
 ```
 
-The Exec Metadata Webhook works in conjunction with the JSON Log Enricher. It's enabled only when JSON Log Enricher is 
+The Exec Metadata and Node Debugging Pod Metadata Webhook works in conjunction with the JSON Log Enricher. It's enabled only when JSON Log Enricher is 
 enabled. For details on its configuration, please refer to the [JSON Log Enricher](#audit-json-log-enricher) section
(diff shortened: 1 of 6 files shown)


ngopalak-redhat:
cc @haircommander @saschagrunert

PR #2909, opened 2025-07-07
#2906: Enricher Filter Options

### LINKED ISSUE #2906: Granular Log Filtering for Enrichers
We need a flexible log filtering mechanism within the Security Profiles Operator's (SPO) `log-enricher` and `json-audit-log-enricher` components. The goal is to significantly reduce the volume of generated logs at these enrichers, preventing unnecessary data transfer and storage costs, and improving the signal-to-noise ratio.

The scope is limited only to these enrichers.

### Problem Statement

The enrichers in SPO generates valuable logs, which are crucial for audit logging and also analyzing syscalls from containers. However, the enricher's log output have these limitations:

1.  **Lack of Granularity:** Existing enrichers provide no built-in capability to filter logs based on common Kubernetes constructs (nodes, pods, containers, namespaces), or other critical attributes like executable names. cmdlines or user activity. The only granularity available is to limit the syscalls or apply seccomp profile to minimal set of pods.
    
2.  **Excessive Log Volume:**
    
    -   Applying a `seccomp` profile cluster-wide or to a large number of nodes/pods, especially with many syscall names in the profiles, results in an overwhelming volume of logs.
        
    -   The recent introduction of the `seccomp` profile flag for privileged containers (https://github.com/cri-o/cri-o/pull/9190) in `crio` increases this, leading to an even greater amount of data from potentially noisy privileged workloads in system namespaces.
        
3.  **Cost and Performance Overhead:** Current approaches require all log data to be transferred from nodes to a central log aggregator (e.g., Splunk) _before_ any filtering can occur. This incurs network bandwidth consumption, storage costs for unneeded data, etc. This high volume of undifferentiated logs makes it difficult to identify the required data.

### Proposal: Policy-Driven Log Filtering

We propose introducing a policy-driven filtering mechanism directly within the `log-enricher` and `json-audit-log-enricher` components. This approach draws inspiration from Kubernetes' own audit logging policy, allowing administrators to define precise rules for which logs are emitted and which are discarded, _before_ they are output to the file/console on the node.

#### Design

A new `filters` field will be added to the SPOD spec, allowing an array of filtering rules. Each rule will specify a `priority`, a `level` (determining the action), and `matchKeys` with optional `matchValues`.

#### Proposed API Structure

The `filters` array would be configured via the `spod` object:
```
kubectl -n security-profiles-operator patch spod spod --type=merge -p '{"spec":{"enableLogEnricher":true,"filters":[]}}'

```
And the `filters` array content would look like this:
```
filters:
  - priority: <integer>      # Order of rule application (lower number = higher priority)
    level: <string>          # "Metadata" (log this line) | "None" (filter this line/do not log)
    matchKeys: []            # Array of keys to match (e.g., ["namespace"])
    matchValues?: []         # Optional: Array of values to match against the matched keys.
                             # If matchValues is empty, presence of key is sufficient.

```

#### Rule Evaluation

1.  Filters will be parsed and loaded when the enrichers start.
    
2.  For each incoming log statement:
    
    -   Rules will be evaluated in ascending order of their `priority` (lower number first).
        
    -   The _first rule_ that a log statement matches will determine its fate :) . Subsequent rules will be ignored for that log line.
        
    -   If a rule matches:
        
        -   If `level: Metadata`, the log line is emitted.
            
        -   If `level: None`, the log line is dropped.
            
    -   If no rule matches, a default behavior (e.g., emit all logs) will be applied.

#### Examples

**1. Filtering JSON Audit Logs for Specific User Activity:**

This example demonstrates logging _only_ audit events associated with a `requestUID`, while filtering everything else. This is helpful for a JSON Audit log enricher to investigate the user activity like exec into a pod or end-user running some script inside a container.

YAML (can be given in JSON also)

```
filters:
  - priority: 100
    level: Metadata
    matchKeys: ["requestUID"]
  - priority: 999 # Catch-all to filter out everything else
    level: None
    matchKeys: ["version"] 
    matchValues:["spo/v1_alpha"] # Matches all logs as its present for all the log lines

```

**2. Filtering Logs for a Specific Kubernetes Namespace:**

This example logs `log-enricher` entries _only_ from the `default` namespace and drops any other log lines.

YAML

```
filters:
  - priority: 101
    level: Metadata
    matchKeys: ["namespace"]
    matchValues: ["default"]
  - priority: 999
    level: None
    matchKeys: ["msg"]
    matchValues: ["audit"]
```

This proposed filtering mechanism will be a significant enhancement to the enrichers in SPO, making it more efficient and user-friendly for managing logs in large-scale Kubernetes environments.

**Note:** I'm interested in implementing this feature myself. My primary use case is filtering logs by `requestUID` to reduce log output when cluster administrator only needs to audit end-user activities like `kubectl exec` into pods.

ngopalak-redhat:
#### What type of PR is this?
/kind feature

This PR introduces Enricher filters. The usage is documented in installation-usage.md file in this PR.

The change adds filters just like any other parameter for the enrichers. The change has no impact on the current behavior of the enrichers. 

Fixes #2906

#### Does this PR have test?
Yes

- I have shuffled the e2e tests so that it covers the use case where the user changes the enrichers.
- For the filter configuration fields in SPOD types (e.g., JsonEnricherFilters), I've used string instead of *string. An empty string ("") provides clearer 'not set' semantics for these fields than a nil pointer.

```release-note
Introduced log filtering capabilities for enrichers, configurable via SPOD, enabling filtering of audit JSON and log-enricher output based on custom rules.
```

### DIFF
--- a/internal/pkg/daemon/enricher/filter.go
+++ b/internal/pkg/daemon/enricher/filter.go
@@ -32,8 +32,7 @@ func GetEnricherFilters(enricherFiltersJsonStr string, logger logr.Logger) ([]ty
 	var enricherFilters []types.EnricherFilterOptions
 	if err := json.Unmarshal([]byte(enricherFiltersJsonStr), &enricherFilters); err != nil {
 		logger.Error(err, "failed to unmarshal enricher filters",
-			"enricherFiltersJsonStr", enricherFiltersJsonStr,
-			"err", err)
+			"enricherFiltersJsonStr", enricherFiltersJsonStr)
 
 		return nil, err
 	}


ngopalak-redhat:
cc: @saschagrunert @haircommander

ngopalak-redhat:
I'm tackling intermittent CI failures caused by `kubectl debug` and `kubectl exec` inconsistencies on the Fedora platform. While our code is good to be reviewed, I'm actively working on a fix to stabilize this platform-specific issues and ensure reliable CI runs.

ngopalak-redhat:
> I'm tackling intermittent CI failures caused by `kubectl debug` and `kubectl exec` inconsistencies on the Fedora platform. While our code is good to be reviewed, I'm actively working on a fix to stabilize this platform-specific issues and ensure reliable CI runs.

I have fixed the issue. The code is good to be reviewed
cc: @saschagrunert @haircommander

cmd/security-profiles-operator/main.go
```diff
@@ -253,6 +254,13 @@ func main() {
 			Name:    "log-enricher",
 			Aliases: []string{"l"},
 			Usage:   "run the audit's log enricher",
+			Flags: []cli.Flag{
+				&cli.StringFlag{
+					Name:  enricherFiltersJsonParam,
+					Value: "",
+					Usage: "Log Enricher filters json.",
```
saschagrunert, line 261:
```suggestion
					Usage: "Log Enricher filters JSON.",
```

cmd/security-profiles-operator/main.go
```diff
@@ -314,6 +322,11 @@ func main() {
 						"The maximum number of days to retain old audit log files based " +
 						"on the timestamp encoded in their filename.",
 				},
+				&cli.StringFlag{
+					Name:  enricherFiltersJsonParam,
+					Value: "",
+					Usage: "JSON Enricher filters json file path.",
```
saschagrunert, line 328:
```suggestion
					Usage: "JSON Enricher filters JSON file path.",
```

internal/pkg/daemon/enricher/filter.go
```diff
@@ -0,0 +1,131 @@
...
+	"strconv"
+	"strings"
+
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/daemon/enricher/types"
+)
+
+func GetEnricherFilters(enricherFiltersJsonStr string) ([]types.EnricherFilterOptions, error) {
+	var enricherFilters []types.EnricherFilterOptions
+	if err := json.Unmarshal([]byte(enricherFiltersJsonStr), &enricherFilters); err != nil {
+		fmt.Printf("failed to unmarshal enricher filters %s: %v\n", enricherFiltersJsonStr, err)
```
saschagrunert, line 33: Can we use a logger here?

api/spod/v1alpha1/spod_types.go
```diff
@@ -365,6 +365,18 @@ type SPODSpec struct {
 	// artifact signature verification.
 	// +optional
 	DisableOCIArtifactSignatureVerification bool `json:"disableOciArtifactSignatureVerification"`
+
+	// LogEnricherFilters if defined, an optional JSON-format filter to determine if log lines should be emitted
+	// for the log-enricher. Defaults to an empty string, meaning no filter is applied and all lines are logged.
+	// +optional
+	// +kubebuilder:default=""
+	LogEnricherFilters string `json:"logEnricherFilters,omitempty"`
```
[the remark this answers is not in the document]
ngopalak-redhat: Initially, I tried defining "spod_types," but this made the solution too complicated. Here's why:

- **Passing Data to DaemonSets Was Complex:** To send JSON data from the controller to each daemon-set, we'd have to modify the configmap. This also meant mounting the configmap into the daemonset pods. A major issue is that manual changes to a configmap don't automatically restart the pods, so the updates wouldn't take effect immediately.

**Filters Are Only for Enrichers**: Only the enrichers use the filters; other parts of SPO don't need them. Therefore, it made more sense to keep the changes localized within the enrichers to avoid unnecessary complexity elsewhere.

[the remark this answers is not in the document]
saschagrunert: Alright this makes sense, thank you for the clarification!

internal/pkg/daemon/enricher/filter.go
ngopalak-redhat: Thanks. Fixed

cmd/security-profiles-operator/main.go
ngopalak-redhat: Fixed

cmd/security-profiles-operator/main.go
ngopalak-redhat: Fixed

internal/pkg/daemon/enricher/filter.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/daemon/enricher/types"
+)
+
+func GetEnricherFilters(enricherFiltersJsonStr string, logger logr.Logger) ([]types.EnricherFilterOptions, error) {
+	var enricherFilters []types.EnricherFilterOptions
+	if err := json.Unmarshal([]byte(enricherFiltersJsonStr), &enricherFilters); err != nil {
+		logger.Error(err, "failed to unmarshal enricher filters",
+			"enricherFiltersJsonStr", enricherFiltersJsonStr,
+			"err", err)
```
saschagrunert, line 36: I don't think we have to add the error twice:
```suggestion
			"enricherFiltersJsonStr", enricherFiltersJsonStr)
```

saschagrunert (commented):
Just a minor nit, otherwise LGTM

ngopalak-redhat:
> Just a minor nit, otherwise LGTM

Fixed and rebased

PR #2947, opened 2025-08-07
Fix Linux release version parsing

mhils:
#### What type of PR is this?

/kind bug

This PR fixes `uname` version parsing for releases with empty build tags, and also stops us from failing hard if we hit a version we cannot parse.

None

#### Does this PR have test?

Yes

```release-note
NONE
```


saschagrunert (approved):
Thanks!

/ok-to-test

PR #3052, opened 2025-11-27
OCPBUGS-62269 - Fix race condition causing missing audit log entries for rapid commands

BhargaviGudi:
#### What type of PR is this?

  /kind bug

  This PR fixes a race condition in the JSON enricher that causes audit log entries to be missing or incomplete when commands are executed in rapid succession (e.g., `touch test1 && touch test2 && touch
  test3`).

  **Problem:**
  When short-lived processes execute quickly, audit events arrive at the JSON enricher before the corresponding BPF events have been processed from the ring buffer. This results in:
  - Empty `cmdLine` fields
  - Missing `requestUID`
  - Null `resource` information (pod/namespace/container)

  **Solution:**
  Implements a retry mechanism in `dispatchSeccompLine()` that attempts to fetch missing information multiple times with progressive delays (0ms, 10ms, 50ms):

  1. **BPF Cache Retry**: Retries cmdLine and requestUID lookup from BPF process cache, giving the BPF ring buffer poller time to process events
  2. **Container Info Retry**: Retries container information lookup if initially null

  Since LogBuckets remain in cache for 60 seconds before being written to the audit log, there's ample time for retries to succeed while BPF events are processed asynchronously.

  **Testing Results:**
  - Before fix: ~33% success rate for rapid consecutive commands
  - After fix: 100% success rate - all commands captured with complete information

  Fixes https://issues.redhat.com/browse/OCPBUGS-62269

  #### Does this PR have test?

  Yes - The fix was validated through:
  - Manual testing with rapid consecutive commands (`oc exec` and `oc rsh`)
  - Existing unit tests continue to pass
  - Build verification completed successfully

  No new automated tests added as this is a timing-dependent race condition that's difficult to reliably reproduce in unit tests.

  - This fix is **scoped exclusively to `jsonenricher.go`** - no modifications to BPF components or ring buffer configuration
  - The retry mechanism only activates when information is missing, avoiding unnecessary delays for normal cases
  - Progressive delays (0ms → 10ms → 50ms) balance responsiveness with reliability
  - Early break on success minimizes latency impact

  ```release-note
  Fixed race condition in JSON enricher causing audit log entries to be missing or incomplete when commands are executed in rapid succession. Audit logs now reliably capture cmdLine, requestUID, and resource
  information for all executed commands, even during bursts of activity.
```

### DIFF
--- a/internal/pkg/daemon/enricher/jsonenricher.go
+++ b/internal/pkg/daemon/enricher/jsonenricher.go
@@ -56,6 +56,9 @@ type JsonEnricher struct {
 	bpfProcessCache  *bpfrecorder.BpfProcessCache
 }
 
+// bpfRetryDelays defines progressive delays for BPF cache retries.
+var bpfRetryDelays = []time.Duration{0, 10 * time.Millisecond, 50 * time.Millisecond}
+
 type JsonEnricherOptions struct {
 	AuditFreq           time.Duration
 	AuditLogPath        string
@@ -204,7 +207,8 @@ func (e *JsonEnricher) Run(ctx context.Context, runErr chan<- error) {
 
 			e.logger.V(config.VerboseLevel).Info("Emit audit log for process",
 				"pid", logItem.Key())
-			e.dispatchSeccompLine(auditLogBucket, nodeName)
+
+			go e.dispatchSeccompLine(ctx, auditLogBucket, nodeName)
 		})
 
 	e.logger.Info(fmt.Sprintf("Setting up caches with expiry of %v", defaultCacheTimeout))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by saschagrunert** [comment]
@BhargaviGudi do you mind a rebase?

File: internal/pkg/daemon/enricher/jsonenricher.go [hunk 1]
```diff
@@ -443,6 +517,13 @@ func (e *JsonEnricher) dispatchSeccompLine(
 		return
 	}
 
+	// Retry BPF cache lookups with progressive delays to allow ring buffer polling to catch up
+	e.retryBpfCmdLine(logBucket)
+	e.retryBpfRequestUID(logBucket)
+
+	// Retry container info lookup if it's still missing
+	e.retryContainerInfo(ctx, logBucket, processID, nodeName)
```
ngopalak-redhat: @BhargaviGudi Thanks for the PR. Can you try to test if `e.retryContainerInfo(ctx, logBucket, processID, nodeName)` alone can fix the issue without the `retryBpfCmdLine` and `retryBpfRequestUID`? I feel missing container info is the cause of the issue.

saschagrunert:
@BhargaviGudi thank you for the PR! Please rebase to fix the CI.

BhargaviGudi:
@ngopalak-redhat Could you please help me review the PR? Thanks

BhargaviGudi:
/retitle OCPBUGS-62269: Fix race condition causing missing audit log entries for rapid commands

BhargaviGudi:
/payload-job periodic-ci-openshift-openshift-tests-private-release-4.22-amd64-nightly-baremetalds-ipi-ovn-lvms-f14-security-profiles

internal/pkg/daemon/enricher/jsonenricher.go
```diff
@@ -411,8 +411,82 @@ func (e *JsonEnricher) fetchProcessInfo(processId int, executable string, uid, g
...
+	if e.bpfProcessCache == nil || logBucket.ProcessInfo.CmdLine != "" || logBucket.ProcessInfo.Pid == 0 {
+		return
+	}
+
+	maxRetries := 3
+	retryDelays := []time.Duration{0, 10 * time.Millisecond, 50 * time.Millisecond}
+
+	for attempt := 0; attempt < maxRetries && logBucket.ProcessInfo.CmdLine == ""; attempt++ {
+		if attempt > 0 {
+			time.Sleep(retryDelays[attempt])
```
saschagrunert: `dispatchSeccompLine` is called from the `OnEviction` callback, and `DeleteExpired()` (line 274) runs every 30s from the `Run()` loop. Evictions are processed synchronously, so sleeping here (up to 60ms per retry function, 120ms total per entry) blocks all subsequent evictions and the main audit line processing loop. For N expired entries, that's N * 120ms of stall.

Consider dispatching into a goroutine so the callback returns immediately:

```go
e.logLinesCache.OnEviction(
    func(ctx context.Context, reason ttlcache.EvictionReason, logItem *ttlcache.Item[int, *types.LogBucket]) {
        auditLogBucket := logItem.Value()
        go e.dispatchSeccompLine(ctx, auditLogBucket, logItem.Key(), nodeName)
    })
```

internal/pkg/daemon/enricher/jsonenricher.go
```diff
@@ -411,8 +411,82 @@ func (e *JsonEnricher) fetchProcessInfo(processId int, executable string, uid, g
...
+		}
+
+		cmdLine, errCmdLine := e.bpfProcessCache.GetCmdLine(logBucket.ProcessInfo.Pid)
+		if errCmdLine == nil {
+			logBucket.ProcessInfo.CmdLine = cmdLine
+			e.logger.V(1).Info("cmdline found in eBPF on retry",
+				"processId", logBucket.ProcessInfo.Pid, "cmdLine", cmdLine, "attempt", attempt+1)
+
+			break
```
saschagrunert: The loop condition already checks `logBucket.ProcessInfo.CmdLine == ""`. After setting `CmdLine` to a non-empty value, the loop terminates on its own. The `break` is redundant.

Same applies to `retryBpfRequestUID` at line 462.

```suggestion
```

internal/pkg/daemon/enricher/jsonenricher.go
```diff
@@ -411,8 +411,82 @@ func (e *JsonEnricher) fetchProcessInfo(processId int, executable string, uid, g
...
+}
+
+// retryBpfRequestUID retries fetching requestUID from BPF environment with progressive delays.
+func (e *JsonEnricher) retryBpfRequestUID(logBucket *types.LogBucket) {
+	if e.bpfProcessCache == nil || logBucket.ProcessInfo.ExecRequestId != nil {
+		return
+	}
+
+	maxRetries := 3
+	retryDelays := []time.Duration{0, 10 * time.Millisecond, 50 * time.Millisecond}
```
saschagrunert: These are identical to lines 421-422. Extract to a package-level variable to avoid the duplication:

```go
var bpfRetryDelays = []time.Duration{0, 10 * time.Millisecond, 50 * time.Millisecond}
```

internal/pkg/daemon/enricher/jsonenricher.go
```diff
@@ -411,8 +411,82 @@ func (e *JsonEnricher) fetchProcessInfo(processId int, executable string, uid, g
...
+				e.logger.V(1).Info("exec request id found in eBPF on retry", "reqId", reqId,
+					"processId", logBucket.ProcessInfo.Pid, "attempt", attempt+1)
+
+				break
+			}
+		}
+	}
+}
+
+// retryContainerInfo retries fetching container info if it's missing.
```
saschagrunert: Unlike the BPF retry functions, this makes a single call with no delay or loop. The name `retryContainerInfo` is misleading. Either rename to something like `refetchContainerInfo`, or add actual retry logic consistent with the other two functions.

internal/pkg/daemon/enricher/jsonenricher.go
```diff
@@ -411,8 +411,82 @@ func (e *JsonEnricher) fetchProcessInfo(processId int, executable string, uid, g
...
+		e.logger.V(1).Info("Container info found on retry",
+			"processID", processID,
+			"pod", logBucket.ContainerInfo.PodName,
+			"namespace", logBucket.ContainerInfo.Namespace)
+	}
+}
+
 func (e *JsonEnricher) dispatchSeccompLine(
-	logBucket *types.LogBucket, nodeName string,
+	ctx context.Context, logBucket *types.LogBucket, processID int, nodeName string,
```
saschagrunert: `processID` is redundant since `logBucket.ProcessInfo.Pid` holds the same value (both sourced from `auditLine.ProcessID`). The BPF retry functions already use `logBucket.ProcessInfo.Pid` directly. Only `retryContainerInfo` uses this parameter, creating an inconsistency.

internal/pkg/daemon/enricher/jsonenricher.go
saschagrunert: The function names already communicate intent. These comments restate what the code does.

```suggestion
	e.retryBpfCmdLine(logBucket)
	e.retryBpfRequestUID(logBucket)
	e.retryContainerInfo(ctx, logBucket, processID, nodeName)
```

saschagrunert (commented):
The diagnosis is correct: short-lived processes with a single audit event miss BPF data because the ring buffer hasn't been polled yet. Retrying at eviction time is a reasonable place since 60 seconds is more than enough for BPF events to arrive. A few concerns and suggestions below.

saschagrunert:
@BhargaviGudi please rebase if this is still an issue.

PR #3110, opened 2026-03-07
feat: add flag to enable anonymous metrics access

jindijamie:
#### What type of PR is this?

/kind feature

Allow metrics access without authentication.
The default behavior is still require authentication.

#### Does this PR have test?

N/A

```release-note
Add --enable-insecure-metrics-access flag for unauthenticated/unencrypted
  metrics access; default secure behavior remains unchanged.
```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -80,22 +80,23 @@ import (
 )
 
 const (
-	spocCmd                      string = "spoc"
-	jsonFlag                     string = "json"
-	nodeStatusControllerFlag     string = "with-nodestatus-controller"
-	spodControllerFlag           string = "with-spod-controller"
-	workloadAnnotatorFlag        string = "with-workload-annotator"
-	recordingMergerFlag          string = "with-recording-merger"
-	recordingFlag                string = "with-recording"
-	seccompFlag                  string = "with-seccomp"
-	selinuxFlag                  string = "with-selinux"
-	apparmorFlag                 string = "with-apparmor"
-	webhookFlag                  string = "webhook"
-	memOptimFlag                 string = "with-mem-optim"
-	defaultWebhookPort           int    = 9443
-	auditLogIntervalSecondsParam string = "audit-log-interval-seconds"
-	auditLogPathParam            string = "audit-log-path"
-	auditLogMaxSizeParam         string = "audit-log-maxsize"
+	spocCmd                          string = "spoc"
+	jsonFlag                         string = "json"
+	nodeStatusControllerFlag         string = "with-nodestatus-controller"
+	spodControllerFlag               string = "with-spod-controller"
+	workloadAnnotatorFlag            string = "with-workload-annotator"
+	recordingMergerFlag              string = "with-recording-merger"
+	recordingFlag                    string = "with-recording"
+	seccompFlag                      string = "with-seccomp"
+	selinuxFlag                      string = "with-selinux"
+	apparmorFlag                     string = "with-apparmor"
+	webhookFlag                      string = "webhook"
+	memOptimFlag                     string = "with-mem-optim"
+	enableAnonymousMetricsAccessFlag string = "enable-anonymous-metrics-access"
+	defaultWebhookPort               int    = 9443
+	auditLogIntervalSecondsParam     string = "audit-log-interval-seconds"
+	auditLogPathParam                string = "audit-log-path"
+	auditLogMaxSizeParam             string = "audit-log-maxsize"
 	// The plural form is not used for audit-log-file-maxbackup to match the k8s api server audit log options.
 	auditLogMaxBackupParam   string = "audit-log-maxbackup"
 	auditLogMaxAgeParam      string = "audit-log-maxage"
@@ -384,6 +385,11 @@ func main() {
 			Value:   config.DefaultProfilingPort,
 			EnvVars: []string{config.ProfilingPortEnvKey},
 		},
+		&cli.BoolFlag{
+			Name:    enableAnonymousMetricsAccessFlag,
+			Usage:   "enable anonymous metrics access",
+			EnvVars: []string{config.EnableAnonymousMetricsAccessEnvKey},
+		},
 	}
 
 	if err := app.Run(os.Args); err != nil {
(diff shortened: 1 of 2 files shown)


ccojocar (approved):

cmd/security-profiles-operator/main.go
```diff
@@ -665,20 +671,29 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
...
 
+	metricsOptions := metricsserver.Options{
+		BindAddress: fmt.Sprintf(":%d", bindata.ContainerPort),
+		ExtraHandlers: map[string]http.Handler{
+			metrics.HandlerPath: met.Handler(),
+		},
+	}
+
+	if ctx.Bool(enableAnonymousMetricsAccessFlag) {
+		setupLog.Info("Anonymous metrics access allowed")
```
saschagrunert: This should be a warning-level log since it is a security-relevant configuration change that weakens the default posture.

cmd/security-profiles-operator/main.go
```diff
@@ -665,20 +671,29 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
...
+
+		metricsOptions.SecureServing = false
```
saschagrunert, line 684: Setting `SecureServing = false` also disables TLS, not just authentication. The flag name suggests only auth is affected, but this silently drops encryption in transit too. Consider keeping TLS enabled even without auth, or at minimum document this behavior clearly.

Also: `disableHTTP2` is only applied in the `else` (authenticated) path. If HTTP/2 was disabled for security reasons (e.g. CVE-2023-44487), that should apply regardless of the auth setting.

internal/pkg/config/config.go
```diff
@@ -94,6 +94,9 @@ const (
 	// EnableRecordingEnvKey is the environment variable key to enabling profile recording.
 	EnableRecordingEnvKey = "ENABLE_RECORDING"
 
+	// EnableAnonymousMetricsAccessEnvKey is the environment variable key for enabling anonymous metrics access.
```
saschagrunert: The release-note block in the PR is empty. This is a user-facing feature and should have a release note.

cmd/security-profiles-operator/main.go
jindijamie: The setupLog is either Info or Error, there is no warning level.

internal/pkg/config/config.go
jindijamie: Updated

cmd/security-profiles-operator/main.go
jindijamie: Update the flag to "enable-insecure-metrics-access" to state the risk. We still want to have the option to disable TLS.

cmd/security-profiles-operator/main.go
```diff
@@ -665,20 +671,29 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
...
+		BindAddress: fmt.Sprintf(":%d", bindata.ContainerPort),
+		ExtraHandlers: map[string]http.Handler{
+			metrics.HandlerPath: met.Handler(),
+		},
+	}
+
+	if ctx.Bool(enableInsecureMetricsAccessFlag) {
+		setupLog.Info("Insecure metrics access allowed (TLS and authentication disabled)")
+
+		metricsOptions.SecureServing = false
```
saschagrunert, line 684: This flag is defined and consumed in `runDaemon`, but there's no mechanism to propagate it to the spod daemonset. The operator manages the daemonset; users can't set env vars on it directly without the operator overwriting them on the next reconciliation.

The established pattern (see `EnableProfiling` in `api/spod/v1alpha1/spod_types.go:289` and `spod_controller.go:773`) is:
1. Add a field to `SPODSpec` (e.g. `EnableInsecureMetricsAccess bool`)
2. Read it in the spod controller and inject `ENABLE_INSECURE_METRICS_ACCESS` as an env var into the daemon container

Without this integration, the flag is unusable in a standard SPO deployment.

cmd/security-profiles-operator/main.go
```diff
@@ -384,6 +385,11 @@ func main() {
 			Value:   config.DefaultProfilingPort,
 			EnvVars: []string{config.ProfilingPortEnvKey},
 		},
+		&cli.BoolFlag{
+			Name:    enableInsecureMetricsAccessFlag,
+			Usage:   "enable insecure metrics access (disables TLS and authentication)",
+			EnvVars: []string{config.EnableInsecureMetricsAccessEnvKey},
+		},
```
saschagrunert, line 392: This flag is only consumed in `runDaemon` but is registered as a global flag. All other daemon-specific feature flags (`seccompFlag`, `selinuxFlag`, `recordingFlag`, `memOptimFlag`) are registered under the `daemon` subcommand's flags (lines 185-212). Move it there for consistency.

cmd/security-profiles-operator/main.go
```diff
@@ -665,20 +671,29 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
...
 
+	metricsOptions := metricsserver.Options{
+		BindAddress: fmt.Sprintf(":%d", bindata.ContainerPort),
+		ExtraHandlers: map[string]http.Handler{
+			metrics.HandlerPath: met.Handler(),
+		},
+	}
+
+	if ctx.Bool(enableInsecureMetricsAccessFlag) {
+		setupLog.Info("Insecure metrics access allowed (TLS and authentication disabled)")
```
saschagrunert, line 682: Follow-up from the previous round: since there's no warning level on `setupLog`, consider making the message stand out by prefixing it:

```suggestion
		setupLog.Info("WARNING: Insecure metrics access enabled, TLS and authentication disabled")
```

saschagrunert (commented):
Follow-up review. The flag rename to `enable-insecure-metrics-access` and the release note address the previous round's concerns, but a few issues remain, one of them significant.

Also: the three commits ("feat: add flag...", "Update the name...", "Fix lint issue") should be squashed into one.

internal/pkg/config/config.go
```diff
@@ -94,6 +94,10 @@ const (
 	// EnableRecordingEnvKey is the environment variable key to enabling profile recording.
 	EnableRecordingEnvKey = "ENABLE_RECORDING"
 
+	// EnableInsecureMetricsAccessEnvKey is the environment variable key for enabling insecure
+	// metrics access(disables TLS and authentication).
```
saschagrunert, line 98: Missing space before parenthesis.

```suggestion
	// EnableInsecureMetricsAccessEnvKey is the environment variable key for enabling insecure
	// metrics access (disables TLS and authentication).
```

ccojocar:
Closing this in favours of #3265.

Issue #3161, opened 2026-05-04
OCI base-profile artifacts at ghcr.io/security-profiles/runc are amd64-only


Ca-moes:
#### What happened:

Setting `baseProfileName: oci://ghcr.io/security-profiles/runc:v1.3.3` (or any other published tag) on a `SeccompProfile` causes spod on a `linux/arm64` host to fail with:

```
ERROR cannot pull base profile  err="read profile: open /tmp/pull-XXXXXXXXX/profile.yaml: no such file or directory"
```

The profile sits in `Pending` indefinitely; no `SecurityProfileNodeStatus` is created.

Inspecting the OCI manifest at any tag in this repository (verified on `v1.1.4` through `v1.4.0`) shows it is a single-arch image manifest (not an OCI image index) with one layer titled `profile-linux-amd64.yaml`:

```bash
TOKEN=$(curl -sSL "https://ghcr.io/token?scope=repository:security-profiles/runc:pull&service=ghcr.io" | jq -r .token)
curl -sSL -H "Authorization: Bearer $TOKEN" \
  -H "Accept: application/vnd.oci.image.index.v1+json,application/vnd.oci.image.manifest.v1+json" \
  "https://ghcr.io/v2/security-profiles/runc/manifests/v1.4.0" \
  | jq '{mediaType, layers: [.layers[].annotations]}'
```

Returns:

```json
{
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "layers": [
    { "org.opencontainers.image.title": "profile-linux-amd64.yaml" }
  ]
}
```

There is no `linux/arm64` (or any other non-amd64) variant published.

#### What you expected to happen:

spod resolves the OCI artifact for the host architecture and merges the runc default profile into the user's `SeccompProfile`. The profile reaches `status.status: Installed`. This is the documented behaviour described in `installation-usage.md` under "Base profile via OCI artifact".

#### How to reproduce it (as minimally and precisely as possible):

1. Bring up a Kubernetes cluster on a `linux/arm64` node (e.g. kind v0.31.0 on an Apple Silicon Mac, or a kind cluster on an arm64 Linux runner).
2. Install cert-manager and SPO from `deploy/operator.yaml`.
3. Apply a `SeccompProfile` that references the OCI base profile:
   ```yaml
   apiVersion: security-profiles-operator.x-k8s.io/v1beta1
   kind: SeccompProfile
   metadata:
     name: test
     namespace: default
   spec:
[...]

PR #3163, opened 2026-05-04
spod: chown emptyDir mounts via fsGroup so cosign TUF cache works

### LINKED ISSUE #3162: spod cosign signature verification fails: cannot write TUF cache to /home/.sigstore
#### What happened:

With the default `disableOciArtifactSignatureVerification: false`, any `SeccompProfile` that references an `oci://` base profile fails to install. spod logs:

```
WARNING: Could not fetch trusted_root.json from the TUF repository.
  chmod /home/.sigstore/root/tuf-repo-cdn.sigstore.dev/tuf_tmpXXXXXXXXX: operation not permitted
ERROR cannot pull base profile oci://ghcr.io/security-profiles/runc:v1.3.3
  err="verify signature: getting Rekor public keys: creating cached local store: operation not permitted"
```

cosign's TUF client wants to persist its bootstrap metadata cache under `$HOME/.sigstore/`, and that write is failing.

cosign successfully fetches and caches its TUF metadata under `$HOME` and proceeds with signature verification. A genuine signature mismatch should produce a verification error, not a filesystem-permission error.

1. Bring up a kind cluster (any host arch).
2. Install cert-manager and SPO from `deploy/operator.yaml`.
3. Leave `disableOciArtifactSignatureVerification` at its default (`false`).
4. Apply a `SeccompProfile` that references any signed OCI base profile:
   ```yaml
   apiVersion: security-profiles-operator.x-k8s.io/v1beta1
   kind: SeccompProfile
   metadata:
     name: test
     namespace: default
   spec:
     baseProfileName: oci://ghcr.io/security-profiles/runc:v1.3.3
     defaultAction: SCMP_ACT_ERRNO
   ```
5. Tail spod logs:
   ```bash
   kubectl -n security-profiles-operator logs ds/spod -c security-profiles-operator
   ```

The errors above appear; the profile never installs.

Setting `disableOciArtifactSignatureVerification: true` on the spod CR makes the profile resolve, at the cost of skipping signature verification entirely.

Looking at `internal/pkg/manager/spod/bindata/spod.go`:

- The spod main container has `HOME=/home` (line 400-402), `RunAsUser: UserRootless` (= 65535), and `ReadOnlyRootFilesystem: true` (line 347, 351-352).
- An `emptyDir` volume named `home-volume` is mounted at `/home` (line 332-333, 874-877), so `/home` itself is *writable* in principle.
- The `PodSecurityContext` (line 156-160) defines only `SeccompProfile` — there is **no `fsGroup`** set.
- `internal/pkg/nonrootenabler/nonrootenabler.go` chowns `/var/lib/security-profiles-operator` to UserRootless but does not chown `/home`.

A PodSpec emptyDir without `fsGroup` is owned by `root:root` with mode `0755`. The container runs as UID 65535 with no supplementary group ownership, so it cannot create `/home/.sigstore/` at all. cosign's TUF client surfaces this as the `chmod ... operation not permitted` line above when its temp-file dance fails.

Suggested fixes (not exclusive):

1. **Set `fsGroup: UserRootless` on the PodSecurityContext.** Smallest change. Kubernetes will recursively chown the `home-volume` emptyDir to `:65535` with group-write, so the cosign TUF cache can be written without changing volume layout.
2. **Mount a dedicated `emptyDir` at `/home/.sigstore` (or wherever cosign caches),** with the right ownership semantics (e.g. via an init container that chowns it to UserRootless, or via `securityContext.fsGroup` on the volume).
3. **Set `XDG_CACHE_HOME` (or whichever env var cosign / sigstore-go honours) to a writable path under a different mount,** so the cache lives somewhere with explicit permissions rather than under `/home`.

Today's effective workaround for users — `disableOciArtifactSignatureVerification: true` — turns off the supply-chain protection that motivates using a signed OCI artifact in the first place; making verification work in the default configuration would be a meaningful security improvement.

#### Environment:

- Cloud provider or hardware configuration: kind v0.31.0 (any host arch — this is a permissions issue, not arch-specific).
- OS: Ubuntu 24.04 / kind node images.
- Kernel: 6.x range.
- Others: SPO `v0.10.0`; reproduced against `main` HEAD `dbc1b7a5c`.

Ca-moes:
#### What type of PR is this?

/kind bug

Sets `FSGroup: UserRootless` on the spod `PodSecurityContext`. The spod main container runs as UID 65535 (`UserRootless`) with `ReadOnlyRootFilesystem: true`, and mounts an `emptyDir` named `home-volume` at `HomeDirectory` (`/home`). With no `fsGroup` on the pod, kubelet leaves emptyDir volumes owned by `root:root` mode `0755`, so the non-root container cannot create `/home/.sigstore/`. cosign's TUF client surfaces this as:

```
WARNING: Could not fetch trusted_root.json from the TUF repository.
  chmod /home/.sigstore/root/tuf-repo-cdn.sigstore.dev/tuf_tmpXXX: operation not permitted
ERROR cannot pull base profile oci://ghcr.io/security-profiles/runc:vX.Y.Z
  err="verify signature: getting Rekor public keys: creating cached local store: operation not permitted"
```

So with the default `disableOciArtifactSignatureVerification: false`, any `SeccompProfile` referencing an `oci://` base profile is unusable today. Users have had to set `disableOciArtifactSignatureVerification: true` on the spod CR, defeating the supply-chain protection that motivated using a signed OCI artifact in the first place.

`fsGroup: UserRootless` makes kubelet recursively chown the emptyDirs to `:65535` with group-write. The other emptyDir mounts on the spod main container (`tmp-volume`, `grpc-server-volume`, `selinuxd-private-volume`, `selinux-drop-dir`) are unaffected (they were already writable by the container UID through different means). HostPath mounts are not chowned by `fsGroup` and remain unchanged. No other containers in the pod (init / non-root-enabler / metrics / log-enricher / bpf-recorder / selinuxd) have a behaviour-altering interaction with `fsGroup`.

The hardcoded TUF cache path comes from the vendored sigstore client (`vendor/github.com/sigstore/sigstore/pkg/tuf/client.go` ≈ line 561), so an env var override (`TUF_ROOT`) was considered as an alternative. Fixing ownership at the volume layer is the smaller and more direct change, and it doesn't introduce a new env var to document.

Fixes #3162

#### Does this PR have test?

No new automated tests. Manual verification: install SPO from this branch into a kind cluster, leave `disableOciArtifactSignatureVerification` at its default `false`, apply a `SeccompProfile` referencing `oci://ghcr.io/security-profiles/runc:v1.3.3`, observe spod logs — cosign successfully fetches and caches TUF metadata at `/home/.sigstore/`, and the profile reaches `status.status: Installed` (still amd64-only on the published artifact side; that's the orthogonal #3161).

`test/tc_base_profiles_oci_test.go` is currently `nolint:unused` ("test is flaky and therefore got deactivated") and patches spod with `"disableOciArtifactSignatureVerification": true` to bypass the very issue this PR fixes. With this change, that patch can be dropped and the test could be re-enabled — but I haven't bundled that here because the original flakiness reason isn't documented in the surrounding history, and re-enabling without understanding the prior flakiness feels out of scope for a fix PR. Happy to do it as a follow-up if you'd prefer.

Related: #3161 (runc OCI base-profile artifacts published as amd64-only) is the other half of the OCI base-profile path. That one is on the publishing side and is independent of this fix.

```release-note
spod: cosign signature verification on OCI base profiles now works in the default configuration. Previously the cosign TUF metadata cache write to `/home/.sigstore/` failed because the `home-volume` emptyDir was owned by `root:root` while the spod main container runs as a non-root UID. Setting `FSGroup: UserRootless` on the PodSecurityContext makes the emptyDir writable by the container user, so users no longer need `disableOciArtifactSignatureVerification: true` as a workaround.
```

### DIFF
--- a/internal/pkg/manager/spod/bindata/spod.go
+++ b/internal/pkg/manager/spod/bindata/spod.go
@@ -157,6 +157,7 @@ var Manifest = &appsv1.DaemonSet{
 					SeccompProfile: &corev1.SeccompProfile{
 						Type: corev1.SeccompProfileTypeRuntimeDefault,
 					},
+					FSGroup: &userRootless,
 				},
 				InitContainers: []corev1.Container{
 					{


saschagrunert (approved):
Thanks!

saschagrunert (commented):
Correct fix using the standard Kubernetes primitive for this exact problem. `FSGroup: &userRootless` is consistent with the main container's existing `RunAsUser`/`RunAsGroup` (lines 351-352). Other containers run as root, so the added group-write for GID 65535 on their emptyDir mounts has no practical effect. HostPath mounts are unaffected.

The deactivated test in `test/tc_base_profiles_oci_test.go` that patches `disableOciArtifactSignatureVerification: true` could drop that patch as a follow-up once this lands.

saschagrunert:
Thanks!

PR #3190, opened 2026-05-14
CMP-4310: added dynamic TLS settings

vickeybrown:
#### What type of PR is this?

/kind feature

This PR adds support for dynamic TLS configuration from the OpenShift cluster 
  API server, making the security-profiles-operator respect cluster-wide TLS
  policies instead of using hardcoded settings.
                  
  Key changes:                                                                  
  - Fetches TLS configuration (cipher suites, minimum TLS version) from the
  OpenShift APIServer resource at startup                                       
  - Respects the cluster's TLS adherence policy to determine whether to honor
  cluster TLS settings                                                          
  - Implements a TLS profile watcher that triggers graceful operator shutdown
  when TLS configuration changes, allowing Kubernetes to restart the pod with   
  updated settings 
                                                                                
  This aligns the security-profiles-operator with OpenShift's cluster-wide      
  security policies and enables centralized TLS management. 

Fixes [CMP-4310](https://redhat.atlassian.net/browse/CMP-4310)

#### Does this PR have test?

N/A

- The operator now requires `get`, `list`, and `watch` permissions on         
  `config.openshift.io/apiservers` resources (RBAC changes included in
  `deploy/base/role.yaml`)                                                      
  - When TLS configuration changes are detected, the operator performs a
  graceful shutdown expecting Kubernetes to restart the pod with the new        
  configuration
  - HTTP/2 is disabled by default for security reasons                          
  - Falls back to safe defaults if the APIServer TLS profile cannot be fetched

```release-note
The `--tls-min-version` flag has been removed from the
  security-profiles-operator. TLS configuration (minimum version and cipher     
  suites) is now dynamically fetched from the OpenShift cluster's APIServer
  resource and automatically updated when the cluster TLS policy changes. 
```


saschagrunert:
@vickeybrown thank you for the PR, do you mind fixing the CI?

vickeybrown:
CI is fixed, I think this is in a good spot for review now

saschagrunert (commented):
Did a first review, see inline comments for details.

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +624,182 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+	}
+
+	setupLog.Info("ending " + componentName)
+
+	// If TLS config changed, exit with a specific code so the pod restarts
+	// This ensures a clean restart with new TLS settings
+	if tlsConfigChanged {
+		setupLog.Info("exiting due to TLS configuration change - pod will restart with new settings")
+		cancelManager()
+		os.Exit(0)
```
saschagrunert: `os.Exit(0)` inside a function that returns `error` bypasses all deferred cleanup in callers and makes the function untestable. Return a sentinel error instead and let `main()` decide whether to exit:

```go
var ErrTLSConfigChanged = errors.New("TLS configuration changed, restart required")
```

Then in the caller:
```go
if errors.Is(err, ErrTLSConfigChanged) {
    os.Exit(0)
}
```

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +624,182 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+) error {
+	// Set up signal handler first
+	sigHandler := ctrl.SetupSignalHandler()
+
+	// Create a cancellable context for graceful shutdown on TLS config changes
+	managerCtx, cancelManager := context.WithCancel(context.Background())
+
+	// Set up TLS profile watcher to trigger graceful shutdown on changes
+	// This is only available in OpenShift environments
+	tlsConfigChanged := false
```
saschagrunert: Data race: this `bool` is written from TLS watcher callbacks (running in controller-manager goroutines) and read on the main goroutine after `mgr.Start()` returns. Use `atomic.Bool` to avoid the race.

```suggestion
	var tlsConfigChanged atomic.Bool
```

Then set with `tlsConfigChanged.Store(true)` and read with `tlsConfigChanged.Load()`.

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +624,182 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+		&configv1.ClusterOperator{},
+	)
+	if err == nil {
+		setupLog.Info("OpenShift detected, fetching cluster TLS configuration")
+
+		isOpenShift = true
+	} else {
+		setupLog.Info("OpenShift not detected, using default TLS configuration", "reason", err)
+
+		isOpenShift = false
```
saschagrunert, line 694: Any error is treated as "not OpenShift", which masks RBAC denials, network timeouts, or transient API server failures. On an actual OpenShift cluster with a transient issue, the operator would silently skip TLS enforcement.

The existing pattern in `internal/pkg/manager/spod/bindata/ca.go` is more careful: it distinguishes "not found / not registered" errors from unexpected ones using the `IsNotFound()` helper. Reuse that pattern here:

```go
if err == nil {
    isOpenShift = true
} else if bindata.IsNotFound(err) {
    isOpenShift = false
} else {
    return nil, configv1.TLSProfileSpec{}, configv1.TLSAdherencePolicyNoOpinion, false,
        fmt.Errorf("detect OpenShift platform: %w", err)
}
```

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +624,182 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+	// Fetch TLS profile and adherence policy if on OpenShift
+	var initialTLSProfile configv1.TLSProfileSpec
+
+	var initialTLSAdherencePolicy configv1.TLSAdherencePolicy
+
+	if isOpenShift {
+		initialTLSProfile, err = tlspkg.FetchAPIServerTLSProfile(ctx, preStartClient)
+		if err != nil {
+			return nil, configv1.TLSProfileSpec{}, configv1.TLSAdherencePolicyNoOpinion, false,
+				fmt.Errorf("fetch OpenShift TLS profile: %w", err)
```
saschagrunert: If the APIServer resource is temporarily unavailable (during upgrades, API server restarts), the operator refuses to start entirely with no retry. Consider falling back to the default TLS profile with a warning instead of blocking startup. Otherwise a transient API server issue turns into a CrashLoopBackOff.

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +624,182 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+// then starts it. Returns true if the manager stopped due to TLS configuration change.
+func setupManagerWithTLSWatcher(
+	mgr ctrl.Manager,
+	initialTLSProfile configv1.TLSProfileSpec,
+	initialTLSAdherencePolicy configv1.TLSAdherencePolicy,
+	isOpenShift bool,
+	componentName string,
+) error {
+	// Set up signal handler first
+	sigHandler := ctrl.SetupSignalHandler()
```
saschagrunert: `ctrl.SetupSignalHandler()` panics if called more than once per process. Moving it from the clearly single-entry-point `run*` functions into a shared helper makes that easy to trigger accidentally. Consider passing the signal context as a parameter.

cmd/security-profiles-operator/main.go
```diff
@@ -101,17 +105,11 @@ const (
 	auditLogMaxAgeParam      string = "audit-log-maxage"
 	enricherFiltersJsonParam string = "enricher-filters-json"
 	enricherLogSourceParam   string = "enricher-log-source"
-	tlsMinVersionParam       string = "tls-min-version"
-	defaultTlsMinVersion     string = "1.2"
 )
```
saschagrunert, line 108: Removing a user-facing flag is a breaking change. Existing deployments or automation passing `--tls-min-version` will fail on upgrade with an unknown flag error. Consider keeping the flag as deprecated-and-ignored for one release (log a warning if set), then removing it in the next release.

internal/pkg/util/tlsobserver.go
```diff
@@ -0,0 +1,72 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
saschagrunert:
```suggestion
Copyright 2026 The Kubernetes Authors.
```

cmd/security-profiles-operator/main.go
```diff
@@ -782,7 +943,7 @@ func runNonRootEnabler(ctx *cli.Context, info *version.Info) error {
 
 	printInfo(component, info)
 
-	runtime := ctx.String("runtime")
+	runtimeType := ctx.String("runtime")
```
saschagrunert: Unrelated rename. This and the other unrelated changes (Flatcar CI tmpfs/cache workarounds, log enricher `--since-time` filtering, error message fix) should be in separate commits to keep the diff reviewable.

hack/ci/e2e-flatcar-dev-container.sh
```diff
@@ -15,7 +15,9 @@
 set -eu
 
 dev() {
-  systemd-nspawn -q -i /opt/bin/flatcar_developer_container.bin --bind=/:/hostfs ${@}
+  # Mount /tmp with 2GB size to accommodate large Go compilations with extensive vendor dependencies
+  # Default tmpfs is only 796MB which is insufficient for building with OpenShift libraries
+  systemd-nspawn -q -i /opt/bin/flatcar_developer_container.bin --bind=/:/hostfs --tmpfs=/tmp:size=2G ${@}
```
saschagrunert, line 19: This tmpfs workaround and the cache cleanup changes in `e2e-flatcar.sh` appear to compensate for the increased binary size from vendoring `openshift/library-go`. They should be in a separate commit with their own rationale, not bundled with the TLS feature.

saschagrunert:
Please rebase and apply the changes to the v1 API and not the old one.

cmd/security-profiles-operator/main.go
vickeybrown: I remember now why I renamed this, the linter doesn't like it - `cmd/security-profiles-operator/main.go:1050:2: importShadow: shadow of imported from 'k8s.io/apimachinery/pkg/runtime' package 'runtime' (gocritic)
	runtime := ctx.String("runtime")`

cmd/security-profiles-operator/main.go
```diff
@@ -409,12 +413,25 @@ func main() {
...
+		}
+
 		setupLog.Error(err, "running security-profiles-operator")
-		//nolint:gocritic // this is intentional to return to correct exit code
+
+		//nolint:gocritic,exitAfterDefer // intentional exit with specific code
 		os.Exit(1)
 	}
 
 	if err := app.RunContext(mainctx, os.Args); err != nil {
```
rhmdnd: This is a pre-existing pattern, so fixing it in your PR probably isn't necessary. But I wonder why we call run twice?

https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/cmd/security-profiles-operator/main.go#L411-L419

cmd/security-profiles-operator/main.go
```diff
@@ -675,6 +692,199 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+func setupManagerWithTLSWatcher(
+	sigHandler context.Context,
+	mgr ctrl.Manager,
+	initialTLSProfile configv1.TLSProfileSpec,
+	initialTLSAdherencePolicy configv1.TLSAdherencePolicy,
+	isOpenShift bool,
+	componentName string,
+) error {
+	// Create a cancellable context derived from sigHandler for graceful shutdown on TLS config changes
+	managerCtx, cancelManager := context.WithCancel(sigHandler)
```
rhmdnd, line 824: What if we `defer cancelManager()` here instead of calling it in various conditional branches below?

bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
```diff
@@ -585,6 +586,14 @@ spec:
           - get
           - patch
           - update
+        - apiGroups:
+            - config.openshift.io
```
rhmdnd: Minor indentation inconsistencies here.

cmd/security-profiles-operator/main.go
```diff
@@ -675,6 +692,199 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+
+	// Set up TLS profile watcher to trigger graceful shutdown on changes
+	// This is only available in OpenShift environments
+	var tlsConfigChanged atomic.Bool
+
+	if isOpenShift {
```
rhmdnd, line 831: I wonder if we can add a dedicated test for this in e2e somewhere?

https://github.com/kubernetes-sigs/security-profiles-operator/tree/main/test

Unless @saschagrunert has a better recommendation of where this testing should live.

An e2e test could have a conditional at the top of the test to short-circuit if SPO isn't running on an OpenShift cluster (e.g., `clusterType == clusterTypeOpenShift`).

Some other tests are doing this, too.

rhmdnd (commented):
The overall approach for following the TLS adherence guidance makes sense with what I've seen in other OpenShift operators. My comments are mostly regarding logic questions on the periphery of the feature, and not the core feature itself, along with some questions about testing.

cmd/security-profiles-operator/main.go
saschagrunert: An e2e smoke test that verifies the operator starts with the correct TLS profile on OpenShift (gated by `clusterType == clusterTypeOpenShift`) sounds reasonable and low-risk.

Testing the dynamic restart-on-TLS-change behavior is harder in e2e though, since it requires mutating the cluster-wide `APIServer` config and waiting for the operator to cycle. That might be better suited for a dedicated integration test environment. Unit tests with mocked clients would give good coverage for that path without the blast radius.

cmd/security-profiles-operator/main.go
vickeybrown: Maybe a bug? Want me to open a new ticket to clean it up?

cmd/security-profiles-operator/main.go
vickeybrown: I updated it so a couple of the conditional branches will use the defer method, but some of them still need to have cancelManager() in them so they can cancel managerCtx

cmd/security-profiles-operator/main.go
vickeybrown: I created https://github.com/kubernetes-sigs/security-profiles-operator/pull/3332 to rename the variable, please take a look when ya get a chance!

cmd/security-profiles-operator/main.go
vickeybrown: https://github.com/kubernetes-sigs/security-profiles-operator/pull/3333

vickeybrown:
This should be ready for another round of review! Build/macos and pull-security-profiles-operator-verify will pass once https://github.com/kubernetes-sigs/security-profiles-operator/pull/3332 is merged,  and I think pull-security-profiles-operator-test-e2e is a flaky test

cmd/security-profiles-operator/main.go
```diff
@@ -671,6 +680,201 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+			setupLog.Info("TLS profile watcher setup failed", "error", err)
+		} else {
+			setupLog.Info("TLS profile watcher enabled - will trigger graceful shutdown on configuration changes")
+		}
+	} else {
+		setupLog.Info("TLS profile watcher not enabled (OpenShift-only feature)")
+	}
+
+	// Merge signal handler with TLS config change context
+	go func() {
```
saschagrunert: `managerCtx` is already a child of `sigHandler` via `context.WithCancel(sigHandler)` at line 821, so it is automatically canceled when `sigHandler` is done. This goroutine waits on `<-sigHandler.Done()` then calls `cancelManager()` on an already-canceled context, so it has no effect.

It also leaks: if `cancelManager()` fires first (TLS config change), this goroutine blocks on `<-sigHandler.Done()` until process exit.

So I guess we can remove the goroutine.

cmd/security-profiles-operator/main.go
```diff
@@ -671,6 +680,201 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+		types.NamespacedName{Name: "openshift-apiserver"},
+		&configv1.ClusterOperator{},
+	)
+
+	switch {
+	case err == nil:
+		setupLog.Info("OpenShift detected, fetching cluster TLS configuration")
+
+		isOpenShift = true
+	case bindata.IsNotFound(err) || errors.Is(err, context.DeadlineExceeded):
```
saschagrunert: Treating `context.DeadlineExceeded` the same as `IsNotFound` reintroduces the problem from the first review round: on a real OpenShift cluster with a slow or loaded API server, a timeout is silently treated as "not OpenShift", and TLS enforcement is skipped without any indication.

The 1-second timeout at line 711 is tight for a real API call during rolling restarts or high load. Consider either raising it (5-10s), or keeping the timeout but returning an error on `DeadlineExceeded` instead of treating it as non-OpenShift:

```go
case bindata.IsNotFound(err):
	isOpenShift = false
case errors.Is(err, context.DeadlineExceeded):
	return nil, configv1.TLSProfileSpec{}, configv1.TLSAdherencePolicyNoOpinion, false,
		fmt.Errorf("timed out detecting OpenShift platform: %w", err)
```

cmd/security-profiles-operator/main.go
```diff
@@ -671,6 +680,201 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+		tlsConfigFunc, unsupportedCiphers := tlspkg.NewTLSConfigFromProfile(initialTLSProfile)
+		if len(unsupportedCiphers) > 0 {
+			setupLog.Info("TLS profile contains unsupported ciphers", "ciphers", unsupportedCiphers)
+		}
+
+		tlsOptions = append(tlsOptions, tlsConfigFunc)
+	} else {
+		setupLog.Info("using default TLS configuration", "adherence-policy", initialTLSAdherencePolicy)
+		// Explicitly apply the default TLS profile instead of relying on Go's built-in defaults
+		defaultProfile := configv1.TLSProfiles[libgocrypto.DefaultTLSProfileType]
```
saschagrunert: On non-OpenShift clusters, `initialTLSAdherencePolicy` is `NoOpinion`, so `ShouldHonorClusterTLSProfile` returns `false` and this `else` branch runs. It applies OpenShift's `IntermediateType` cipher suite list via `NewTLSConfigFromProfile`, replacing Go's built-in cipher selection.

Previously, non-OpenShift webhooks only set `MinVersion = TLS 1.2` and disabled HTTP/2. Now they also get an explicit (and more restrictive) cipher suite list from OpenShift's intermediate profile. This could break vanilla Kubernetes users whose clients rely on ciphers that Go allows but OpenShift's intermediate profile does not.

Consider keeping Go's defaults for the non-OpenShift path (just disable HTTP/2 and set min version) and only applying OpenShift profiles when `isOpenShift` is true.

cmd/security-profiles-operator/main.go
```diff
@@ -895,7 +1098,27 @@ func runWebhook(ctx *cli.Context, info *version.Info) error {
...
 	if err != nil {
-		return fmt.Errorf("create cluster webhook: %w", err)
+		return fmt.Errorf("create cluster manager: %w", err)
+	}
+
+	if err := configv1.AddToScheme(mgr.GetScheme()); err != nil {
+		return fmt.Errorf("add OpenShift config API to scheme: %w", err)
+	}
+
+	if err := spodv1alpha1.AddToScheme(mgr.GetScheme()); err != nil {
```
saschagrunert: These four scheme registrations (`spodv1alpha1`, `spodv1`, `secprofnodestatusv1alpha1`, `secprofnodestatusv1`) are new to `runWebhook`. The current main branch does not have them here. Are they required for the TLS watcher or the `SecurityProfileWatcher` controller? If so, a comment explaining the dependency would help. If not, they should be in a separate change.

cmd/security-profiles-operator/main.go
vickeybrown: Makes sense, I added the 1 second timeout to try to get the e2e-apparmor-profile test working - but it didn't help so I'll probably put it back how it was, with the longer timeout

saschagrunert:
@vickeybrown do you mind a rebase?

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +646,212 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+
+	// Fetch TLS profile and adherence policy if on OpenShift
+	var initialTLSProfile configv1.TLSProfileSpec
+
+	var initialTLSAdherencePolicy configv1.TLSAdherencePolicy
+
+	if isOpenShift {
+		// Create a timeout context for fetching TLS configuration from OpenShift
+		// Use a longer timeout since we're actually reading configuration data
+		fetchCtx, fetchCancel := context.WithTimeout(ctx, 10*time.Second)
```
saschagrunert: This 10-second context is shared between `FetchAPIServerTLSProfile` (line 711) and `FetchAPIServerTLSAdherencePolicy` (line 727) sequentially. If the first call is slow (e.g. 9 seconds during an API server rolling restart), the second gets only 1 second. If the first times out entirely, the context is already expired and the second call fails immediately too.

Give each call its own timeout:

```go
profileCtx, profileCancel := context.WithTimeout(ctx, 10*time.Second)
defer profileCancel()
initialTLSProfile, err = tlspkg.FetchAPIServerTLSProfile(profileCtx, preStartClient)
// ...
adherenceCtx, adherenceCancel := context.WithTimeout(ctx, 10*time.Second)
defer adherenceCancel()
initialTLSAdherencePolicy, err = tlspkg.FetchAPIServerTLSAdherencePolicy(adherenceCtx, preStartClient)
```

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +646,212 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+				cancelManager()
+			},
+			func(ctx context.Context, oldPolicy, newPolicy configv1.TLSAdherencePolicy) {
+				setupLog.Info("TLS adherence policy changed - triggering graceful shutdown to apply new configuration")
+
+				tlsConfigChanged.Store(true)
+
+				cancelManager()
+			}); err != nil {
+			setupLog.Info("TLS profile watcher setup failed", "error", err)
```
saschagrunert: On OpenShift, if this fails the operator runs its entire lifetime without detecting TLS config changes. Logging at Info makes this easy to miss in monitoring setups that filter on Warning/Error.

```suggestion
			setupLog.Error(err, "TLS profile watcher setup failed, TLS configuration changes will not be detected")
```

cmd/security-profiles-operator/main.go
```diff
@@ -635,6 +646,212 @@ func newMemoryOptimizedCache(ctx *cli.Context) cache.NewCacheFunc {
...
+	defer cancelManager()
+
+	// Set up TLS profile watcher to trigger graceful shutdown on changes
+	// This is only available in OpenShift environments
+	var tlsConfigChanged atomic.Bool
+
+	if isOpenShift {
+		if err := util.SetupTLSWatcher(mgr, initialTLSProfile, initialTLSAdherencePolicy,
+			func(ctx context.Context, oldProfile, newProfile configv1.TLSProfileSpec) {
+				setupLog.Info("TLS profile changed - triggering graceful shutdown to apply new configuration")
```
saschagrunert: `SetupTLSWatcher` in `tlsobserver.go:52` already logs `"TLS profile changed - triggering graceful shutdown for restart"` with old/new min-version details before calling this callback. This produces two near-identical log lines per TLS change event. Same for the adherence policy callback at line 822.

Remove the logs from these callbacks since `tlsobserver.go` already logs with the more useful details.

internal/pkg/daemon/apparmorprofile/apparmorprofile.go
```diff
@@ -97,6 +97,10 @@ func (r *Reconciler) checkAppArmor() error {
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=apparmorprofiles,verbs=get;list;watch;create;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=apparmorprofiles/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=apparmorprofiles/finalizers,verbs=delete;get;update;patch
+//
+// OpenShift (This is ignored in other distros):
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators,verbs=get;list;watch
```
saschagrunert: These `config.openshift.io` permissions are needed by `fetchTLSOptions` in `main.go`, not by the AppArmor reconciler. `deploy/base/role.yaml` already has these entries (added in this PR), so the kubebuilder markers here are redundant. They also create a maintenance risk: if the AppArmor controller is ever removed, the generated ClusterRole silently loses these entries even though the daemon startup code still needs them.

Same applies to `internal/pkg/webhooks/recording/recording.go:83`.

Consider removing the markers from both files and relying on the entries already in `deploy/base/role.yaml`.

internal/pkg/webhooks/recording/recording.go
```diff
@@ -78,6 +78,10 @@ func RegisterWebhook(
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=profilerecordings/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=profilerecordings/finalizers,verbs=delete;get;update;patch
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+//
+// OpenShift (This is ignored in other distros):
+// +kubebuilder:rbac:groups=config.openshift.io,resources=clusteroperators,verbs=get;list;watch
```
saschagrunert: Same as the comment on `apparmorprofile.go:102`. The recording webhook doesn't use `config.openshift.io` resources; `deploy/base/role.yaml` already covers this.

saschagrunert (approved):

saschagrunert:
Thank you!

PR #3212, opened 2026-06-01
Make sure that the apparmor profile name is prefixed with _spo

ccojocar:
Make sure that the apparmor profile installed into node is always prefixed with _spo

#### What type of PR is this?

/kind bug

The apparmor profile installed into the kernel and the associated file name
should be prefixed with `spo_` to avoid overwriting any existing
profiles already loaded into the kerenel.

This can allow an attacker which control the CRD name of the
ApparmorProfile to overwrite an existing profile which is already loaded
into the kernel.

```release-note
The apparmor profile installed by the spo into the nodes are now prefixed with `spo_` to avoid 
any colisions with existing profiles already loaded by the kernel.
```


saschagrunert (approved):

ccojocar:
The fix is a bit more involved. This breaks the tests. I'm working a proper fix.

saschagrunert (approved):

saschagrunert (approved):

ccojocar:
Closing this for now because is breaking the `securityContext` where the profile name is referenced by name, and this means everyone will need to prefix everything with `spo_`.  

We need to find a different solution for preventing overwriting existing profiles which are not owned by spo.

PR #3235, opened 2026-06-09
test adjustments

vickeybrown:
#### What type of PR is this?

/kind failing-test

In https://github.com/kubernetes-sigs/security-profiles-operator/pull/3190 we pull in a large vendor package that causes e2e test failures. This PR adjusts the test to allow them to properly handle testing the bigger vendor files.

#### Which issue(s) this PR fixes

None

#### Does this PR have test?

N/A

```release-note
NONE
```

### DIFF
--- a/hack/ci/e2e-flatcar-dev-container.sh
+++ b/hack/ci/e2e-flatcar-dev-container.sh
@@ -15,7 +15,8 @@
 set -eu
 
 dev() {
-  systemd-nspawn -q -i /opt/bin/flatcar_developer_container.bin --bind=/:/hostfs ${@}
+  # --tmpfs=/tmp:size=2G: Allocate 2GB tmpfs to accommodate large vendor dependencies during test runs
+  systemd-nspawn -q -i /opt/bin/flatcar_developer_container.bin --bind=/:/hostfs --tmpfs=/tmp:size=2G ${@}
 }
 
 dev --chdir=/hostfs/vagrant -- /hostfs/vagrant/hack/ci/e2e-flatcar.sh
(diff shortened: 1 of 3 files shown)


hack/ci/e2e-flatcar.sh
```diff
@@ -41,6 +41,26 @@ alias k=kubectl
 # Configure git to consider the mounted host filesystem as safe
 git config --global --add safe.directory "/hostfs/vagrant"
 
+# Clean up Go build cache to free up disk space before running tests
+go clean -cache -testcache -modcache || true
```
saschagrunert: `-modcache` will wipe all downloaded modules right before `make test-e2e` tries to compile, which will force a full re-download or fail if the build is offline. This should be just:
```suggestion
go clean -cache -testcache || true
```

hack/ci/e2e-flatcar-dev-container.sh
```diff
@@ -15,7 +15,7 @@
 set -eu
 
 dev() {
-  systemd-nspawn -q -i /opt/bin/flatcar_developer_container.bin --bind=/:/hostfs ${@}
+  systemd-nspawn -q -i /opt/bin/flatcar_developer_container.bin --bind=/:/hostfs --tmpfs=/tmp:size=2G ${@}
```
saschagrunert, line 19: Nit: a short comment explaining why 2G was chosen would help future maintainers decide whether it is safe to change this value.

saschagrunert (commented):
Thanks for fixing the log enricher log scoping, that aligns with how `waitForEnricherLogs` already works. A couple of suggestions below.

saschagrunert (approved):

PR #3241, opened 2026-06-10
fix: partial profiles never merged due to namespace filter

mirza-src:
#### What type of PR is this?

/kind bug

`SeccompProfile`, `SelinuxProfile`, and `AppArmorProfile` are all cluster-scoped resources. The `listPartialProfiles` and `deletePartialProfiles` functions were passing `client.InNamespace()` when listing/deleting them, which silently prevented profile merging from ever running when a `ProfileRecording` with `mergeStrategy: containers` was deleted.

Due to a known controller-runtime inconsistency (kubernetes-sigs/controller-runtime#988), `List()` and `DeleteAllOf()` behave differently when `InNamespace` is used with cluster-scoped resources. `List()` returns empty results with no error, while `DeleteAllOf()` ignores the namespace and deletes cluster-wide.

Additionally, the `spo.x-k8s.io/recording-id` label only stored the recording name, not its namespace, meaning recordings with the same name in different namespaces would share and interfere with each other's partial profiles. A new label `spo.x-k8s.io/recording-namespace` is introduced and added to profiles at creation time, and used as an additional filter during merge and delete.
None

#### Does this PR have test?

N/A

The `client.InNamespace()` removal from `deletePartialProfiles` is also necessary: without the namespace label filter, `DeleteAllOf()` would have deleted partial profiles belonging to same-named recordings in other namespaces.

```release-note
Fixed issue with profile merging for recordings caused by namespace filtering of cluster-scoped profiles.
action required: Partial profiles created prior to this release are completely orphaned. They will not be merged or deleted even after upgrading because they lack the `spo.x-k8s.io/recording-namespace` label. If you have partial profiles created prior to this release, delete them manually: `kubectl delete selinuxprofiles,seccompprofiles,apparmorprofiles -l 'spo.x-k8s.io/partial=true,!spo.x-k8s.io/recording-namespace'`
```

### DIFF
--- a/api/profilebase/v1/profilebase.go
+++ b/api/profilebase/v1/profilebase.go
@@ -35,7 +35,7 @@ const ProfilePartialLabel = "spo.x-k8s.io/partial"
 type SecurityProfileBase interface {
 	client.Object
 
-	ListProfilesByRecording(ctx context.Context, cli client.Client, recording string) ([]metav1.Object, error)
+	ListProfilesByRecording(ctx context.Context, cli client.Client, recording string, recordingNamespace string) ([]metav1.Object, error)
 	IsPartial() bool
 	IsDisabled() bool
 	IsReconcilable() bool
@@ -65,8 +65,10 @@ func ListProfilesByRecording(
 	if err := cli.List(
 		ctx,
 		list,
-		client.InNamespace(recordingNamespace),
-		client.MatchingLabels{profilerecordingv1.ProfileToRecordingLabel: recordingName}); err != nil {
+		client.MatchingLabels{
+			profilerecordingv1.ProfileToRecordingLabel:          recordingName,
+			profilerecordingv1.ProfileToRecordingNamespaceLabel: recordingNamespace,
+		}); err != nil {
 		return nil, fmt.Errorf("listing recorded profiles for %s: %w", recordingName, err)
 	}
 
--- a/api/profilebase/v1/profilebase.go
+++ b/api/profilebase/v1/profilebase.go
@@ -36,7 +36,8 @@ type SecurityProfileBase interface {
 	client.Object
 
 	ListProfilesByRecording(
-		ctx context.Context, cli client.Client, recording string, recordingNamespace string,
+		ctx context.Context, cli client.Client,
+		recording, recordingNamespace string,
 	) ([]metav1.Object, error)
 	IsPartial() bool
 	IsDisabled() bool
--- a/api/profilebase/v1/profilebase.go
+++ b/api/profilebase/v1/profilebase.go
@@ -35,7 +35,9 @@ const ProfilePartialLabel = "spo.x-k8s.io/partial"
 type SecurityProfileBase interface {
 	client.Object
 
-	ListProfilesByRecording(ctx context.Context, cli client.Client, recording string, recordingNamespace string) ([]metav1.Object, error)
+	ListProfilesByRecording(
+		ctx context.Context, cli client.Client, recording string, recordingNamespace string,
+	) ([]metav1.Object, error)
 	IsPartial() bool
 	IsDisabled() bool
 	IsReconcilable() bool
--- a/internal/pkg/manager/recordingmerger/merge_utils.go
+++ b/internal/pkg/manager/recordingmerger/merge_utils.go
@@ -39,7 +39,8 @@ func mergedObjectMeta(profileName, recordingName, namespace string) *metav1.Obje
 		Name:      profileName,
 		Namespace: namespace,
 		Labels: map[string]string{
-			profilerecordingapi.ProfileToRecordingLabel: recordingName,
+			profilerecordingapi.ProfileToRecordingLabel:          recordingName,
+			profilerecordingapi.ProfileToRecordingNamespaceLabel: namespace,
 		},
 	}
 }
--- a/internal/pkg/manager/recordingmerger/merge_utils.go
+++ b/internal/pkg/manager/recordingmerger/merge_utils.go
@@ -85,10 +85,10 @@ func listPartialProfiles(
 	if err := cli.List(
 		ctx,
 		list,
-		client.InNamespace(recording.Namespace),
 		client.MatchingLabels{
-			profilerecordingapi.ProfileToRecordingLabel: recording.Name,
-			profilebase.ProfilePartialLabel:             "true",
+			profilerecordingapi.ProfileToRecordingLabel:          recording.Name,
+			profilerecordingapi.ProfileToRecordingNamespaceLabel: recording.Namespace,
+			profilebase.ProfilePartialLabel:                      "true",
 		}); err != nil {
 		return nil, fmt.Errorf("listing partial profiles for %s: %w", recording.Name, err)
 	}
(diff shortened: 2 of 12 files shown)


saschagrunert:
Thank you for the PR @mirza-src, do you mind signing the CLA?

mirza-src:
The `RecordingHasUnmergedProfiles` finalizer was also not being removed after successful merge because of using `policy.GetNamespace()` to lookup `ProfileRecording`.

internal/pkg/manager/recordingmerger/merge_utils.go
```diff
@@ -85,10 +85,10 @@ func listPartialProfiles(
 	if err := cli.List(
 		ctx,
 		list,
-		client.InNamespace(recording.Namespace),
 		client.MatchingLabels{
-			profilerecording1alpha1.ProfileToRecordingLabel: recording.Name,
-			profilebase.ProfilePartialLabel:                 "true",
+			profilerecording1alpha1.ProfileToRecordingLabel:          recording.Name,
+			profilerecording1alpha1.ProfileToRecordingNamespaceLabel: recording.Namespace,
```
saschagrunert: Worth considering: existing partial profiles created before this change won't have the `spo.x-k8s.io/recording-namespace` label. After an upgrade, those profiles become invisible to both `listPartialProfiles` and `deletePartialProfiles`, since the label filter won't match them.

In practice this may not matter much since the old `List()` with `InNamespace` was already returning empty results for these cluster-scoped resources (so merging was already broken). But `DeleteAllOf` with the old code *did* ignore the namespace and would still clean up old profiles, while after this change it won't find them either.

Depending on whether anyone has accumulated orphaned partial profiles in the wild, it might be worth either:
- Documenting this as a known upgrade consideration in the release notes
- Adding a one-time migration/cleanup (e.g., a controller that backfills the namespace label on existing partial profiles)

Not necessarily a blocker, just something to be aware of.

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -374,7 +376,9 @@ func handleRecordingFinalizer(ctx context.Context, c client.Client, pol profileb
 	}
 
 	profilerecording := &v1alpha1.ProfileRecording{}
-	recordingName := util.NamespacedName(polLabels[v1alpha1.ProfileToRecordingLabel], pol.GetNamespace())
+	recordingName := util.NamespacedName(
+		polLabels[v1alpha1.ProfileToRecordingLabel], polLabels[v1alpha1.ProfileToRecordingNamespaceLabel],
```
saschagrunert: Good fix. The old `pol.GetNamespace()` was always empty for these cluster-scoped resources, so the `Get()` call below would have failed to find the `ProfileRecording` (which is namespace-scoped). Using the label value makes this actually work.

saschagrunert (commented):
Well-explained PR with a clear root cause analysis. The approach is solid: replacing `client.InNamespace()` (which silently returns empty results for cluster-scoped resources) with a label-based filter, and adding a namespace label to disambiguate same-named recordings across namespaces.

The fix in `nodestatus.go` where `pol.GetNamespace()` (always empty for cluster-scoped profiles) is replaced with the label value is a nice catch too.

One thing worth considering inline.

internal/pkg/manager/recordingmerger/merge_utils.go
mirza-src: @saschagrunert  So I didn't considering any back-fill strategy because I think the merge was previously completely broken. Even though the `DeleteAllOf` would have previously deleted the partial profiles if triggered, it was never being triggered due the early return here: https://github.com/kubernetes-sigs/security-profiles-operator/blob/7cef32ea8f369112642f56fbfb90ddab9368b563/internal/pkg/manager/recordingmerger/recordingmerger.go#L158-L163

So they were always orphaned, never merged or deleted.

mirza-src: Updated the release-notes with a relevant "action required"

api/profilerecording/v1alpha1/profilerecording_types.go
```diff
@@ -51,6 +51,9 @@ const (
 const (
 	// ProfileToRecordingLabel is the name of the ProfileRecording CR that produced this profile.
 	ProfileToRecordingLabel = "spo.x-k8s.io/recording-id"
+	// ProfileToRecordingNamespaceLabel is the namespace of the ProfileRecording CR that produced this profile.
+	// Required to disambiguate cluster-scoped profiles from recordings with the same name in different namespaces.
+	ProfileToRecordingNamespaceLabel = "spo.x-k8s.io/recording-namespace"
```
saschagrunert: let's fix that in the new v1 API now, not v1alpha1

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -335,7 +335,10 @@ func handleRecordingFinalizer(ctx context.Context, c client.Client, pol profileb
 	}
 
 	// if there are other policies recorded by the same recording, we don't need to do anything either
-	otherPolicies, err := pol.ListProfilesByRecording(ctx, c, polLabels[profilerecordingapi.ProfileToRecordingLabel])
+	otherPolicies, err := pol.ListProfilesByRecording(ctx, c,
+		polLabels[profilerecordingapi.ProfileToRecordingLabel],
+		polLabels[profilerecordingapi.ProfileToRecordingNamespaceLabel],
```
saschagrunert: For pre-existing profiles without `ProfileToRecordingNamespaceLabel`, this reads as empty string. That propagates to `ListProfilesByRecording` (matches nothing) and to `util.NamespacedName` at line 382 (empty namespace, always NotFound). The recording's `RecordingHasUnmergedProfiles` finalizer gets silently stuck.

Consider extending the guard at line 333 to also check for the namespace label:

```go
if polLabels == nil || polLabels[profilerecordingapi.ProfileToRecordingLabel] == "" ||
	polLabels[profilerecordingapi.ProfileToRecordingNamespaceLabel] == "" {
	return nil
}
```

saschagrunert (commented):
`mergedObjectMeta` (`merge_utils.go:37-45`) should also set `ProfileToRecordingNamespaceLabel`. The `namespace` parameter is already available from callers at `recordingmerger.go:302,323,343`. Without it, merged profiles lack the namespace label that all partial profiles carry.

```go
func mergedObjectMeta(profileName, recordingName, namespace string) *metav1.ObjectMeta {
	return &metav1.ObjectMeta{
		Name:      profileName,
		Namespace: namespace,
		Labels: map[string]string{
			profilerecordingapi.ProfileToRecordingLabel:          recordingName,
			profilerecordingapi.ProfileToRecordingNamespaceLabel: namespace,
		},
	}
}
```

(Can't post this as an inline comment since the function is outside the diff.)

api/profilebase/v1/profilebase.go
```diff
@@ -35,7 +35,9 @@ const ProfilePartialLabel = "spo.x-k8s.io/partial"
 type SecurityProfileBase interface {
 	client.Object
 
-	ListProfilesByRecording(ctx context.Context, cli client.Client, recording string) ([]metav1.Object, error)
+	ListProfilesByRecording(
+		ctx context.Context, cli client.Client, recording string, recordingNamespace string,
+	) ([]metav1.Object, error)
```
saschagrunert, line 41: Nit: group parameters of the same type for consistency with the implementations.

```suggestion
	ListProfilesByRecording(
		ctx context.Context, cli client.Client,
		recording, recordingNamespace string,
	) ([]metav1.Object, error)
```

saschagrunert (approved):

PR #3240, opened 2026-06-10
api: fix comment casing, add required/optional and enum markers, integrate kal

saschagrunert:
#### What type of PR is this?

/kind cleanup
/kind api-change

#### What this PR does / why we need it:

Aligns API types with Kubernetes API conventions by enabling kube-api-linter (kal) checks. Part of the v1 API graduation effort.

Changes:
- Fix comment casing to lowercase for all API field descriptions
- Add `+required`/`+optional` markers to all API fields
- Switch from custom `common.Condition` to `metav1.Condition`
- Add `+kubebuilder:validation:Enum` markers on type alias fields (Action, Operator)
- Add `omitempty`/`omitzero` tags per requiredfields linter
- Make `Arg.Index` a pointer to allow valid zero value
- Fix `PodSelector` to pointer type for optional field
- Add `+required` with `omitzero` on CRD Spec fields
- Enable `requiredfields`, `conditions`, and `nonpointerstructs` kal linters

Remaining disabled kal linters: `nomaps`, `optionalfields`

#### Which issue(s) this PR fixes:

None

#### Does this PR have test?

Existing tests updated to match API changes.

#### Special notes for your reviewer:

Breaking API changes are expected. These will be handled during the v1 conversion phase.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #3243, opened 2026-06-10
fix: add default selinux options so allowedSystemProfiles is applied

mirza-src:
#### What type of PR is this?

/kind bug

`allowedSystemProfiles` was never getting its default value `["container"]` when deploying via Helm. Kubernetes only applies child defaults if the parent object is present, since `selinux.options` is absent from the helm template, the API server never applied the child default.

Fix: add `+kubebuilder:default={}` on `Options` so the API server creates the parent and applies the child default.

None

#### Does this PR have test?

N/A

None

```release-note
Fixed `allowedSystemProfiles` default (`["container"]`) not being applied when installing via Helm.
```

### DIFF
--- a/api/spod/v1alpha1/spod_types.go
+++ b/api/spod/v1alpha1/spod_types.go
@@ -166,6 +166,7 @@ type SPODSelinuxConfig struct {
 	TypeTag string `json:"typeTag,omitempty"`
 	// options defines options specific to the SELinux functionality.
 	// +optional
+	// +default={}
 	Options SelinuxOptions `json:"options,omitzero,omitempty"`
 }
 
--- a/deploy/base-crds/crds/securityprofilesoperatordaemon.yaml
+++ b/deploy/base-crds/crds/securityprofilesoperatordaemon.yaml
@@ -1266,6 +1266,7 @@ spec:
                       controller will not be started. Defaults to true when SELinux is enabled.
                     type: boolean
                   options:
+                    default: {}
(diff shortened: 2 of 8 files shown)


ccojocar (approved):

saschagrunert:
/ok-to-test

/hold 

for the larger API change in https://github.com/kubernetes-sigs/security-profiles-operator/pull/3240

api/spod/v1alpha1/spod_types.go
```diff
@@ -165,6 +165,7 @@ type SPODSelinuxConfig struct {
 	TypeTag string `json:"typeTag,omitempty"`
 	// options defines options specific to the SELinux functionality.
 	// +optional
+	// +kubebuilder:default={}
```
saschagrunert: Nit: the rest of this file uses the shorter `+default=` form (e.g. `+default=["container"]` on line 35, `+default="spc_t"` on line 165). Would be nice to stay consistent:
```suggestion
	// +default={}
```

saschagrunert (approved):
Nice find. The fix correctly addresses the Helm defaulting gap: the Helm template (`deploy/helm/templates/config.yaml`) always sets `selinux.enable`, so the `selinux` parent object exists, but `options` was absent, preventing the API server from applying the child `allowedSystemProfiles` default. Adding `default: {}` on `options` closes that gap. CRD changes across all deploy manifests are consistent.

One small style nit inline.

saschagrunert (approved):

PR #3244, opened 2026-06-11
Webhook should overwrite an existing recording annotation on a pod

ccojocar:
#### What type of PR is this?

/kind bug

The webhook should overwrite an existing profile recording annotation
with the expected profile name, instead of just skipping it. This can
allow an attacker to spoof the profile recording into its own profile is
one manually applies the recording annotation to the pod.

```release-note
Webhook always overwrite the values of existing profile recording annotation on a pod to avoid
profile recording spoofing.
```

### DIFF
--- a/internal/pkg/webhooks/recording/recording.go
+++ b/internal/pkg/webhooks/recording/recording.go
@@ -235,12 +235,13 @@ func (p *podSeccompRecorder) updatePod(
 
 		if existingValue != value {
 			// Overwrite the existing value with the expected value to avoid that
-			// an attacker will spoof a profile recording into it's own controlled
+			// an attacker will spoof a profile recording into its own controlled
 			// profile instead of the one expected.
 			pod.Annotations[key] = value
+			podChanged = true
 			p.log.Info(
 				fmt.Sprintf(
-					"workload %s already has annotation %q is overwritten by %q.",
+					"workload %s already has annotation %q, overwriting with %q.",
 					podName,
 					existingValue,
 					value,
--- a/internal/pkg/webhooks/recording/recording.go
+++ b/internal/pkg/webhooks/recording/recording.go
@@ -239,6 +239,7 @@ func (p *podSeccompRecorder) updatePod(
 			// profile instead of the one expected.
 			pod.Annotations[key] = value
 			podChanged = true
+
 			p.log.Info(
 				fmt.Sprintf(
 					"workload %s already has annotation %q, overwriting with %q.",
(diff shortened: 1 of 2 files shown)


saschagrunert (dismissed):
The security fix is the right call, but the overwrite is never actually applied due to a missing `podChanged = true`. See inline comments.

internal/pkg/webhooks/recording/recording.go
```diff
@@ -235,10 +234,13 @@ func (p *podSeccompRecorder) updatePod(
 		}
 
 		if existingValue != value {
-			p.log.Error(
-				errors.New("existing annotation"),
+			// Overwrite the existing value with the expected value to avoid that
+			// an attacker will spoof a profile recording into it's own controlled
+			// profile instead of the one expected.
+			pod.Annotations[key] = value
```
saschagrunert, line 240: I think we also need `podChanged = true` here, otherwise the caller (line ~163) sees `!podChanged` and returns `admission.Allowed("pod unchanged")` without generating a patch. The annotation gets overwritten on the Go object but the mutation never reaches the API server.

The "annotation doesn't exist" branch above (line ~232) sets it, so this one should too:
```suggestion
			pod.Annotations[key] = value
			podChanged = true
```

internal/pkg/webhooks/recording/recording.go
```diff
@@ -235,10 +234,13 @@ func (p *podSeccompRecorder) updatePod(
 		}
 
 		if existingValue != value {
-			p.log.Error(
-				errors.New("existing annotation"),
+			// Overwrite the existing value with the expected value to avoid that
+			// an attacker will spoof a profile recording into it's own controlled
```
saschagrunert: Tiny nit: "it's" -> "its" (possessive).
```suggestion
			// an attacker will spoof a profile recording into its own controlled
```

internal/pkg/webhooks/recording/recording.go
```diff
@@ -235,10 +234,13 @@ func (p *podSeccompRecorder) updatePod(
...
+			// profile instead of the one expected.
+			pod.Annotations[key] = value
+			p.log.Info(
 				fmt.Sprintf(
-					"workload %s already has annotation %s (not mutating to %s).",
+					"workload %s already has annotation %q is overwritten by %q.",
```
saschagrunert: Minor grammar nit:
```suggestion
					"workload %s already has annotation %q, overwriting with %q.",
```

saschagrunert (changes requested):
Good catch on the spoofing vector! The approach of overwriting instead of skipping is the right fix. Just one thing that needs addressing before this works end-to-end, plus a couple of small nits.

ccojocar:
Thanks for review. I think the suggestions are addressed.

saschagrunert (approved):

saschagrunert (approved):

PR #3245, opened 2026-06-11
Drop the labels with unbounded cardinlity from prometheus counters

ccojocar:
#### What type of PR is this?

/kind bug

The "executable" and "name" labels have unbounded cardinality in the
prometheus counter which could cause memory exhaustion in both side
the spod exposing the metric and the Prometheus server.

```release-note
Remove "executable" and "name" from Prometheus counter because they could have unbounded cardinality
and could lead to memory exhaustion.
```

### DIFF
--- a/internal/pkg/daemon/metrics/grpc.go
+++ b/internal/pkg/daemon/metrics/grpc.go
@@ -100,7 +100,7 @@ func (m *Metrics) AuditInc(
 				r.GetNamespace(),
 				r.GetPod(),
 				r.GetContainer(),
-				r.GetExecutable(),
+				r.GetSeccompReq().GetSyscall(),
 			)
 		} else if r.GetSelinuxReq() != nil {
 			m.IncSelinuxProfileAudit(
@@ -108,7 +108,6 @@ func (m *Metrics) AuditInc(
 				r.GetNamespace(),
 				r.GetPod(),
 				r.GetContainer(),
-				r.GetExecutable(),
 				r.GetSelinuxReq().GetScontext(),
 				r.GetSelinuxReq().GetTcontext(),
 			)
--- a/internal/pkg/daemon/metrics/grpc.go
+++ b/internal/pkg/daemon/metrics/grpc.go
@@ -101,7 +101,6 @@ func (m *Metrics) AuditInc(
 				r.GetPod(),
 				r.GetContainer(),
 				r.GetExecutable(),
-				r.GetSeccompReq().GetSyscall(),
 			)
 		} else if r.GetSelinuxReq() != nil {
 			m.IncSelinuxProfileAudit(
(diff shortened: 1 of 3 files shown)


internal/pkg/daemon/metrics/grpc.go
```diff
@@ -101,7 +101,6 @@ func (m *Metrics) AuditInc(
 				r.GetPod(),
 				r.GetContainer(),
 				r.GetExecutable(),
```
saschagrunert: Wrong line removed: `r.GetSeccompReq().GetSyscall()` was deleted (line above) but `r.GetExecutable()` was kept. After the function signature change to `(node, namespace, pod, container, syscall string)`, the executable value gets silently recorded under the `syscall` label.

This should be the other way around:
```suggestion
				r.GetSeccompReq().GetSyscall(),
```

internal/pkg/daemon/metrics/metrics.go
```diff
@@ -273,10 +271,10 @@ func (m *Metrics) IncSeccompProfileDelete() {
 // IncSeccompProfileAudit increments the seccomp profile audit counter for the
 // provided labels.
 func (m *Metrics) IncSeccompProfileAudit(
-	node, namespace, pod, container, executable, syscall string,
+	node, namespace, pod, container, syscall string,
 ) {
 	m.metricSeccompProfileAudit.WithLabelValues(
-		node, namespace, pod, container, executable, syscall,
+		node, namespace, pod, container, syscall,
```
saschagrunert, line 274: This will panic at runtime. The `metricSeccompProfileAudit` counter definition (line ~116-123 in this file) still has `metricsLabelExecutable` in its label list, giving it 6 labels. But only 5 values are passed here. Prometheus requires an exact match and will panic on the mismatch.

The counter definition needs `metricsLabelExecutable` removed from its label list too.

saschagrunert (changes requested):
The goal is right: `executable` and `name` are unbounded-cardinality labels that can cause OOM. But the seccomp changes have two bugs that will cause a runtime panic and silently corrupt metric data (see inline comments).

Two additional items not in the diff:

- **Selinux audit counter not addressed**: `metricSelinuxProfileAudit` and `IncSelinuxProfileAudit` still include `executable` with the same unbounded cardinality concern. Worth addressing here or as a tracked follow-up.
- **`metricsLabelName` constant now unused**: after removing it from the AppArmor counter, the constant at line 70 has no remaining references.

ccojocar:
@saschagrunert thanks for catching these issues. I think that they are fixed now.

saschagrunert (approved):

saschagrunert:
Oh build fails
/lgtm cancel

```
       > internal/pkg/daemon/metrics/grpc.go:112:5: not enough arguments in call to m.IncSelinuxProfileAudit
       >   have (string, string, string, string, string, string)
       >   want (string, string, string, string, string, string, string)
       > make: *** [Makefile:166: build/security-profiles-operator] Error 1
       For full logs, run:
```

saschagrunert (approved):

PR #3246, opened 2026-06-11
Cap the maximum number of recorded files and the file length in eBpf recorder to avoid OOM

ccojocar:
#### What type of PR is this?

/kind bug

Cap the maximum number of recorded files and the file path length in the eBPF recorder to avoid OOM.

```release-note
Cap the maximum recorded files and the file legth to avoid OOM in eBpf recorder.
```

### DIFF
--- a/internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
+++ b/internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
@@ -183,16 +183,18 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
 	}
 
 	// Enforce a limit on max tracked files to avoid OOM.
-	if len(b.recordedFiles[mid]) > maxTrackedPaths {
-		b.logger.Info("Max tracked file reached. Profile will be truncated to avoid OOM",
-			"mntns", mid, "limit", maxTrackedPaths)
+	if len(b.recordedFiles[mid]) >= maxTrackedPaths {
+		b.logger.Info("Max tracked files reached. Profile will be truncated to avoid OOM",
+			"level", "warn", "mntns", mid, "limit", maxTrackedPaths)
 
 		return
 	}
 
 	// Cap path length to prevent single-string memory bloat
 	if len(fileName) > maxPathLength {
-		fileName = fileName[:maxPathLength]
+		b.logger.Info("Skipping file with excessively long path",
+			"level", "warn", "length", len(fileName), "mntns", mid)
+		return
 	}
 
 	path, ok := b.recordedFiles[mid][fileName]
--- a/internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
+++ b/internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
@@ -185,7 +185,7 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
 	// Enforce a limit on max tracked files to avoid OOM.
 	if len(b.recordedFiles[mid]) >= maxTrackedPaths {
 		b.logger.Info("Max tracked files reached. Profile will be truncated to avoid OOM",
-			"level", "warn", "mntns", mid, "limit", maxTrackedPaths)
+			"mntns", mid, "limit", maxTrackedPaths)
 
 		return
 	}
@@ -193,7 +193,7 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
 	// Cap path length to prevent single-string memory bloat
 	if len(fileName) > maxPathLength {
 		b.logger.Info("Skipping file with excessively long path",
-			"level", "warn", "length", len(fileName), "mntns", mid)
+			"length", len(fileName), "mntns", mid)
 
 		return
 	}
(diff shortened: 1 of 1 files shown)


saschagrunert (commented):
The OOM mitigation concept is sound and addresses a real risk. The `log.Printf` to structured logger migration is a welcome cleanup.

A few items worth addressing, see inline comments.

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -175,6 +182,19 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
 		b.recordedFiles[mid] = map[string]*fileAccess{}
 	}
 
+	// Enforce a limit on max tracked files to avoid OOM.
+	if len(b.recordedFiles[mid]) > maxTrackedPaths {
```
saschagrunert: Off-by-one: when `len` is 10000, the condition `10000 > 10000` is false, so a new entry is still inserted, allowing 10001 entries before triggering. Should be `>=`:
```suggestion
	if len(b.recordedFiles[mid]) >= maxTrackedPaths {
```

saschagrunert (commented):
The OOM mitigation concept is sound and addresses a real risk. The `log.Printf` to structured logger migration is a welcome cleanup.

A few items worth addressing, see inline comments.

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -175,6 +182,19 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
...
+		b.logger.Info("Max tracked file reached. Profile will be truncated to avoid OOM",
```
saschagrunert: Hitting this limit means either a very busy workload or an intentional attempt to exhaust recorder memory, and the resulting profile will be incomplete. `Warn` would be more appropriate than `Info` here.

Also, grammar nit: "Max tracked file" should be "Max tracked files" (plural).

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -175,6 +182,19 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
...
+			"mntns", mid, "limit", maxTrackedPaths)
+
+		return
+	}
+
+	// Cap path length to prevent single-string memory bloat
+	if len(fileName) > maxPathLength {
+		fileName = fileName[:maxPathLength]
```
saschagrunert: Slicing at an arbitrary byte offset produces an invalid filesystem path (e.g. `/very/long/.../some_fi`) that ends up in the AppArmor profile. The resulting rule would either never match or could match something unintended.

It would be safer to skip the file entirely with a warning rather than recording a corrupted path:
```suggestion
		b.logger.Info("Skipping file with excessively long path", "length", len(fileName), "mntns", mid)
		return
```

The practical risk is low since BPF data is already capped at `pathMax` (4096, the same value), and `ReplaceVarianceInFilePath` expansions are minimal.

ccojocar:
@saschagrunert I think that I addressed your suggestions. Thanks

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -175,6 +182,22 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
 		b.recordedFiles[mid] = map[string]*fileAccess{}
 	}
 
+	// Enforce a limit on max tracked files to avoid OOM.
+	if len(b.recordedFiles[mid]) >= maxTrackedPaths {
+		b.logger.Info("Max tracked files reached. Profile will be truncated to avoid OOM",
+			"level", "warn", "mntns", mid, "limit", maxTrackedPaths)
```
saschagrunert: The `logr` interface has no `Warn()` method, so `"level", "warn"` is just a regular key-value pair in the structured output, not an actual level change. A log aggregator will see `level=info` in the metadata alongside a confusing `level=warn` data field. Drop it:

```suggestion
		b.logger.Info("Max tracked files reached, profile will be truncated to avoid OOM",
			"mntns", mid, "limit", maxTrackedPaths)
```

internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
```diff
@@ -175,6 +182,22 @@ func (b *AppArmorRecorder) handleFileEvent(fileEvent *bpfEvent) {
...
+
+		return
+	}
+
+	// Cap path length to prevent single-string memory bloat
+	if len(fileName) > maxPathLength {
+		b.logger.Info("Skipping file with excessively long path",
+			"level", "warn", "length", len(fileName), "mntns", mid)
```
saschagrunert: Same here, drop the misleading `"level", "warn"` pair:

```suggestion
		b.logger.Info("Skipping file with excessively long path",
			"length", len(fileName), "mntns", mid)
```

saschagrunert (changes requested):
The `log.Printf` to structured logger migration and the OOM guards look good. Two small fixes needed: the `"level", "warn"` key-value pairs passed to `b.logger.Info()` do not change the actual log level in logr, they just add a misleading field to the structured output. They should be dropped.

Also, release note typo: "file legth" -> "file length", "eBpf" -> "eBPF".

ccojocar:
@saschagrunert I addressed the review comments. Thanks

saschagrunert (approved):

PR #3247, opened 2026-06-11
Combine the pid with the start time of the process when caching the container ID

ccojocar:
#### What type of PR is this?

/kind bug

This is a mitigation to avoid a fork-bomb attack when a PID is reused
within the TTL of the cached container ID to introduce malicious content
within the profile recording.

```release-note
Mitigate PID reuse attack in container ID caching by combining PID with process start time as the cache key
```

### DIFF
--- a/internal/pkg/util/containers.go
+++ b/internal/pkg/util/containers.go
@@ -108,7 +108,6 @@ func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string, 
 
 // getProcessStartTimeTicks return the start time for a process.
 func getProcessStartTimeTicks(pid int, procStatReader procStatReader) (string, error) {
-	//stat, err := os.ReadFile(fmt.Sprintf("/proc/%d/stat", pid))
 	stat, err := procStatReader(pid)
 	if err != nil {
 		return "", fmt.Errorf("reading proc start time for %d pid: %w", pid, err)
--- a/internal/pkg/util/containers.go
+++ b/internal/pkg/util/containers.go
@@ -110,6 +110,7 @@ func getProcessStartTime(pid int) (string, error) {
 	// The comm field (field 2) is in parentheses and can contain spaces,
 	// so split after the last ")" to get reliable field indices.
 	raw := string(stat)
+
 	i := strings.LastIndexByte(raw, ')')
 	if i < 0 || i+2 >= len(raw) {
 		return "", fmt.Errorf("invalid proc stat format for pid: %d", pid)
@@ -120,5 +121,6 @@ func getProcessStartTime(pid int) (string, error) {
 	if len(fields) < 20 {
 		return "", fmt.Errorf("invalid proc stat format for pid: %d", pid)
 	}
+
 	return fields[19], nil
 }
--- a/internal/pkg/util/containers.go
+++ b/internal/pkg/util/containers.go
@@ -42,22 +42,28 @@ var (
 	// cgroup does not contain any container ID.
 	ErrContainerIDNotFound = errors.New("unable to find container ID in cgroup path")
 
-	// ErrContainerIDSearchFailed is the error returned with a reason by
+	// errContainerIDSearchFailed is the error returned with a reason by
 	// ContainerIDForPID if it fails to parse the container ID from the logs.
-	ErrContainerIDSearchFailed = errors.New("failed looking for container ID")
+	errContainerIDSearchFailed = errors.New("failed looking for container ID")
 )
+
+// procStatReader closure function to read the /proc/<pid>stat which allows for
+// more testability.
+type procStatReader func(pid int) ([]byte, error)
 
 // ContainerIDForPID tries to find the 64 digit container ID for the provided
 // PID by using its cgroup. It supports caching via the cache argument.
 func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string, error) {
-	stat, err := getProcessStartTime(pid)
+	startTime, err := getProcessStartTimeTicks(pid, func(pid int) ([]byte, error) {
+		return os.ReadFile(fmt.Sprintf("/proc/%d/stat", pid))
+	})
 	if err != nil {
 		return "", fmt.Errorf("reading proc start time: %w", err)
 	}
 
 	// Combine the pid with the process start time as a cache key to avoid "fork-bomb"
 	// attack which reuses a PID for a different container within the cache TTL.
-	cacheKey := strconv.Itoa(pid) + "_" + stat
+	cacheKey := strconv.Itoa(pid) + "_" + startTime
 
 	// Check the cache first
 	item := cache.Get(cacheKey)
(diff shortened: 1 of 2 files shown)


internal/pkg/util/containers.go
```diff
@@ -40,13 +41,24 @@ var (
 	// ErrContainerIDNotFound is the error returned by ContainerIDForPID if the
 	// cgroup does not contain any container ID.
 	ErrContainerIDNotFound = errors.New("unable to find container ID in cgroup path")
+
+	ErrContainerIDSearchFailed = errors.New("failed looking for container ID")
```
saschagrunert: Exported error variable missing a doc comment, unlike the others in this block.

```suggestion
	// ErrContainerIDSearchFailed is the error returned by ContainerIDForPID if
	// scanning the cgroup file fails.
	ErrContainerIDSearchFailed = errors.New("failed looking for container ID")
```

internal/pkg/util/containers.go
```diff
@@ -40,13 +41,24 @@ var (
...
 // ContainerIDForPID tries to find the 64 digit container ID for the provided
 // PID by using its cgroup. It supports caching via the cache argument.
 func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string, error) {
+	stat, err := getProcessStartTime(pid)
+	if err != nil {
+		return "", fmt.Errorf("reading proc start time: %w", err)
+	}
+
+	// Combine the pid with the process start time as a cache key to avoid "fork-bomb"
+	// attack which reuse a PID for a different container within the cache TTL.
```
saschagrunert: nit: grammar

```suggestion
	// Combine the pid with the process start time as a cache key to avoid a "fork-bomb"
	// attack which reuses a PID for a different container within the cache TTL.
```

internal/pkg/util/containers.go
```diff
@@ -73,11 +85,28 @@ func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string,
...
+	stat, err := os.ReadFile(fmt.Sprintf("/proc/%d/stat", pid))
+	if err != nil {
+		return "", fmt.Errorf("reading proc start time for %d pid: %w", pid, err)
+	}
+	fields := strings.Fields(string(stat))
+	if len(fields) < 22 {
+		return "", fmt.Errorf("invalid proc stat format for pid: %d", pid)
+	}
+	return fields[21], nil
+}
```
saschagrunert: Bug: `strings.Fields()` splits on whitespace, but the `comm` field (field 2) in `/proc/PID/stat` is parenthesized and can contain spaces. For example a process named `my app` produces:

```
1234 (my app) S 1233 1234 ...
```

`strings.Fields()` turns `(my` and `app)` into separate tokens, shifting every subsequent index. `fields[21]` then returns the wrong value or panics on short lines.

The standard fix is to find the last `)` of the comm field and split from there:

```suggestion
// getProcessStartTime returns the start time for a process.
func getProcessStartTime(pid int) (string, error) {
	stat, err := os.ReadFile(fmt.Sprintf("/proc/%d/stat", pid))
	if err != nil {
		return "", fmt.Errorf("reading proc start time for %d pid: %w", pid, err)
	}
	// The comm field (field 2) is in parentheses and can contain spaces,
	// so split after the last ")" to get reliable field indices.
	raw := string(stat)
	i := strings.LastIndexByte(raw, ')')
	if i < 0 || i+2 >= len(raw) {
		return "", fmt.Errorf("invalid proc stat format for pid: %d", pid)
	}
	// After ")" the fields are: state(3) ppid(4) ... starttime(22),
	// which is index 19 in the zero-based slice after ")".
	fields := strings.Fields(raw[i+2:])
	if len(fields) < 20 {
		return "", fmt.Errorf("invalid proc stat format for pid: %d", pid)
	}
	return fields[19], nil
}
```

ccojocar:
@saschagrunert I addressed your suggestions. Thanks

internal/pkg/util/containers.go
```diff
@@ -40,13 +41,26 @@ var (
 	// ErrContainerIDNotFound is the error returned by ContainerIDForPID if the
 	// cgroup does not contain any container ID.
 	ErrContainerIDNotFound = errors.New("unable to find container ID in cgroup path")
+
+	// ErrContainerIDSearchFailed is the error returned with a reason by
+	// ContainerIDForPID if it fails to parse the container ID from the logs.
+	ErrContainerIDSearchFailed = errors.New("failed looking for container ID")
```
saschagrunert: nit: This is exported but no caller checks for it. All callers use either `errors.Is(err, os.ErrNotExist)` or a generic `err != nil` check. Consider keeping it unexported (`errContainerIDSearchFailed`) unless there is a planned consumer.

internal/pkg/util/containers.go
```diff
@@ -40,13 +41,26 @@ var (
...
 )
 
 // ContainerIDForPID tries to find the 64 digit container ID for the provided
 // PID by using its cgroup. It supports caching via the cache argument.
 func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string, error) {
+	stat, err := getProcessStartTime(pid)
```
saschagrunert: nit: `stat` here holds the start time string (clock ticks), but in `getProcessStartTime` the same name `stat` holds the full file content. Consider naming this `startTime` at the call site to make the cache key construction easier to follow:
```suggestion
	startTime, err := getProcessStartTime(pid)
```

internal/pkg/util/containers.go
```diff
@@ -73,11 +87,40 @@ func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string,
...
+	// Check if not finding a container ID was caused by any scanning errors.
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("%w: %w", ErrContainerIDSearchFailed, err)
+	}
 
 	return "", ErrContainerIDNotFound
 }
+
+// getProcessStartTime return the start time for a process.
+func getProcessStartTime(pid int) (string, error) {
```
saschagrunert: nit: The function returns a raw string of kernel clock ticks, not a parsed time value. Consider renaming to `getProcessStartTimeTicks` or adjusting the doc comment to clarify the return format.

internal/pkg/util/containers.go
```diff
@@ -73,11 +87,40 @@ func ContainerIDForPID(cache *ttlcache.Cache[string, string], pid int) (string,
...
+	}
+	// After ")" the fields are: state(3) ppid(4) ... starttime(22),
+	// which is index 19 in the zero-based slice after ")".
+	fields := strings.Fields(raw[i+2:])
+	if len(fields) < 20 {
+		return "", fmt.Errorf("invalid proc stat format for pid: %d", pid)
+	}
+
+	return fields[19], nil
+}
```
saschagrunert, line 132: Since this is a security mitigation, consider adding unit tests for:
- The core property: same PID with different start times produces different cache keys (cache miss)
- `getProcessStartTime` parsing with edge-case comm fields (spaces, nested parens)
- Error paths: missing `/proc/<pid>/stat`, truncated stat output

saschagrunert (commented):
Overall the mitigation approach is sound: combining PID with process start time prevents PID-reuse attacks within the cache TTL. The `/proc/<pid>/stat` parsing is correctly implemented (handles comm fields with spaces/parens, correct field index for starttime). A few suggestions below.

ccojocar:
@saschagrunert I addressed the review comments. Thanks

saschagrunert (approved):

PR #3251, opened 2026-06-15
Fix data races, OOM guards, and performance in BPF recorder and container ID cache

saschagrunert:
#### What type of PR is this?

/kind bug
/kind cleanup

Follow-up fixes from #3247 and #3246:
- Fix data race in `GetAppArmorProcessed`: socket and capability map access lacked locks while event handlers write concurrently. Fold cleanup deletes into the locked processing methods.
- Fix lock ordering inconsistency: align `clearMntns` and `GetKnownMntns` to acquire locks in the same order as `StopRecording` (Sockets, Capabilities, Files) to prevent potential deadlocks.
- Prevent log spam when `maxTrackedPaths` limit is hit: log only once per mount namespace instead of on every subsequent event.
- Cap tracked mount namespaces at 1000 to prevent unbounded map growth when many containers start concurrently.
- Move 8 `regexp.MustCompile` calls from `ReplaceVarianceInFilePath` (called per file event) to package-level vars.
- Change per-event "File access" and "Exclude file" logs from `Info` to `V(config.VerboseLevel)` to match other hot-path logging in the recorder.
- Remove dead `maxPathLength` check (redundant with BPF buffer `pathMax=4096`).
- Add integration tests for `ContainerIDForPID` cache behavior (PID reuse produces cache miss, same PID+starttime produces cache hit).
- Refactor `ContainerIDForPID` for testability by extracting internal function with injected readers.

None

#### Does this PR have test?

Yes, added `TestContainerIDForPID_CacheKeyWithStartTime` and `TestContainerIDForPID_CacheHit`.

```release-note
NONE
```

### DIFF
--- a/internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
+++ b/internal/pkg/daemon/bpfrecorder/bpfrecorder_apparmor.go
@@ -28,6 +28,8 @@ import (
 	"sync"
 
 	"github.com/go-logr/logr"
+
+	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


saschagrunert:

ccojocar (approved):

PR #3252, opened 2026-06-15
Verify the base profile OIDC on immutable image digest and add options to spod config to setup allowed identity and Oidc Issuer regexp(s)

ccojocar:
#### What type of PR is this?

/kind bug

Make sure that the signature verification of the base profile OIDC image is done on the immutable
image digest to avoid a TOCTOU attack on the immutable tag between verification and copy.

Adds two new options in the spod configuration to configure the allowed identities and allowed
Oidc issuer regexp(s) instead of allowing any signed image without an option to restrict it.

Refactor the puller interface to pass these options as single option struct, and extended the 
puller CLI command to allow them to be configured as CLI arguments.

#### Does this PR have test?

Yes

```release-note
Add two new flags in spod configuration to `AllowedIdentityRegexp` and `AllowedOidcIssuerRegexp` to
restrict which identities and Oidc issuers are allowed during the image verification.
```

### DIFF
--- a/api/spod/v1/spod_types.go
+++ b/api/spod/v1/spod_types.go
@@ -267,7 +267,7 @@ type SPODSecurityConfig struct {
 	// +default=".*"
 	AllowedIdentityRegexp string `json:"allowedIdentityRegexp,omitempty"`
 
-	// allowedOidcIssuerRegexp regexp for allowed Oidc issuer when verifying the signature of OIDC
+	// allowedOidcIssuerRegexp regexp for allowed Oidc issuer when verifying the signature of OCI
 	// image used to distribute the base profile in the cluster.
 	// +optional
 	// +default=".*"
--- a/api/spod/v1/spod_types.go
+++ b/api/spod/v1/spod_types.go
@@ -260,6 +260,18 @@ type SPODSecurityConfig struct {
 	// +optional
 	// +default=false
 	DisableOCIArtifactSignatureVerification *bool `json:"disableOciArtifactSignatureVerification,omitempty"`
+
+	// allowedIdentityRegexp regexp for allowed identity when verifying the signature of OCI
+	// image used to distribute the base profile in the cluster.
+	// +optional
+	// +default=".*"
+	AllowedIdentityRegexp string `json:"allowedIdentityRegexp,omitempty"`
+
+	// allowedOidcIssuerRegexp regexp for allowed Oidc issuer when verifying the signature of OIDC
+	// image used to distribute the base profile in the cluster.
+	// +optional
+	// +default=".*"
+	AllowedOidcIssuerRegexp string `json:"allowedOidcIssuerRegexp,omitempty"`
 }
 
 // SPODState defines the state that the spod is in.
--- a/internal/pkg/artifact/artifact.go
+++ b/internal/pkg/artifact/artifact.go
@@ -83,6 +83,18 @@ type Artifact struct {
 	logger logr.Logger
 }
 
+// PullSignatureOptions options for verifying the OCI image signature during pulling.
+type PullSignatureOptions struct {
+	// DisableSignatureVerification disables signature verification during pulling.
+	DisableSignatureVerification bool
+
+	// AllowedIdentityRegexp regexp for allowed identities for signature verification.
+	AllowedIdentityRegexp string
+
+	// AllowedOidcIssuerRegexp regexp for allowed Oidc issuer for signature verification.
+	AllowedOidcIssuerRegexp string
+}
+
 // New returns a new Artifact instance.
 func New(logger logr.Logger) *Artifact {
 	return &Artifact{
@@ -266,7 +278,7 @@ func (a *Artifact) Pull(
 	c context.Context,
 	from, username, password string,
 	platform *v1.Platform,
-	disableSignatureVerification bool,
+	signOpts *PullSignatureOptions,
 ) (*PullResult, error) {
 	ctx, cancel := context.WithTimeout(c, defaultTimeout)
 	defer cancel()
@@ -282,15 +294,13 @@ func (a *Artifact) Pull(
 		return nil, fmt.Errorf("resolving digest for image %q: %w", from, err)
 	}
 
-	if !disableSignatureVerification {
+	if signOpts != nil && !signOpts.DisableSignatureVerification {
 		a.logger.Info("Verifying signature")
-
-		const all = ".*"
 
 		v := verify.VerifyCommand{
 			CertVerifyOptions: options.CertVerifyOptions{
-				CertIdentityRegexp:   all,
-				CertOidcIssuerRegexp: all,
+				CertIdentityRegexp:   signOpts.AllowedIdentityRegexp,
+				CertOidcIssuerRegexp: signOpts.AllowedOidcIssuerRegexp,
 			},
 		}
 		if err := a.VerifyCmd(ctx, v, from); err != nil {
--- a/internal/pkg/artifact/artifact.go
+++ b/internal/pkg/artifact/artifact.go
@@ -271,11 +271,12 @@ func (a *Artifact) Pull(
 	ctx, cancel := context.WithTimeout(c, defaultTimeout)
 	defer cancel()
 
+	a.logger.Info("Resolving digest of image: " + from)
+
 	// Retrieve the immutable image digest before doing any verification to
 	// prevent a TOCTOU attack on the mutable tag of the base image, which
 	// it might lead to a malicious based profile being injected between
-	// verification and copying the content
-	a.logger.Info("Resolving digest of image: " + from)
+	// verification and copying the content.
 	from, repo, sha, err := a.imageWithDigest(ctx, from, username, password)
 	if err != nil {
 		return nil, fmt.Errorf("resolving digest for image %q: %w", from, err)
--- a/internal/pkg/artifact/artifact.go
+++ b/internal/pkg/artifact/artifact.go
@@ -276,7 +276,7 @@ func (a *Artifact) Pull(
 	// it might lead to a malicious based profile being injected between
 	// verification and copying the content
 	a.logger.Info("Resolving digest of image: " + from)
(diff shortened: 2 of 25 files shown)


internal/pkg/artifact/artifact.go
```diff
@@ -395,6 +388,46 @@ func (a *Artifact) Pull(
...
+	}
+
+	desc, err := a.ResolveRepository(ctx, repo, ref.Identifier())
+	if err != nil {
+		return "", nil, "",
+			fmt.Errorf("resolving image identifier %q: %w", ref.Identifier(), err)
+	}
+
+	return fmt.Sprintf("%s@%s", ref.Name(),
+		desc.Digest.String()), repo, desc.Digest, nil
```
saschagrunert, line 439: Bug: `ref.Name()` includes the tag/digest from the original input. For a tag input `registry.io/repo:v1` this produces `registry.io/repo:v1@sha256:abc`. For a digest input `registry.io/repo@sha256:abc` this produces `registry.io/repo@sha256:abc@sha256:abc` (invalid, two `@` separators).

Use `ref.Context().Name()` to get just the repository, matching the pattern the old code used:

```suggestion
	return fmt.Sprintf("%s@%s", ref.Context().Name(),
		desc.Digest.String()), repo, desc.Digest, nil
```

internal/pkg/artifact/artifact.go
```diff
@@ -395,6 +388,46 @@ func (a *Artifact) Pull(
...
+// digest and the repository and the digest as separate return arguments.
+func (a *Artifact) imageWithDigest(ctx context.Context, image, username, password string) (
+	string, *remote.Repository, digest.Digest, error,
+) {
+	ref, err := a.ParseReference(image)
+	if err != nil {
+		return "", nil, "", fmt.Errorf("parsing ref for image %q: %w", image, err)
+	}
+
+	repo, err := a.NewRepository(ref.Name())
```
saschagrunert: Same issue as below: `ref.Name()` includes the tag. The old code used `parsedRef.Context().Name()` (just the repository) for `NewRepository`. Should be consistent:

```suggestion
	repo, err := a.NewRepository(ref.Context().Name())
```

internal/pkg/artifact/artifact.go
```diff
@@ -264,20 +278,29 @@ func (a *Artifact) Pull(
...
+	a.logger.Info("Resolving digest of image: " + from)
 
-		const all = ".*"
+	// Retrieve the immutable image digest before doing any verification to
+	// prevent a TOCTOU attack on the mutable tag of the base image, which
+	// it might lead to a malicious based profile being injected between
+	// verification and copying the content.
+	from, repo, sha, err := a.imageWithDigest(ctx, from, username, password)
+	if err != nil {
+		return nil, fmt.Errorf("resolving digest for image %q: %w", from, err)
```
saschagrunert: `from` gets reassigned to `""` (zero value) when `imageWithDigest` returns an error, so the error message logs `resolving digest for image ""`. Use a separate variable to preserve the original:

```suggestion
	resolvedFrom, repo, sha, err := a.imageWithDigest(ctx, from, username, password)
	if err != nil {
		return nil, fmt.Errorf("resolving digest for image %q: %w", from, err)
	}
	from = resolvedFrom
```

internal/pkg/daemon/seccompprofile/seccompprofile.go
```diff
@@ -390,16 +390,22 @@ func (r *Reconciler) resolveSyscallsForProfile(
 				return nil, fmt.Errorf("retrieving the SPOD configuration: %w", err)
 			}
 
-			disableVerification := ptr.Deref(spod.Spec.Security.DisableOCIArtifactSignatureVerification, false)
+			signOpts := &artifact.PullSignatureOptions{
+				DisableSignatureVerification: ptr.Deref(spod.Spec.Security.DisableOCIArtifactSignatureVerification, false),
+				AllowedIdentityRegexp:        spod.Spec.Security.AllowedIdentityRegexp,
+				AllowedOidcIssuerRegexp:      spod.Spec.Security.AllowedOidcIssuerRegexp,
```
saschagrunert, line 406: The CRD `+default=".*"` only applies when objects are created/defaulted by the API server. Existing SPOD resources on upgrade will have empty strings here, causing cosign verification to fail with empty `CertIdentityRegexp`/`CertOidcIssuerRegexp`. Should default to `".*"` when empty.

api/spod/v1/spod_types.go
```diff
@@ -260,6 +260,18 @@ type SPODSecurityConfig struct {
...
 	// +default=false
 	DisableOCIArtifactSignatureVerification *bool `json:"disableOciArtifactSignatureVerification,omitempty"`
+
+	// allowedIdentityRegexp regexp for allowed identity when verifying the signature of OCI
+	// image used to distribute the base profile in the cluster.
+	// +optional
+	// +default=".*"
+	AllowedIdentityRegexp string `json:"allowedIdentityRegexp,omitempty"`
+
+	// allowedOidcIssuerRegexp regexp for allowed Oidc issuer when verifying the signature of OIDC
```
saschagrunert: Typo: "OIDC image" should be "OCI image". OIDC is the auth protocol, not the image type. The `allowedIdentityRegexp` comment above correctly says "OCI".

```suggestion
	// allowedOidcIssuerRegexp regexp for allowed Oidc issuer when verifying the signature of OCI
```

internal/pkg/cli/puller/consts.go
```diff
@@ -35,4 +35,12 @@ const (
 	// FlagDisableSignatureVerification is the flag for disabling the signature
 	// verification on pull.
 	FlagDisableSignatureVerification string = "disable-signature-verification"
+
+	// FlagAllowedIdentitiesRegexp is the flag for defining the allowed identity
+	// regexp when verifying the image signature.
+	FlagAllowedIdentitiesRegexp string = "allowed-identities-regexp"
```
saschagrunert: Naming mismatch: the API field is `AllowedIdentityRegexp` (singular) but this flag uses plural "identities". Should be consistent, suggest singular to match the API.

saschagrunert (changes requested):
Nice work on the TOCTOU fix, resolving the digest before both verification and copy is the right approach. Found a few issues. See inline comments for details and suggestions.

ccojocar:
> Nice work on the TOCTOU fix, resolving the digest before both verification and copy is the right approach. Found a few issues. See inline comments for details and suggestions.

Thanks, I addressed them.

internal/pkg/cli/puller/consts.go
```diff
@@ -35,4 +35,12 @@ const (
 	// FlagDisableSignatureVerification is the flag for disabling the signature
 	// verification on pull.
 	FlagDisableSignatureVerification string = "disable-signature-verification"
+
+	// FlagAllowedIdentitiesRegexp is the flag for defining the allowed identity
+	// regexp when verifying the image signature.
+	FlagAllowedIdentitiesRegexp string = "allowed-identity-regexp"
```
saschagrunert: Naming mismatch: the constant name uses plural "Identities" but the flag value, the API field, and the comment all use singular "Identity". Should be consistent:

```suggestion
	// FlagAllowedIdentityRegexp is the flag for defining the allowed identity
	// regexp when verifying the image signature.
	FlagAllowedIdentityRegexp string = "allowed-identity-regexp"
```

Also update references in `options.go` and `cmd/spoc/main.go`.

internal/pkg/artifact/artifact.go
```diff
@@ -264,20 +278,30 @@ func (a *Artifact) Pull(
...
 	defer cancel()
 
-	if !disableSignatureVerification {
-		a.logger.Info("Verifying signature")
+	a.logger.Info("Resolving digest of image: " + from)
 
-		const all = ".*"
+	// Retrieve the immutable image digest before doing any verification to
+	// prevent a TOCTOU attack on the mutable tag of the base image, which
+	// it might lead to a malicious based profile being injected between
```
saschagrunert: Grammar: remove stray "it" and fix "based" → "base":

```suggestion
	// prevent a TOCTOU attack on the mutable tag of the base image,
	// which might lead to a malicious base profile being injected between
```

internal/pkg/artifact/artifact.go
```diff
@@ -264,20 +278,30 @@ func (a *Artifact) Pull(
...
+	// verification and copying the content.
+	originalImage := from
+	from, repo, sha, err := a.imageWithDigest(ctx, originalImage, username, password)
+	if err != nil {
+		return nil, fmt.Errorf("resolving digest for image %q: %w", originalImage, err)
+	}
+
+	if signOpts != nil && !signOpts.DisableSignatureVerification {
```
saschagrunert: A nil `signOpts` silently skips verification. The old boolean defaulted to `false` (verification on), so nil flips the safety default. Consider treating nil as "verify with defaults" to stay backward-compatible:

```suggestion
	if signOpts == nil {
		signOpts = &PullSignatureOptions{
			AllowedIdentityRegexp:   ".*",
			AllowedOidcIssuerRegexp: ".*",
		}
	}

	if !signOpts.DisableSignatureVerification {
```

internal/pkg/artifact/artifact.go
```diff
@@ -264,20 +280,37 @@ func (a *Artifact) Pull(
...
 	ctx, cancel := context.WithTimeout(c, defaultTimeout)
 	defer cancel()
 
-	if !disableSignatureVerification {
-		a.logger.Info("Verifying signature")
+	a.logger.Info("Resolving digest of image: " + from)
+
+	// Retrieve the immutable image digest before doing any verification to
+	// prevent a TOCTOU attack on the mutable tag of the base image, which
+	// which might lead to a malicious base profile being injected between
```
saschagrunert: nit:
```suggestion
	// might lead to a malicious base profile being injected between
```

I think the CRD's needs to get re-generated now.

saschagrunert (approved):

PR #3256, opened 2026-06-16
Make sure that the profiles are really bound when another profile exists in security context

ccojocar:
#### What type of PR is this?

/kind bug

Make sure that the bound profile is really in the security context and
not skip if some other profile is set already.

This can lead to a security issue when the set profile is less
permissive the than the profile being enforced. For example, an attacker
can set "type": "Unconfined" in the security context and skip the
profile being bound by the security profile operator.

#### Does this PR have test?

Yes

```release-note
NONE
```

### DIFF
--- a/internal/pkg/webhooks/binding/binding.go
+++ b/internal/pkg/webhooks/binding/binding.go
@@ -419,6 +419,7 @@ func (p *podBinder) addAppArmorContext(
 
 	if c.SecurityContext.AppArmorProfile == nil {
 		c.SecurityContext.AppArmorProfile = &aa
+
 		return true
 	}
 
@@ -427,6 +428,7 @@ func (p *podBinder) addAppArmorContext(
 	// profile is enforced through a binding.
 	if !ptr.Equal(c.SecurityContext.AppArmorProfile, &aa) {
 		c.SecurityContext.AppArmorProfile = &aa
+
 		return true
 	}
 
--- a/internal/pkg/webhooks/binding/binding.go
+++ b/internal/pkg/webhooks/binding/binding.go
@@ -30,6 +30,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -346,7 +347,6 @@ func (p *podBinder) addSecurityContext(
 func (p *podBinder) addSeccompContext(
 	c *corev1.Container, seccompProfile *seccompprofileapi.SeccompProfile,
 ) bool {
-	podChanged := false
 	profileRef := seccompProfile.Status.LocalhostProfile
 	sp := corev1.SeccompProfile{
 		Type:             corev1.SeccompProfileTypeLocalhost,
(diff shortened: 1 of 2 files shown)


internal/pkg/webhooks/binding/binding.go
```diff
@@ -380,20 +387,27 @@ func (p *podBinder) addSelinuxContext(
 		c.SecurityContext = &corev1.SecurityContext{}
 	}
 
-	if c.SecurityContext.SELinuxOptions != nil {
-		p.log.Info("cannot override existing selinux profile for pod or container")
-	} else {
+	if c.SecurityContext.SeccompProfile == nil {
```
saschagrunert: Bug: This checks `SeccompProfile` but should check `SELinuxOptions`. Copy-paste from `addSeccompContext`.

```suggestion
	if c.SecurityContext.SELinuxOptions == nil {
```

internal/pkg/webhooks/binding/binding.go
```diff
@@ -357,20 +357,27 @@ func (p *podBinder) addSeccompContext(
...
 		c.SecurityContext.SeccompProfile = &sp
-		podChanged = true
+
+		return true
 	}
 
-	return podChanged
+	// Make srue that the bound profile is really in the pod security context if already a profile
```
saschagrunert: Typo: "srue" -> "sure"

```suggestion
	// Make sure that the bound profile is really in the pod security context if already a profile
```

internal/pkg/webhooks/binding/binding.go
```diff
@@ -380,20 +387,27 @@ func (p *podBinder) addSelinuxContext(
...
-	} else {
+	if c.SecurityContext.SeccompProfile == nil {
 		c.SecurityContext.SELinuxOptions = &sl
-		podChanged = true
+
+		return true
 	}
 
-	return podChanged
+	// Make srue that the bound profile is really in the pod security context if the profile exists,
```
saschagrunert: Typo: "srue" -> "sure"

```suggestion
	// Make sure that the bound profile is really in the pod security context if the profile exists,
```

internal/pkg/webhooks/binding/binding.go
```diff
@@ -404,14 +418,22 @@ func (p *podBinder) addAppArmorContext(
...
-	} else {
+	if c.SecurityContext.AppArmorProfile == nil {
 		c.SecurityContext.AppArmorProfile = &aa
-		podChanged = true
+
+		return true
 	}
 
-	return podChanged
+	// Make srue that the bound profile is really in the pod security context, otherwise
```
saschagrunert: Typo: "srue" -> "sure"

```suggestion
	// Make sure that the bound profile is really in the pod security context, otherwise
```

internal/pkg/webhooks/binding/binding_test.go
```diff
@@ -214,6 +262,53 @@ func TestHandle(t *testing.T) {
...
+						},
+						Usage: "test-usage",
+					},
+				}, nil)
+			},
+			request: admission.Request{
+				AdmissionRequest: admissionv1.AdmissionRequest{
+					Object: runtime.RawExtension{
+						Raw: func() []byte {
+							podWithSecurityContext := testPod.DeepCopy()
```
saschagrunert, line 293: This sets `pod.Spec.SecurityContext` (pod-level `PodSecurityContext`), but `addSelinuxContext` operates on the container-level `c.SecurityContext`. The test passes by accident because the container's `SeccompProfile` is nil, taking the early-return branch. A test that sets `container.SecurityContext.SELinuxOptions` would actually exercise the overwrite path and catch the `SeccompProfile`/`SELinuxOptions` copy-paste bug above.

saschagrunert (changes requested):
The fix addresses a real security issue, but there's a copy-paste bug in `addSelinuxContext` that needs fixing before merge. Also the pod-level functions (`addPodSeccompContext`, `addPodSelinuxContext`, `addPodAppArmorContext`) still have the old "cannot override" behavior, so the same bypass applies at the pod level.

internal/pkg/webhooks/binding/binding_test.go
ccojocar: Switched them to use the container security context. Thanks for catching the issue!

ccojocar:
@saschagrunert I addressed your review comments. Thanks

saschagrunert (approved):

PR #3257, opened 2026-06-17
Skip binding when a referenced profile is not available

ccojocar:
#### What type of PR is this?

/kind bug

Skip binding when a referenced profile is not available.

The binding should be skipped when a profile is not found for a pod.
This is a measure to avoid that all Pod CRUD(s) operations in a
namespace with binding enabled will not be blocked with a 500 error.
This might also lead to a DoS when a ProfileBinding has a reference to a
non-existing profileRef.

Fix the retry logic to also propagate the last retried error in order
for the caller to understand the root cause of the retry timeout error.

#### Does this PR have test?

Yes

```release-note
Skip binding a prod to a profile when a referenced security profile is not available. This is
a measure to prevent blocking all CRUD(s) operations in a namespace with binding enabled.
```

### DIFF
--- a/internal/pkg/util/retries.go
+++ b/internal/pkg/util/retries.go
@@ -47,12 +47,14 @@ func Retry(fn func() error, retryCondition func(error) bool) error {
 
 func RetryEx(backoff *wait.Backoff, fn func() error, retryCondition func(error) bool) error {
 	var lastRetryErr error
+
 	waitErr := wait.ExponentialBackoff(*backoff, func() (bool, error) {
 		err := fn()
 		if err == nil {
 			return true, nil
 		} else if retryCondition(err) {
 			lastRetryErr = err
+
 			return false, nil
 		}
 
@@ -62,6 +64,7 @@ func RetryEx(backoff *wait.Backoff, fn func() error, retryCondition func(error) 
 		if lastRetryErr != nil {
 			return fmt.Errorf("wait on retry: %w, last retry error: %w", waitErr, lastRetryErr)
 		}
+
 		return fmt.Errorf("wait on retry: %w", waitErr)
 	}
 
--- a/internal/pkg/util/retries.go
+++ b/internal/pkg/util/retries.go
@@ -17,7 +17,6 @@ limitations under the License.
 package util
 
 import (
-	"errors"
 	"fmt"
 	"time"
 
@@ -62,7 +61,7 @@ func RetryEx(backoff *wait.Backoff, fn func() error, retryCondition func(error) 
 		return false, fmt.Errorf("retry function: %w", err)
 	})
 	if waitErr != nil {
-		if lastRetryErr != nil && errors.Is(waitErr, wait.ErrWaitTimeout) {
+		if lastRetryErr != nil && wait.Interrupted(waitErr) {
 			return fmt.Errorf("wait on retry: %w, last retry error: %w", waitErr, lastRetryErr)
 		}
 
--- a/internal/pkg/util/retries.go
+++ b/internal/pkg/util/retries.go
@@ -17,10 +17,11 @@ limitations under the License.
 package util
 
 import (
+	"errors"
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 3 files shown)


internal/pkg/util/retries.go
```diff
@@ -46,17 +46,25 @@ func Retry(fn func() error, retryCondition func(error) bool) error {
...
 		} else if retryCondition(err) {
+			lastRetryErr = err
+
 			return false, nil
 		}
 
 		return false, fmt.Errorf("retry function: %w", err)
 	})
 	if waitErr != nil {
+		if lastRetryErr != nil {
```
saschagrunert: If retry 1 gets `NotFound` (setting `lastRetryErr`) and retry 2 hits a different non-retryable error (e.g. `Forbidden`), the returned error wraps both via multiple `%w`. Callers using `kerrors.IsNotFound` would then match on the stale `NotFound` from retry 1, masking the real terminal error.

Guarding on `ErrWaitTimeout` ensures `lastRetryErr` is only included when all retries actually exhausted:

```suggestion
		if lastRetryErr != nil && errors.Is(waitErr, wait.ErrWaitTimeout) {
```

saschagrunert (commented):
Overall the approach is sound: preventing DoS by skipping bindings for missing profiles instead of returning 500. Two concerns below, one about a correctness edge case in `RetryEx` and one about the coupling between the two changes.

internal/pkg/webhooks/binding/binding.go
```diff
@@ -202,6 +202,14 @@ func (p *podBinder) updatePod(
 		}
 
 		if err != nil {
+			if kerrors.IsNotFound(err) {
```
saschagrunert, line 208: This check relies on `RetryEx` propagating `lastRetryErr` via multiple `%w` verbs (Go 1.20+). Without that change, the error after retry exhaustion is just `wait.ErrWaitTimeout` wrapped in `"wait on retry:"`, and `IsNotFound` would never match here, making this branch dead code.

The two changes are coupled but that coupling is not obvious. A future refactor of `RetryEx` could silently break this path. Worth a short comment noting the dependency.

internal/pkg/webhooks/binding/binding_test.go
```diff
@@ -790,6 +792,40 @@ func TestHandle(t *testing.T) {
...
+							b, err := json.Marshal(testPod.DeepCopy())
+							require.NoError(t, err)
+
+							return b
+						}(),
+					},
+				},
+			},
+			assert: func(resp admission.Response) {
+				require.True(t, resp.Allowed)
```
saschagrunert, line 825: nit: Consider also asserting on the response message to verify the skip reason is surfaced:

```go
require.Contains(t, resp.Result.Message, "skip binding")
```

ccojocar:
@saschagrunert I addressed the review comments. Thanks for review! 

Just curious: are you using any LLM for these reviews?

internal/pkg/webhooks/binding/binding.go
```diff
@@ -202,6 +202,17 @@ func (p *podBinder) updatePod(
...
 		if err != nil {
+			// This relies on until.Retry to propagate the last retried error though the tree of wrapped errors when a
+			// resource is not found. Without this, the last error when the retried reached the timeout would only be
+			// a wait.ErrWaitTimeout error which will never be matched by this if statement.
+			if kerrors.IsNotFound(err) {
+				p.log.Info("skip binding due to unavailable profile", "profile-kind", profileKind, "profile", namespacedName)
+				// When a profile is not found for a pod, the binding should be just skipped. Otherwise all pod CRUD(s)
+				// operation in a namespace with binding enabled will be blocked with 500 error. This might also lead
+				// to a DoS when a ProfileBinding has a non-existing profileRef.
+				return pod, admission.Allowed(fmt.Sprintf("skip binding, not found: %v %#v", profileKind, namespacedName))
```
saschagrunert: This `return` exits `updatePod` entirely, which means any remaining profile bindings in the loop are skipped. If a pod has two ProfileBindings and the first references a missing profile, the second binding is never applied.

The same loop already uses `continue` for other skip conditions (unsupported profile kind at line 201, container mismatch at line 219). This should do the same:

```suggestion
				continue
```

internal/pkg/webhooks/binding/binding.go
```diff
@@ -202,6 +202,17 @@ func (p *podBinder) updatePod(
 		}
 
 		if err != nil {
+			// This relies on until.Retry to propagate the last retried error though the tree of wrapped errors when a
```
saschagrunert: Typo: `until.Retry` should be `util.Retry`.

```suggestion
			// This relies on util.Retry to propagate the last retried error though the tree of wrapped errors when a
```

saschagrunert (commented):
The `RetryEx` change and the test look good. Two comments on the binding webhook:

saschagrunert (approved):

PR #3258, opened 2026-06-17
Improve the SELinux policy validation to check unescaped ) and use of restrictive global directives

ccojocar:
#### What type of PR is this?

/kind bug

Improve the SELinux policy validation to check unescaped ) and use of restrictive global directives.

#### Does this PR have test?

Yes

```release-note
Improve the SELinux policy validation to check unescaped ) and use of restrictive global directives.
```

### DIFF
--- a/api/selinuxprofile/v1/rawselinuxprofile_types.go
+++ b/api/selinuxprofile/v1/rawselinuxprofile_types.go
@@ -152,6 +152,7 @@ func (sp *RawSelinuxProfile) ValidatePolicy() error {
 
 	// Prevent block escape via unbalanced parentheses.
 	depth := 0
+
 	for _, r := range policy {
 		switch r {
 		case '(':
@@ -164,6 +165,7 @@ func (sp *RawSelinuxProfile) ValidatePolicy() error {
 			}
 		}
 	}
+
 	if depth != 0 {
 		return errors.New("invalid policy: unbalanced parentheses")
 	}
--- a/api/selinuxprofile/v1/rawselinuxprofile_types.go
+++ b/api/selinuxprofile/v1/rawselinuxprofile_types.go
@@ -20,9 +20,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"regexp"
-	"slices"
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -33,35 +32,42 @@ import (
 
 // restrictedDirectives contains CIL statements that alter global node state
 // or should not be allowed within a namespace-scoped container profile.
-var restrictedDirectives = []string{
-	"typepermissive",
-	"class",
-	"classorder",
-	"classmap",
-	"classmapping",
-	"sid",
-	"sidorder",
-	"sidcontext",
-	"category",
-	"categoryorder",
-	"sensitivity",
-	"sensitivityorder",
-	"dominance",
-	"level",
-	"levelrange",
-	"context",
-	"filecon",
-	"portcon",
-	"nodecon",
-	"netifcon",
-	"genfscon",
-	"policycap",
-	"mls",
+var restrictedDirectives = map[string]struct{}{
+	"block":            {},
+	"blockinherit":     {},
+	"blockstart":       {},
+	"booleanif":        {},
+	"category":         {},
+	"categoryorder":    {},
+	"class":            {},
+	"classmap":         {},
+	"classmapping":     {},
+	"classorder":       {},
+	"context":          {},
+	"dominance":        {},
+	"filecon":          {},
+	"genfscon":         {},
+	"level":            {},
+	"levelrange":       {},
+	"mls":              {},
+	"netifcon":         {},
+	"nodecon":          {},
+	"optional":         {},
+	"policycap":        {},
+	"portcon":          {},
+	"role":             {},
+	"roletype":         {},
+	"sensitivity":      {},
+	"sensitivityorder": {},
+	"sid":              {},
+	"sidcontext":       {},
+	"sidorder":         {},
+	"tunable":          {},
+	"tunableif":        {},
+	"typepermissive":   {},
+	"user":             {},
+	"userrole":         {},
 }
-
-// directiveRegex finds CIL statements by looking for an open parenthesis
-// followed by optional whitespace and an alphanumeric directive word.
-var directiveRegex = regexp.MustCompile(`\(\s*([a-zA-Z0-9_-]+)`)
 
 var (
 	// Ensure RawSelinuxProfile implements the StatusBaseUser and SecurityProfileBase interfaces.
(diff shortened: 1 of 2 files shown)


saschagrunert (approved):

api/selinuxprofile/v1/rawselinuxprofile_types.go
```diff
@@ -28,6 +31,38 @@ import (
...
+	"levelrange",
+	"context",
+	"filecon",
+	"portcon",
+	"nodecon",
+	"netifcon",
+	"genfscon",
+	"policycap",
+	"mls",
+}
```
saschagrunert: Consider adding these CIL directives to the restricted list:

- `block`, `blockinherit`, `blockabstract`: the operator wraps user policy in `(block ...)`. A user-declared `(blockinherit ...)` could inherit permissions from other blocks on the system.
- `user`, `role`, `roletype`, `userrole`: define/modify SELinux users and roles at a global level.
- `optional`, `booleanif`, `tunable`, `tunableif`: can conditionally include statements or gate policy on runtime booleans, potentially bypassing the static check.

api/selinuxprofile/v1/rawselinuxprofile_types.go
```diff
@@ -28,6 +31,38 @@ import (
...
+
+// directiveRegex finds CIL statements by looking for an open parenthesis
+// followed by optional whitespace and an alphanumeric directive word.
+var directiveRegex = regexp.MustCompile(`\(\s*([a-zA-Z0-9_-]+)`)
+
 var (
```
saschagrunert, line 72: CIL supports `;` line comments. The regex will match directives inside comments (e.g. `; (typepermissive spc_t)`) as false positive rejections. If comment stripping is out of scope, a brief code comment noting this trade-off would help future readers.

api/selinuxprofile/v1/rawselinuxprofile_types.go
```diff
@@ -115,6 +150,40 @@ func (sp *RawSelinuxProfile) ValidatePolicy() error {
...
+	}
+
+	if depth != 0 {
+		return errors.New("invalid policy: unbalanced parentheses")
+	}
+
+	// Prevent Global Privilege Escalation: Reject dangerous CIL directives.
+	// This scans for tokens immediately following an open parenthesis.
+	matches := directiveRegex.FindAllStringSubmatch(policy, -1)
+	for _, match := range matches {
```
saschagrunert: This allocates a `[][]string` for every `(\s*word` match in the policy, then does a linear `slices.Contains` against 24 entries per match. For a 500KB policy (the kubebuilder max), this adds up.

You could fold the directive check into the parentheses-balancing loop above: when you see `(`, skip whitespace, extract the next word, and look it up in a `map[string]struct{}`. One pass, no regex, no extra allocations.

saschagrunert (dismissed):
The new parentheses balancing check and directive blocklist are good security hardening. Minor note: `api/selinuxprofile/v1alpha2/rawselinuxprofile_types.go` has a duplicated `ValidatePolicy()` that isn't updated here. The webhook and controller only use v1, so this isn't a bypass risk, but it would be good to keep them in sync (or extract a shared helper) to avoid future drift.

ccojocar:
@saschagrunert I improved the policy validation check to be more performant. PTAL. Thanks

api/selinuxprofile/v1/rawselinuxprofile_types.go
```diff
@@ -115,9 +156,63 @@ func (sp *RawSelinuxProfile) ValidatePolicy() error {
...
+	// Prevent Global Privilege Escalation: Reject dangerous CIL directives.
+	// Check out all found directives in the policy and see if any of them are
+	// in the restricted list.
+	for directive := range directives {
+		if _, ok := restrictedDirectives[strings.ToLower(directive)]; ok {
+			return fmt.Errorf(
+				"invalid policy: use of restricted global directive '%s' is not allowed",
+				directive)
+		}
+	}
```
saschagrunert, line 221: Five problems in this block:

1. **False positives (test fails)**: All words are collected into `directives`, not just the first token after `(`. In CIL the directive is the first s-expression element; everything else is arguments. `(allow typepermissive file (read))` is incorrectly flagged because `typepermissive` appears as *any* word.
2. **Bypass via `)` boundary**: The `directive` buffer is never flushed on `)`. `(typepermissive)` passes validation because `typepermissive` is silently dropped.
3. **Bypass via EOF**: After the loop the remaining `directive` buffer is discarded, so a trailing word is never checked.
4. **Nondeterministic error message (test fails)**: Map iteration order makes the reported directive name unpredictable when multiple restricted words exist (the `mls`/`sensitivity` test).
5. **Case mismatch in error message (test fails)**: The error includes the original-case map key (`TYPEPERMISSIVE`) but the test expects lowercase (`typepermissive`).

Fix: track whether we're at the directive position (first token after `(`), check inline instead of collecting into a map, flush on `)` and EOF, and use the lowercased form in the error message.

```suggestion
	// Prevent block escape via unbalanced parentheses.
	// Also reject restricted CIL directives (the first token after each '(').
	depth := 0
	directive := []rune{}
	atDirectivePosition := false

	checkDirective := func() error {
		if len(directive) == 0 || !atDirectivePosition {
			return nil
		}
		d := strings.ToLower(string(directive))
		if _, ok := restrictedDirectives[d]; ok {
			return fmt.Errorf(
				"invalid policy: use of restricted global directive '%s' is not allowed", d)
		}
		return nil
	}

	for _, r := range policy {
		switch r {
		case '(':
			if err := checkDirective(); err != nil {
				return err
			}
			depth++
			directive = []rune{}
			atDirectivePosition = true
		case ')':
			if err := checkDirective(); err != nil {
				return err
			}
			depth--
			if depth < 0 {
				return errors.New(
					"invalid policy: unmatched closing parenthesis ')' allows block escape")
			}
			directive = []rune{}
			atDirectivePosition = false
		default:
			if isDirectiveCharacter(r) {
				directive = append(directive, r)
			} else if len(directive) > 0 {
				if err := checkDirective(); err != nil {
					return err
				}
				directive = []rune{}
				atDirectivePosition = false
			}
		}
	}

	if err := checkDirective(); err != nil {
		return err
	}

	if depth != 0 {
		return errors.New("invalid policy: unbalanced parentheses")
	}
```

api/selinuxprofile/v1/rawselinuxprofile_types.go
```diff
@@ -115,9 +156,63 @@ func (sp *RawSelinuxProfile) ValidatePolicy() error {
...
+
 	return nil
 }
 
+func isDirectiveCharacter(r rune) bool {
```
saschagrunert, line 226: nit: This can be simplified.

```suggestion
func isDirectiveCharacter(r rune) bool {
	return unicode.IsLetter(r) || unicode.IsDigit(r) || r == '_' || r == '-'
}
```

saschagrunert (changes requested):
Three tests fail and there are two bypass vectors in the directive check. The parenthesis balancing is correct, but the directive-checking logic needs structural changes.

See inline suggestions for fixes.

ccojocar:
@saschagrunert review comments, are addressed.

saschagrunert (approved):

PR #3259, opened 2026-06-17
fix: make selinuxd policy compatible with refpolicy

mirza-src:
#### What type of PR is this?

/kind bug

The `selinuxd.cil` policy shipped by SPO is incompatible with [refpolicy](https://github.com/SELinuxProject/refpolicy)-based distributions (such as Flatcar Linux):

1. **`selinux_login_config_t` is not defined**: refpolicy does not define the `selinux_login_config_t` type.

2. **`load_policy` neverallow violation**: refpolicy enforces [`neverallow ~{ selinux_unconfined_type can_load_policy } security_t:security load_policy`](https://github.com/SELinuxProject/refpolicy/blob/30d3cf5abd1872d3da5dd44de37de4251674f736/policy/modules/kernel/selinux.te#L71). The current policy grants `load_policy` directly via an `allow` rule, violating this rule.

This PR fixes both issues:
- Wraps the `selinux_login_config_t` rules in an `(optional rhel_only ...)` block so they are silently skipped on systems where that type is not defined.
- Replaces the direct `allow ... load_policy` with `(typeattributeset can_load_policy process)`

None

#### Does this PR have test?

N/A

The `can_load_policy` attribute exist in both  [refpolicy](https://github.com/SELinuxProject/refpolicy) and [fedora-selinux/selinux-policy](https://github.com/fedora-selinux/selinux-policy).

```release-note
Fixed selinuxd CIL policy compatibility with refpolicy-based distributions (e.g. Flatcar Linux)
```

### DIFF
--- a/deploy/base/profiles/selinuxd.cil
+++ b/deploy/base/profiles/selinuxd.cil
@@ -15,7 +15,6 @@
     (allow process file_context_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
     (allow process file_context_t ( sock_file ( append getattr open read write )))
     (allow process security_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
-    (typeattributeset can_load_policy process)
     (allow process selinux_config_t ( dir ( add_name create getattr ioctl lock open read remove_name rename rmdir search setattr write )))
     (allow process selinux_config_t ( fifo_file ( getattr read write append ioctl lock open )))
     (allow process selinux_config_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
@@ -37,7 +36,17 @@
     (allow process sysfs_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
     (allow process sysfs_t ( sock_file ( append getattr open read write )))
 
-    (optional rhel_only
+    ; Best effort to allow loading policies
+    (optional load_policy_rule_support
+        ; Targets distros that allow direct rules but might lack the 'can_load_policy' typeattribute (fails safely if blocked by a neverallow)
+        (allow process security_t ( security ( load_policy )))
+    )
+    (optional can_load_policy_attr_support
+        ; Targets distros where 'can_load_policy' typeattribute exists
+        (typeattributeset can_load_policy process)
+    )
+
+    (optional login_config_support
         (allow process selinux_login_config_t ( dir ( add_name create getattr ioctl lock open read remove_name rmdir search setattr write )))
         (allow process selinux_login_config_t ( fifo_file ( getattr read write append ioctl lock open )))
         (allow process selinux_login_config_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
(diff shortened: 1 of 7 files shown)


deploy/base/profiles/selinuxd.cil
```diff
@@ -15,15 +15,11 @@
     (allow process file_context_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
     (allow process file_context_t ( sock_file ( append getattr open read write )))
     (allow process security_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
-    (allow process security_t ( security ( load_policy )))
+    (typeattributeset can_load_policy process)
```
saschagrunert:
```suggestion
    (optional can_load_policy_support
        (typeattributeset can_load_policy process)
    )
```

If `can_load_policy` is missing on some niche distro, this hard-fails policy loading. The `selinux_login_config_t` rules already got the defensive treatment; this should too for consistency. If the attribute is absent, the process simply won't be able to load policy, which is a safer failure mode than refusing to load the entire module.

deploy/base/profiles/selinuxd.cil
```diff
@@ -40,4 +36,11 @@
     (allow process sysfs_t ( fifo_file ( getattr read write append ioctl lock open )))
     (allow process sysfs_t ( file ( append create getattr ioctl lock map open read rename setattr unlink write )))
     (allow process sysfs_t ( sock_file ( append getattr open read write )))
+
+    (optional rhel_only
```
saschagrunert:
```suggestion
    (optional login_config_support
```

`selinux_login_config_t` exists on Fedora, CentOS Stream, and other fedora-selinux derivatives, not just RHEL. The name shows up in `semodule -l` and policy debug output, so accuracy helps anyone triaging policy issues on non-RHEL systems.

deploy/base/profiles/selinuxd.cil
mirza-src: @saschagrunert should I also add the old `(allow process security_t ( security ( load_policy )))` rule in a separate `optional` block for possibly better compatibility?

deploy/base/profiles/selinuxd.cil
```diff
@@ -40,4 +35,21 @@
...
+
+    ; Best effort to allow loading policies
+    (optional load_policy_rule_support
+        ; Targets distros that allow direct rules but might lack the 'can_load_policy' typeattribute (fails safely if blocked by a neverallow)
+        (allow process security_t ( security ( load_policy )))
+    )
+    (optional can_load_policy_attr_support
+        ; Targets distros where 'can_load_policy' typeattribute exists
+        (typeattributeset can_load_policy process)
+    )
```
saschagrunert, line 47: The comment on the first optional is misleading: CIL `optional` blocks handle missing types/attributes, not neverallow violations. If all types resolve but a neverallow is violated, the entire module load fails regardless of the `optional` wrapper.

The actual safety mechanism here is the second optional: by adding `process` to `can_load_policy`, the neverallow (which exempts `can_load_policy` members) is satisfied before it's ever checked. On any system with refpolicy's neverallow, `can_load_policy` also exists, so the second optional always succeeds and the allow rule is legal.

```suggestion
    ; Best effort to allow loading policies
    (optional load_policy_rule_support
        ; Direct allow rule for distros without 'can_load_policy' (e.g. minimal SELinux setups)
        (allow process security_t ( security ( load_policy )))
    )
    (optional can_load_policy_attr_support
        ; Join the can_load_policy attribute, which also satisfies refpolicy's neverallow
        (typeattributeset can_load_policy process)
    )
```

saschagrunert (approved):

PR #3261, opened 2026-06-18
Add SELinux semantic validation for dangerous types, classes and permissions

ccojocar:
#### What type of PR is this?

/kind bug

* Add semantic validation for dangerous SELinux types, classes and permissions.
* Extend the SELinux options to allow defining the denied types, classes and permissions.

#### Does this PR have test?

Yes

```release-note
Add semantic validation for dangerous SELinux types, classes and permissions, including extending
the spod configuration to allow setting additional denied types, classes and permissions. 
```


internal/pkg/daemon/selinuxprofile/selinuxprofile.go
```diff
@@ -219,11 +219,22 @@ func (sph *selinuxProfileHandler) handleInheritSystemPolicy(
 }
 
 func (sph *selinuxProfileHandler) GetCILPolicy() (string, error) {
+	spod, err := common.GetSPOD(context.Background(), sph.cli)
```
saschagrunert: `GetSPOD()` is already called during `handleInheritSystemPolicy()` (line 201), which runs during the `Validate()` phase before `GetCILPolicy()` is reached. This adds a redundant API server round-trip on every reconciliation.

Consider fetching the denied options once during `Validate()` and storing them on the `selinuxProfileHandler` struct (similar to how `systemInherits` is populated and cached):

```go
type selinuxProfileHandler struct {
    sp                *selinuxprofileapi.SelinuxProfile
    cli               client.Client
    systemInherits    []string
    objInherits       []selinuxprofileapi.SelinuxProfileObject
    deniedOpts        *translator.Options  // populated during Validate
    ...
}
```

This would also avoid the `context.Background()` problem here, since `handleInheritSystemPolicy` already has access to the reconciliation context.

internal/pkg/translator/obj2cil.go
```diff
@@ -31,11 +32,90 @@ const (
...
+		"security_t": {}, // SELinux fs
+		"shadow_t":   {}, // /etc/shadow
+	}
+
+	// deniedClasses prevents manipulation of the SELinux system itself.
+	deniedClasses = map[string]struct{}{
+		"capability": {}, // Linux capabilities
+		"security":   {}, // Allows managing SELinux state
+		"system":     {}, // Core system operations
+	}
```
saschagrunert, line 51: `capability2` is missing. SELinux splits capabilities into `capability` (CAP_CHOWN through CAP_SYS_TTY_CONFIG) and `capability2` (CAP_MAC_OVERRIDE, CAP_MAC_ADMIN, CAP_SYSLOG, etc.). Without denying `capability2`, a profile can request those capabilities through the second class.

```suggestion
	deniedClasses = map[string]struct{}{
		"capability":  {}, // Linux capabilities
		"capability2": {}, // Extended Linux capabilities (mac_override, syslog, etc.)
		"security":    {}, // Allows managing SELinux state
		"system":      {}, // Core system operations
	}
```

saschagrunert (commented):
Denylist gaps and architecture concerns noted inline. The release note also has a typo: "denited" should be "denied".

internal/pkg/translator/obj2cil.go
```diff
@@ -31,11 +32,90 @@ const (
...
+
+	// deniedPermissions prevents hostile actions on otherwise standard classes.
+	deniedPermissions = map[string]struct{}{
+		"load_policy": {}, // Can rewrite SELinux rules
+		"mac_admin":   {},
+		"mounton":     {},
+		"relabelto":   {}, // Can hijack other containers' files
```
saschagrunert, line 58: `relabelfrom` should also be denied. If `relabelto` is blocked because it can hijack other containers' files, `relabelfrom` on a sensitive type is equally dangerous (allows reading files that should be inaccessible).

```suggestion
		"load_policy":  {}, // Can rewrite SELinux rules
		"mac_admin":    {},
		"mounton":      {},
		"relabelfrom":  {}, // Can read files from other security contexts
		"relabelto":    {}, // Can hijack other containers' files
		"setbool":      {},
		"setenforce":   {}, // Can disable SELinux
		"setsecparam":  {},
```

internal/pkg/translator/obj2cil_test.go
```diff
@@ -334,12 +336,127 @@ func TestObject2CIL(t *testing.T) {
...
+			options: &Options{
+				DeniedTypes:       []string{"var_log_t"},
+				DeniedClasses:     []string{"dir"},
+				DeniedPermissions: []string{"open"},
+			},
+			wantErr: true,
+			inheritsys: []string{
+				"container",
+			},
+		},
```
saschagrunert, line 529: This test case sets all three denied options (`DeniedTypes`, `DeniedClasses`, `DeniedPermissions`) but validation short-circuits on the first match (type `var_log_t`). The `DeniedClasses` and `DeniedPermissions` paths through the options are never actually exercised.

Consider splitting into three test cases, each with only one of the three denied option fields set, so all three code paths are covered independently.

ccojocar:
@saschagrunert Thanks for review! I addressed all your comments.

internal/pkg/daemon/selinuxprofile/selinuxprofile.go
```diff
@@ -195,14 +196,20 @@ func (sph *selinuxProfileHandler) handleInheritSPOPolicy(
...
 	spod, err := common.GetSPOD(context.Background(), sph.cli)
 	if err != nil {
 		return fmt.Errorf("couldn't get spod to verify system inheritance: %w", err)
 	}
 
+	sph.deniedOpts = &translator.Options{
+		DeniedTypes:       spod.Spec.Selinux.Options.DeniedTypes,
+		DeniedClasses:     spod.Spec.Selinux.Options.DeniedClasses,
+		DeniedPermissions: spod.Spec.Selinux.Options.DeniedPermissions,
+	}
```
saschagrunert, line 215: This is only called via `handleSelinuxOptions`, which runs for `SystemPolicyKind`/empty kind. If a profile inherits exclusively from other `SelinuxProfile` objects, `handleSelinuxOptions` is never called, `sph.deniedOpts` stays `nil`, and `Object2CIL` only enforces the hardcoded defaults.

That means any admin-configured denied types/classes/permissions from SPOD are silently bypassed.

The previous approach in the earlier commit (fetching SPOD in `GetCILPolicy()`) didn't have this problem since it ran for every profile. Consider loading the SPOD denied options there, or at the start of `Validate()`, so they apply regardless of inheritance kind.

internal/pkg/daemon/selinuxprofile/selinuxprofile.go
```diff
@@ -223,7 +230,7 @@ func (sph *selinuxProfileHandler) GetCILPolicy() (string, error) {
 	// have been initialized already
 	// At this point, validation has happened and no errors will happen when
 	// rendering
```
saschagrunert: This comment is now outdated since `Object2CIL` can return semantic validation errors. Worth removing or updating.

internal/pkg/translator/obj2cil.go
```diff
@@ -31,11 +32,92 @@ const (
...
+
+type deniedOptions struct {
+	deniedTypes       map[string]struct{}
+	deniedClasses     map[string]struct{}
+	deniedPermissions map[string]struct{}
+}
+
+func deniedOptionsFromOpts(opts *Options) *deniedOptions {
+	// Global maps are cloned to avoid data race in this function is
+	// executed in parallel (e.g. in tests).
```
saschagrunert, line 89: Small grammar nit: "in this function is" should be "if this function is".

```suggestion
	// Clone global maps to avoid data races when executed in parallel.
```

internal/pkg/translator/obj2cil_test.go
```diff
@@ -31,10 +31,12 @@ func TestObject2CIL(t *testing.T) {
 	tests := []struct {
 		name        string
 		profile     *selinuxprofileapi.SelinuxProfile
+		options     *Options
```
saschagrunert, line 34: Nice test coverage for the error paths. One thing that's missing: a positive test with `Options` set where the profile doesn't hit any denied entries. The existing passing tests all use `nil` options, so the success path through `deniedOptionsFromOpts` with non-nil options is not covered.

ccojocar:
@saschagrunert review comments addressed. Thanks!

saschagrunert (approved):
Most of the previous review feedback has been addressed. The core logic is sound, the hardcoded denylists are well chosen, and the SPOD extension mechanism is clean.

A few minor items remain (duplicate comment in `deniedOptionsFromOpts`, missing positive test with non-nil Options, error message redundancy), but nothing blocking.

PR #3264, opened 2026-06-19
Add safety bound check for path length in ebpf recorder

ccojocar:
#### What type of PR is this?

/kind bug

Adds a safety bounds check for path length in the ebpf recorder.

```release-note
NONE
```


saschagrunert (approved):
LGTM, clean defensive hardening. The early return before any state mutation (`*buflen`, `*buffer`) is the right spot for this check.

PR #3265, opened 2026-06-19
Add a new flag to enable insecure metrics access on daemon container

ccojocar:
#### What type of PR is this?

/kind feature

Adds a new flag `--with-insecure-metrics-access` on daemon command and a new configuration flag
in spod `enableInsecureMetricsAccess` to enable insecure metrics access. This flag will disable
TLS and authentication on metrics endpoint.

```release-note
Add a new flag --with-insecure-metrics-access on daemon command a new configuration flag to
spod which disables TLS and authentication.
```

### DIFF
--- a/api/spod/v1/spod_types.go
+++ b/api/spod/v1/spod_types.go
@@ -118,6 +118,11 @@ type SPODSpec struct {
 	// +optional
 	// +default=false
 	EnableMemoryOptimization *bool `json:"enableMemoryOptimization,omitempty"`
+	// enableInsecureMetricsAccess enables unauthenticated access to the metrics
+	// endpoint. This will disable TLS and authentication for the metrics endpoint.
+	// +optional
+	// +default=false
+	EnableInsecureMetricsAccess *bool `json:"enableInsecureMetricsAccess,omitempty"`
 	// enableAppArmor tells the operator whether or not to enable AppArmor
 	// support for this SPOD instance.
 	// +optional
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -227,7 +227,7 @@ func main() {
 				},
 				&cli.BoolFlag{
 					Name:    insecureMetricsAccessFlag,
-					Usage:   "Allow unauthenticated access to metrics endpoint (default: false)",
+					Usage:   "Allow unauthenticated access to metrics endpoint",
 					Value:   false,
 					EnvVars: []string{config.EnableInsecureMetricsAccessEnvKey},
 				},
@@ -722,7 +722,7 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
 	}
 
 	if ctx.Bool(insecureMetricsAccessFlag) {
-		setupLog.Info("WARNING: insecure metrics access enabled, TLS and authentication are disabled")
+		setupLog.Info("Insecure metrics access enabled, TLS and authentication are disabled")
 
 		ctrlOpts.Metrics.SecureServing = false
 		ctrlOpts.Metrics.CertDir = ""
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -102,6 +102,7 @@ const (
 	rawSelinuxFlag               string = "with-raw-selinux"
 	webhookFlag                  string = "webhook"
 	memOptimFlag                 string = "with-mem-optim"
+	insecureMetricsAccessFlag    string = "with-insecure-metrics-access"
 	defaultWebhookPort           int    = 9443
 	auditLogIntervalSecondsParam string = "audit-log-interval-seconds"
 	auditLogPathParam            string = "audit-log-path"
@@ -223,6 +224,12 @@ func main() {
 					Name:  memOptimFlag,
 					Usage: "Enable memory optimization by watching only labeled pods",
 					Value: false,
+				},
+				&cli.BoolFlag{
+					Name:    insecureMetricsAccessFlag,
+					Usage:   "Allow unauthenticated access to metrics endpoint (default: false)",
+					Value:   false,
+					EnvVars: []string{config.EnableInsecureMetricsAccessEnvKey},
 				},
 			},
 		},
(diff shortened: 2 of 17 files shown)


internal/pkg/manager/spod/spod_controller.go
```diff
@@ -753,6 +753,13 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 			"--with-mem-optim=true")
 	}
 
+	// Enable insecure metrics access if requested
+	if isInsecureMetricsEnabled(cfg) {
+		templateSpec.Containers[bindata.ContainerIDDaemon].Args = append(
+			templateSpec.Containers[bindata.ContainerIDDaemon].Args,
+			"--with-insecure-metrics=true")
```
saschagrunert: The flag name doesn't match the CLI definition. The daemon CLI expects `--with-insecure-metrics-access` (defined in `main.go:105`) but the controller passes `--with-insecure-metrics`. The daemon will ignore this unknown flag, so setting `enableInsecureMetricsAccess` in the SPOD CR will have no effect.

```suggestion
			"--with-insecure-metrics-access=true")
```

cmd/security-profiles-operator/main.go
```diff
@@ -714,6 +721,15 @@ func runDaemon(ctx *cli.Context, info *version.Info) error {
 		},
 	}
 
+	if ctx.Bool(insecureMetricsAccessFlag) {
+		setupLog.Info("WARNING: insecure metrics access enabled, TLS and authentication are disabled")
```
saschagrunert: Embedding `WARNING:` in an info-level log message is misleading. Either log at warning level or drop the prefix.

```suggestion
		setupLog.Info("Insecure metrics access enabled, TLS and authentication are disabled")
```

cmd/security-profiles-operator/main.go
```diff
@@ -224,6 +225,12 @@ func main() {
 					Usage: "Enable memory optimization by watching only labeled pods",
 					Value: false,
 				},
+				&cli.BoolFlag{
+					Name:    insecureMetricsAccessFlag,
+					Usage:   "Allow unauthenticated access to metrics endpoint (default: false)",
```
saschagrunert: The `(default: false)` suffix is redundant. The `cli` framework already renders the default value from the `Value` field in `--help` output.

```suggestion
					Usage:   "Allow unauthenticated access to metrics endpoint",
```

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -753,6 +753,13 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 			"--with-mem-optim=true")
 	}
 
+	// Enable insecure metrics access if requested
```
saschagrunert: nit: this comment restates what the code does. The function name `isInsecureMetricsEnabled` already conveys the intent, and other similar blocks in this function (e.g., the `--with-mem-optim` block above) don't have comments either.

```suggestion
```

saschagrunert (commented):
Missing initialization in `internal/pkg/manager/spod/bindata/spod.go`: the `DefaultSPOD` struct initializes every other `*bool` field with `new(bool)` (`EnableProfiling`, `EnableMemoryOptimization`, `EnableAppArmor`, etc.) but `EnableInsecureMetricsAccess` is missing. Should be added for consistency.

ccojocar:
@saschagrunert review comments addressed. Thanks for review!

api/spod/v1/spod_types.go
```diff
@@ -118,6 +118,11 @@ type SPODSpec struct {
 	// +optional
 	// +default=false
 	EnableMemoryOptimization *bool `json:"enableMemoryOptimization,omitempty"`
+	// enableInsecureMetricsAccess enables unauthenticated access to the metrics
+	// endpoint. This will disable TLS and authentication for the metrics endpoint.
+	// +optional
+	// +default=false
+	EnableInsecureMetricsAccess *bool `json:"enableInsecureMetricsAccess,omitempty"`
```
saschagrunert, line 125: The field is added to the v1 `SPODSpec` but not to `api/spod/v1alpha1/spod_types.go`. The conversion functions in `api/spod/v1alpha1/conversion.go` also don't include `EnableInsecureMetricsAccess`. Every other boolean field (`EnableProfiling`, `EnableMemoryOptimization`, `EnableAppArmor`) exists in both API versions with corresponding conversion lines in `ConvertTo` (line 66-68) and `ConvertFrom` (line 137-139).

Without this, a v1 -> v1alpha1 -> v1 roundtrip silently drops the setting, and v1alpha1 users can't set the flag at all.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -753,6 +753,12 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 			"--with-mem-optim=true")
 	}
 
+	if isInsecureMetricsEnabled(cfg) {
```
saschagrunert, line 757: When insecure metrics is enabled, the `ServiceMonitor` (created in `bindata/servicemonitor.go`) is still hardcoded to `scheme: "https"` with TLS config and bearer token authorization. Prometheus would attempt an HTTPS connection to a plain HTTP endpoint and fail.

The `ServiceMonitor` and the metrics `Service` should be adjusted when insecure metrics is enabled, for example by passing the insecure flag through to `ServiceMonitor()` / `GetMetricsService()` and conditionally building the endpoint without TLS/auth.

saschagrunert (changes requested):
The feature makes sense but a few things need to be addressed before this can be merged.

### Missing test coverage

This is a security-sensitive feature (disabling TLS and authentication on the metrics endpoint). There should be at least unit tests for `isInsecureMetricsEnabled()` and for the controller logic that appends `--with-insecure-metrics-access=true` to the daemon args.

### Release note typo

"configuraiton" should be "configuration" in the release note.

ccojocar:
@saschagrunert the review comments are addressed.

saschagrunert (approved):

PR #3271, opened 2026-06-22
Use security-profiles-merger for seccomp syscall union

saschagrunert:
/kind feature
/kind cleanup

#### What type of PR is this?

/kind cleanup

Replaces `dario.cat/mergo` with the [security-profiles-merger](https://github.com/saschagrunert/security-profiles-merger) library (v0.2.1) for both seccomp and AppArmor profile union merging.

For seccomp: provides per-name syscall normalization (one entry per syscall name), action-aware merging (less restrictive action wins for union), and proper argument deduplication.

For AppArmor: adds glob-aware path matching for filesystem rules, proper read/write permission promotion, and sorted output for deterministic comparison.

Adds a type conversion layer between SPO's CRD types and OCI/merger types. The `UnionSyscalls` function signature is unchanged. Profiles are normalized before `--check` mode comparison to avoid false positives from representation differences.

None

#### Does this PR have test?

Yes, updated existing tests and added conversion round-trip tests, normalization tests, and AppArmor merge tests.

The merger normalizes syscalls per-name, so test expectations changed from multi-name grouped entries to single-name entries sorted alphabetically. Tests previously used fake actions ("foo", "bar") or empty actions, which are now replaced with real seccomp actions since the merger uses action severity ordering.

```release-note
NONE
```


saschagrunert:

ccojocar (approved):

PR #3276, opened 2026-06-24
fix: add default selinux options to v1 api so allowedSystemProfiles is applied

mirza-src:
#### What type of PR is this?

/kind bug

#3243 was merged before the v1 API and it looks like it was missed in the v1 API.

Fix: add `+default={}` on `Options` so the API server creates the parent and applies the child default.

None

#### Does this PR have test?

N/A

None

```release-note
Fixed `allowedSystemProfiles` default (`["container"]`) not being applied in v1 API.
```


saschagrunert (approved):
Thanks!

PR #3287, opened 2026-06-30
feat: add podSelector to ProfileBinding

### LINKED ISSUE #3280: ProfileBinding only matches by image, cannot scope a profile to a specific workload

Allow `ProfileBinding` to select target pods by something more specific than the container image alone, e.g. a `podSelector` (label selector) match, so that a binding can be scoped to a particular workload instead of every pod in the namespace that happens to run a given image.

The binding webhook lists all `ProfileBinding`s in the pod's namespace and matches purely on the container image string, there is no pod selector filter. So a binding is applied to **every** container in the namespace that uses the matching image, regardless of which workload that container belongs to.

A proposed shape:

```yaml
apiVersion: security-profiles-operator.x-k8s.io/v1
kind: ProfileBinding
metadata:
  name: ebs-csi-node
  namespace: kube-system
spec:
  profileRef:
    kind: SelinuxProfile
    name: ebs-csi-node
  image: "*" # still required
  podSelector: # NEW: scope the binding to a specific workload
    matchLabels:
      app: ebs-csi-node
```

#### Why is this needed:

The image is not a reliable identifier for "which SELinux/seccomp/AppArmor policy a container needs," because the same image can require different policies depending on the workload it runs in.

Concrete case while recording an SELinux policy for the [aws-ebs-csi-driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver):

- The `ebs-csi-node` DaemonSet's pod has containers that run as `container_t` and need to `connectto` a socket in a shared volume that is served by a privileged container running as `spc_t`. The default `container_t` policy didn't allow `connectto` to an `spc_t` socket, so I recorded a custom `SelinuxProfile` for the node pod. The recorded policy worked correctly.
- To enforce it, I created a `ProfileBinding`. The only available filter is  `image`, so I bound on the container image.
- The problem: the `livenessprobe` image used in the node DaemonSet is **also** used by another component (the controller). In the controller, everything runs unprivileged as `container_t` and the probe connects to a `container_t` socket, it worked fine with the default type.
- Because `ProfileBinding` matches by image only, the binding intended for the node DaemonSet was **also applied to the controller's `livenessprobe` container**. That forced the controller's probe into the node-specific domain (which only permits `connectto` to `spc_t`), so it could no longer connect to the `container_t` socket it actually needs. The controller broke.

The root cause is the assumption baked into `ProfileBinding` that "same image ⇒ same required policy everywhere." That breaks down whenever the policy a container needs depends on the rest of the pod it runs in (here: which SELinux domain the peer socket runs as).

A label/pod selector on `ProfileBinding` would let the binding be scoped to the exact workload the profile was recorded for, the same way `ProfileRecording` is already scoped via `podSelector`.

#### User story covered

Operators should be able to bind a profile to a specific workload (by pod label selector), so that workloads which happen to share a container image but require different policies are not forced into the wrong profile.

mirza-src:
#### What type of PR is this?

/kind feature
/kind api-change

`ProfileBinding` currently matches containers solely by image name. Workloads that share the same image but require different security policies cannot be differentiated.

This PR adds an optional `podSelector` field to `ProfileBindingSpec`. When set, the binding is only applied to pods whose labels match the selector (in addition to the image match).

Fixes #3280

#### Does this PR have test?

Yes. Added two cases to in `internal/pkg/webhooks/binding/binding_test.go`:
- Pod is patched when its labels match the binding's `podSelector`
- Pod is left unchanged when its labels do not match

- Bindings without a `podSelector` are unaffected
- An invalid selector logs an error and skips only that binding
- Updated `installation-usage.md` with an example

```release-note
Added an optional `podSelector` field to `ProfileBinding`.
```


saschagrunert (approved):
LGTM. Implementation is correct and backward compatible. Minor test coverage gaps (invalid podSelector, `image: "*"` + podSelector combo) could be addressed in a follow-up.

saschagrunert:
Thanks!

PR #3288, opened 2026-06-30
feat: add allowlist in SELinux options to override default denylist

### LINKED ISSUE #3281: Allow overriding the hardcoded SELinux denylist

A way to **override / allow** specific entries from the hardcoded SELinux denylist introduced in #3261, so that a legitimate workload that genuinely requires an otherwise-denied type, class, or permission can have a profile recorded and translated.

Right now the denylist is hardcoded. The SPOD config exposes `deniedTypes` / `deniedClasses` / `deniedPermissions`, but it only ever **adds** to the hardcoded maps, with no path to remove them.

Possible shapes:

- A per-`SelinuxProfile` allowlist, also expressible on `ProfileRecording` so the recorded profile carries the override through. This scopes the override to the workload that needs it.
- A SPOD-level allowlist. Simplest, but global, it relaxes the denylist for every translated policy.

#### Why is this needed:

The hardcoded denylist is a good safe default, but it assumes no legitimate workload ever needs those types/permissions. That assumption doesn't hold for some real, low-level monitoring workloads.

Concrete case: recording an `SelinuxProfile` for [node-exporter](https://github.com/prometheus/node_exporter). node-exporter reads host/kernel state and legitimately needs access that the denylist blocks (e.g. `kernel_t`). Because the denied entries cannot be overridden, the recorded profile fails translation.

An operator who understands the risk should be able to explicitly allow a specific type/class/permission for a specific workload, the same way `allowedSystemProfiles` already lets operators opt into otherwise-broad system profiles "with care".

#### User story covered

Operators should be able to explicitly allow a specific SELinux type, class, or permission from the built-in denylist for a workload that genuinely requires it (e.g. node-exporter needing `kernel_t`).

mirza-src:
#### What type of PR is this?

/kind feature
/kind api-change

This adds SPOD-level fields in `spec.selinuxOptions` `allowedTypes`, `allowedClasses`, and `allowedPermissions` that allows the built-in denylist to be overriden. 

Fixes #3281

#### Does this PR have test?

Yes, unit tests covering an allowing built-in denied type (`kernel_t`), class (`capability`), and permission (`mounton`).

None

```release-note
Added `allowedTypes`, `allowedClasses`, and `allowedPermissions` fields to the SPOD `selinuxOptions`, allowing operators to remove specific entries from the built-in SELinux denylist.
```


internal/pkg/translator/obj2cil.go
```diff
@@ -110,6 +122,18 @@ func deniedOptionsFromOpts(opts *Options) *deniedOptions {
 		denied.deniedPermissions[p] = struct{}{}
 	}
 
+	for _, t := range opts.AllowedTypes {
```
saschagrunert, line 114: These deletions run after the user-specified `DeniedTypes`/`DeniedClasses`/`DeniedPermissions` additions above (lines 113-123), so allowed entries override both the built-in denylist *and* user-specified denied entries. If someone sets `DeniedTypes: ["foo_t"]` and `AllowedTypes: ["foo_t"]` on the same SPOD, the allowed silently wins.

Is that intentional? If the goal is to only override the built-in defaults, these loops should run before the user-denied additions (or only delete keys that exist in the hardcoded `deniedTypes`/`deniedClasses`/`deniedPermissions` maps). If it is intentional, worth a comment here noting the precedence.

api/spod/v1/spod_types.go
```diff
@@ -53,6 +53,27 @@ type SelinuxOptions struct {
 	// +optional
 	// +listType=set
 	DeniedPermissions []string `json:"deniedPermissions,omitempty"`
+
+	// allowedTypes if specified, a list of SELinux types which are removed
```
saschagrunert, line 57: The doc says "removed from the built-in denylist" but the implementation also removes entries added via `deniedTypes`. Consider clarifying: "removed from the effective denylist (built-in and user-specified)." Or, if only the built-in defaults should be overridable, the implementation in `deniedOptionsFromOpts` needs adjusting.

internal/pkg/translator/obj2cil_test.go
```diff
@@ -527,6 +527,102 @@ func TestObject2CIL(t *testing.T) {
 				"container",
 			},
 		},
+		{
```
saschagrunert, line 530: These three test cases cover the happy path for each dimension individually. Could you add a test for the interaction when an entry appears in both allowed and denied (e.g., `DeniedTypes: ["var_log_t"], AllowedTypes: ["var_log_t"]`)? That documents the intended precedence behavior.

internal/pkg/daemon/selinuxprofile/selinuxprofile.go
```diff
@@ -209,9 +209,12 @@ func (sph *selinuxProfileHandler) handleSelinuxOptions(
 	spod *spodapi.SecurityProfilesOperatorDaemon,
 ) {
 	sph.deniedOpts = &translator.Options{
```
saschagrunert: nit: `deniedOpts` now carries both denied and allowed options. Consider renaming to `selinuxOpts` or `translatorOpts` to match the broader scope.

api/spod/v1/spod_types.go
mirza-src: updated `deniedOptionsFromOpts` with higher precedence for user-specified denylist

internal/pkg/translator/obj2cil.go
mirza-src: updated with higher precedence for user-specified denylist

internal/pkg/daemon/selinuxprofile/selinuxprofile.go
mirza-src: renamed to `translatorOpts`

internal/pkg/translator/obj2cil_test.go
```diff
@@ -527,6 +527,134 @@ func TestObject2CIL(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+			// A user-specified deny takes precedence over an allow for the same
+			// entry, so translation is rejected.
+			options: &Options{
+				DeniedTypes:  []string{"var_log_t"},
+				AllowedTypes: []string{"var_log_t"},
```
saschagrunert: This test uses `var_log_t`, which is not in the built-in denylist. That means the `AllowedTypes` delete on line 115 is a no-op (deleting a key that doesn't exist), and only the `DeniedTypes` addition is exercised, which is already covered by the existing "Test translation with denied type" case above.

To actually test the "allowed removes from built-in, then denied re-adds" path in `deniedOptionsFromOpts`, use a built-in denied type like `kernel_t`:

```suggestion
				DeniedTypes:  []string{"kernel_t"},
				AllowedTypes: []string{"kernel_t"},
```

and change the Allow block to reference `kernel_t` instead of `var_log_t`. This exercises: clone (`kernel_t` present) -> AllowedTypes removes it -> DeniedTypes re-adds it -> still denied.

saschagrunert (approved):

saschagrunert:
Thanks!

PR #3304, opened 2026-07-03
fix: skip binding mutation on pod UPDATE

### LINKED ISSUE #3296: binding webhook applies seccomp mutation on pod UPDATE, which the API server always rejects - pre-existing matching pods become un-updatable
#### What happened:

The binding webhook registers for pod CREATE, UPDATE, and DELETE, and applies the same mutation logic to UPDATE as to CREATE. When any client updates a pod whose image matches a `ProfileBinding`, but which does not yet have a seccomp profile (ie: a pod created before the binding existed, or before the namespace was labeled w/ `spo.x-k8s.io/enable-binding=`), the webhook patches `securityContext.seccompProfile` into the update. Container `securityContext` is immutable on live pods, so the API server rejects the entire update:

```
 Pod "..." is invalid: spec: Forbidden: pod updates may not change fields other than
  `spec.containers[*].image`, `spec.initContainers[*].image`, `spec.activeDeadlineSeconds`,
  `spec.tolerations` (only additions), `spec.terminationGracePeriodSeconds` ...
    -    "SeccompProfile": null,
    +    "SeccompProfile": { "Type": "Localhost", "LocalhostProfile": "operator/my-profile.json" },
```

This means that every matching pod that predates the binding can no longer be updated by anything until it is deleted. In our system this broke a controller that assigns pods to workloads via a label update until we removed the namespace label.

Per the documented semantics (["Once the binding is created and the Pod is created or recreated, the SeccompProfile should be applied to the container whose image matches the binding"](https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/installation-usage.md?plain=1#L1771-L1773)), the mutation should apply at pod creation only. Updates to pre-existing pods should pass through unmodified. Mutation-on-UPDATE seems to have no success case at all anyways - for an un-mutated pod the patch is always rejected by pod-spec immutability, and for an already-bound pod the webhook already skips.

1. Create a pod (w/ no seccompProfile in spec, and no enable-binding label on the namespace yet)
2. Install a `SeccompProfile` and a `ProfileBinding` with `spec.image` pointing to the pod's exact image in the pod's namespace
3. Label the namespace w/ `spo.x-k8s.io/enable-binding=`
4. Try to update the pod (ie: `kubectl -n <ns> label pod <pod> foo=bar`)
5. Metadata only change is rejected w/ the `spec: Forbidden` error above

Workaround seems to be: after labeling a namespace, immediately delete all unbound pods whose image matches a binding (since replacements will be bound at creation).

#### Environment:

- SPO v0.10.1
- Kubernetes v1.35

saschagrunert:
#### What type of PR is this?

/kind bug

The binding webhook registered for CREATE, UPDATE, and DELETE on pods but applied the same mutation logic to UPDATE as to CREATE. Since pod security context fields (seccompProfile, selinuxOptions, appArmorProfile) are immutable after creation, mutating on UPDATE always gets rejected by the API server. This makes pre-existing pods (created before the binding or namespace label existed) completely un-updatable.

Changes:
- Skip mutation for non-CREATE/non-DELETE operations in the binding handler
- Remove UPDATE from the binding webhook's registered operations and regenerated `deploy/webhook-operator.yaml`
- Use typed `admissionv1` constants instead of raw strings for operation checks
- Add unit test for UPDATE skip behavior (asserts response message), set `Operation` explicitly on all existing test cases, and add webhook config test asserting operations per webhook

The recording webhook keeps UPDATE since it has different semantics.

Fixes #3296

#### Does this PR have test?

Yes. Added a test case verifying that UPDATE requests return `Allowed` with message `"pod update, skipping mutation"` and no patches. Updated all existing test cases to set `Operation` explicitly. Added assertions in `TestWebhook_getWebhookConfig` verifying that the binding webhook registers CREATE+DELETE and the recording webhook registers CREATE+UPDATE+DELETE.

```release-note
Fixed binding webhook mutating pods on UPDATE, which the API server always rejects since security context fields are immutable after creation. Pre-existing pods matching a ProfileBinding are no longer blocked from updates.
```

### DIFF
--- a/deploy/overlays/webhook/webhook_config.yaml
+++ b/deploy/overlays/webhook/webhook_config.yaml
@@ -12,7 +12,7 @@ webhooks:
     timeoutSeconds: 5
     sideEffects: None
     rules:
-      - operations: ["CREATE", "UPDATE", "DELETE"]
+      - operations: ["CREATE", "DELETE"]
         apiGroups: ["*"]
         apiVersions: ["v1"]
         resources: ["pods"]
--- a/deploy/webhook-operator.yaml
+++ b/deploy/webhook-operator.yaml
@@ -6799,7 +6799,6 @@ webhooks:
     - v1
     operations:
     - CREATE
-    - UPDATE
     - DELETE
     resources:
     - pods
--- a/internal/pkg/manager/spod/bindata/webhook.go
+++ b/internal/pkg/manager/spod/bindata/webhook.go
@@ -44,7 +44,19 @@ var (
 	caBundle                      = []byte("Cg==")
 	sideEffects                   = admissionregv1.SideEffectClassNone
 	admissionReviewVersions       = []string{"v1beta1"}
-	rules                         = []admissionregv1.RuleWithOperations{
+	bindingRules                  = []admissionregv1.RuleWithOperations{
+		{
+			Operations: []admissionregv1.OperationType{
+				"CREATE", "DELETE",
+			},
+			Rule: admissionregv1.Rule{
+				APIGroups:   []string{""},
+				APIVersions: []string{"v1"},
+				Resources:   []string{"pods"},
+			},
+		},
+	}
+	recordingRules = []admissionregv1.RuleWithOperations{
 		{
 			Operations: []admissionregv1.OperationType{
 				"CREATE", "UPDATE", "DELETE",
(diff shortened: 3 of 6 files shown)


saschagrunert:

saschagrunert:

saschagrunert:

ccojocar (approved):

PR #3321, opened 2026-07-16
feat: add SELinux options in helm chart

mirza-src:
#### What type of PR is this?

/kind feature
/kind deprecation

The Helm chart currently only exposes the SELinux `enable` flag (`enableSelinux`), while the `SecurityProfilesOperatorDaemon` CRD supports a much richer `selinux` block: `enableRawSelinuxProfiles`, `typeTag`, and others `options`. Users installing via Helm had no way to set these without manually patching the `spod` CR after install.

This PR adds a `selinux` value to the chart that mirrors the CRD structure:

```yaml
selinux:
  enable: false
  enableRawSelinuxProfiles: true
  typeTag: spc_t
  options: {}
```

N/A

**Deprecation of `enableSelinux`:** I thought it was best to deprecate the top-level `enableSelinux` value in favor of `selinux.enable`. With the rest of the SELinux configuration now living under `selinux` key, keeping the enable flag as a separate top-level value would split one logical configuration block across two places. `enableSelinux` still works, so existing values files should be backwards compatible.

```release-note
Exposed the full SELinux configuration of the SPOD in Helm chart values. The top-level `enableSelinux` value is deprecated in favor of `selinux.enable`.
```

### DIFF
--- a/deploy/helm/templates/config.yaml
+++ b/deploy/helm/templates/config.yaml
@@ -20,7 +20,7 @@ spec:
   selinux:
     enable: {{ or .Values.enableSelinux .Values.selinux.enable }}
     enableRawSelinuxProfiles: {{ .Values.selinux.enableRawSelinuxProfiles }}
-    typeTag: {{ .Values.selinux.typeTag }}
+    typeTag: {{ .Values.selinux.typeTag | quote }}
     {{- with .Values.selinux.options }}
     options:
       {{- toYaml . | nindent 6 }}
--- a/deploy/helm/templates/config.yaml
+++ b/deploy/helm/templates/config.yaml
@@ -18,7 +18,13 @@ spec:
   daemonResourceRequirements:
     {{- toYaml .Values.daemon.resources | nindent 4 }}
   selinux:
-    enable: {{ .Values.enableSelinux }}
+    enable: {{ or .Values.enableSelinux .Values.selinux.enable }}
+    enableRawSelinuxProfiles: {{ .Values.selinux.enableRawSelinuxProfiles }}
+    typeTag: {{ .Values.selinux.typeTag }}
+    {{- with .Values.selinux.options }}
+    options:
+      {{- toYaml . | nindent 6 }}
+    {{- end }}
   enricher:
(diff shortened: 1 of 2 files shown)


deploy/helm/templates/config.yaml
```diff
@@ -18,7 +18,13 @@ spec:
   daemonResourceRequirements:
     {{- toYaml .Values.daemon.resources | nindent 4 }}
   selinux:
-    enable: {{ .Values.enableSelinux }}
+    enable: {{ or .Values.enableSelinux .Values.selinux.enable }}
```
saschagrunert: The `or` template function acts as a logical OR: it returns the first truthy value. This means if a user has `enableSelinux: true` in an existing values file and sets `selinux.enable: false` to migrate, SELinux stays enabled. The defaults work fine, but it creates a confusing migration path.

Consider using `ternary` so the deprecated flag only applies when the new one is not explicitly set, or at minimum add a comment explaining that `enableSelinux: true` always wins.

deploy/helm/templates/config.yaml
```diff
@@ -18,7 +18,13 @@ spec:
...
+    enableRawSelinuxProfiles: {{ .Values.selinux.enableRawSelinuxProfiles }}
+    typeTag: {{ .Values.selinux.typeTag }}
```
saschagrunert: Quoting the value would be safer against special YAML characters:
```suggestion
    typeTag: {{ .Values.selinux.typeTag | quote }}
```

deploy/helm/values.yaml
```diff
@@ -24,7 +24,13 @@ selinuxdImage:
     repository: security-profiles-operator/selinuxd-fedora
     tag: latest
 
+# Deprecated, use selinux.enable instead
 enableSelinux: false
+selinux:
+  enable: false
+  enableRawSelinuxProfiles: true
+  typeTag: spc_t
+  options: {}
```
saschagrunert, line 33: A commented-out example would help users discover the available sub-fields:
```suggestion
  options: {}
  # allowedSystemProfiles:
  #   - container
```

saschagrunert (approved):

PR #3333, opened 2026-07-21
remove duplicate run method

vickeybrown:
#### What type of PR is this?
/kind cleanup

PR kubernetes-sigs/security-profiles-operator#2871 introduced mainctx with signal handling and app.RunContext() for graceful shutdown of the JSON Log Enricher, but added it after the existing app.Run() call instead of replacing it. Since app.Run() blocks until the subcommand completes (or exits on error via os.Exit(1)), the RunContext call was dead code and mainctx was never actually used.

None 

#### Does this PR have test?
N/A

```release-note
NONE
```

### DIFF
--- a/cmd/security-profiles-operator/main.go
+++ b/cmd/security-profiles-operator/main.go
@@ -408,14 +408,10 @@ func main() {
 		},
 	}
 
-	if err := app.Run(os.Args); err != nil {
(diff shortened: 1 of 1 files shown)


saschagrunert (approved):
Thanks!

PR #3355, opened 2026-08-05
feat(recording): expose per-syscall coverage on merged seccomp profiles

### LINKED ISSUE #3354: Should ProfileRecording expose per-syscall observation coverage across merged recordings?

Not a concrete design yet — a question first: would per-syscall **observation
coverage** across merged recordings be considered in scope for
`ProfileRecording`?

Concretely: when `mergeStrategy: Containers` unions several recorded
instances into one `SeccompProfile`, today the union only keeps *whether* a
syscall was ever seen. It doesn't keep *how many* of the merged instances
contained it. That information exists in memory during the merge, for a
moment, before the union discards it.

Example — a profile merged from 5 recorded instances:

```
read     -> observed in 5 of 5 merged recorded instances
write    -> observed in 5 of 5 merged recorded instances
openat   -> observed in 5 of 5 merged recorded instances
socket   -> observed in 1 of 5 merged recorded instances
```

`read`/`write`/`openat` being 5/5 says "every recorded instance needed this."
`socket` at 1/5 doesn't mean "remove it" — the generated profile still
allows it, unchanged — it means "this looks like it belongs to a path only
some instances exercised, might be worth a look." Purely descriptive
metadata alongside the profile, not a change to it.

**Why this specific shape, and not "confidence" or a probability:** merged
instances aren't independent statistical samples (they're often the same
Deployment's replicas, or restarts of the same container), so a probability
or confidence-score framing would overstate what this actually is. "Observed
in N of M" is a plain, honest count of what the merge reconciler already
saw — nothing inferred.

One possible implementation: compute the per-syscall count in the existing
merge loop (`recordingmerger`/`merge_utils.go`), before the union collapses
it, and expose the result as advisory metadata alongside the generated
profile — without touching the recorder, the eBPF/log collection path, or
the profile's own enforced content. That's one option, not a proposal for
this issue to settle; whether it's an annotation, a status field, a separate
report object, or better left to downstream tooling entirely seems like it
should be a maintainer call, not something to presuppose here.

#### Why is this needed:

Right now, once several recordings are merged, there's no way to tell which
rules in the result were consistently exercised versus which came from a
single instance's specific execution path. That distinction seems useful
when deciding how much scrutiny a given rule deserves before an enforced
profile goes to production — without requiring anyone to inspect each
partial recording by hand, since the merge step already has that
information in memory before it's discarded.

#### User story covered

Closest existing one in `doc/user-stories.md` is Ming's ("As an application
developer... automatically generate initial security profiles that are
specific to the application") — this would extend that generation step with
more context to evaluate the result, rather than being a new capability on
its own. I didn't see an exact existing story for "understand how
consistently a rule was observed across merged recordings" specifically —
happy to help draft one in `doc/user-stories.md`/`doc/personas.md` if this
is a direction you'd want to pursue.

---

Happy to explore this further (including a possible small PR scoped to
merge-time-only, no recorder/API changes) if it's a direction that fits —
but wanted to check whether the idea itself belongs here before proposing
anything more concrete.

idriss-eliguene:
#### What type of PR is this?

/kind feature

When partial `SeccompProfile` resources are merged from a `ProfileRecording`
using `mergeStrategy: Containers`, the resulting profile currently retains only
the final union of observed syscalls.

Consequently, the merge loses information about how many collected partial
profiles contained each syscall.

This PR preserves that information by adding the following informational
annotation to the merged `SeccompProfile`:

`spo.x-k8s.io/syscall-coverage`

For example:

```json
{
  "version": "v1",
  "total": 3,
  "syscalls": {
    "read": 3,
    "socket": 1,
    "write": 3
  }
}
```

In this representation:

- `total` is the number of collected partial profiles included in the merge;
- each syscall value is the number of those partial profiles containing that
  syscall;
- a syscall contributes at most once per partial profile, even if it appears
  multiple times or in multiple syscall rules.

The counts are computed in the existing recording-merger path before the
current syscall union discards the per-partial occurrence information.

The annotation is informational only. It does not modify:

- the merged syscall allowlist;
- the default seccomp action;
- profile architectures;
- enforcement behavior;
- the recording lifecycle.

The annotation describes observation coverage across collected partial
profiles. It does not represent syscall invocation frequency, probability,
statistical confidence, or independent executions.

The implementation is intentionally limited to merged `SeccompProfile`
resources. It does not modify API types, CRDs, the BPF recorder, the log
enricher, or the recording webhook.

Fixes #3354

#### Does this PR have test?

Yes.

The following checks completed successfully:

- `go test ./internal/pkg/manager/recordingmerger/...`
- `go vet ./internal/pkg/manager/recordingmerger/...`
- related manager and `ProfileRecording` package tests
- `gofmt` verification

The tests cover:

- syscall counts across heterogeneous partial profiles;
- homogeneous partial profiles;
- a single partial profile;
- duplicate syscall names within one syscall rule;
- the same syscall appearing in multiple rules within one partial profile;
- deterministic JSON serialization;
- deterministic output when partial-profile ordering changes;
- source partial profiles remaining unmodified;
- unchanged syscall-union behavior;
- isolation between container merge groups;
- initialization of a nil annotations map;
- preservation of unrelated annotations;
- replacement of an existing controller-owned coverage annotation;
- exclusion of SELinux and AppArmor merge paths;
- integration through `createUpdateProfile` using the repository's fake-client
  pattern.

An end-to-end BPF recording test was also attempted.

The modified manager and SPOD images deployed successfully, the BPF recorder
was enabled, and the recording webhook annotated the test Pods. In the tested
environment, however, the recording pipeline did not produce partial
`SeccompProfile` resources. The recording-merger path therefore could not be
reached end to end.

This change consumes partial profiles already produced by the existing
recording pipeline and does not modify their collection or creation.

The JSON value is versioned so its format can be evolved explicitly if needed.

The annotation is deliberately described as observation coverage rather than
confidence. Recorded container instances may exercise different code paths and
are not necessarily statistically independent.

A low count must not be used automatically to remove a syscall from the merged
profile. A syscall observed in only one partial profile may still be required
by a valid but infrequently exercised code path.

The implementation remains additive and backward compatible:

- no API or CRD changes;
- no change to the generated seccomp specification;
- no changes to recorder or webhook components;
- no effect on SELinux or AppArmor profile merges.

```release-note
Merged SeccompProfiles generated from ProfileRecording using
mergeStrategy: Containers now include a spo.x-k8s.io/syscall-coverage
annotation containing per-syscall observation counts across the collected
partial profiles.
```

### DIFF
--- a/internal/pkg/manager/recordingmerger/coverage.go
+++ b/internal/pkg/manager/recordingmerger/coverage.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package recordingmerger
+
+import (
+	"encoding/json"
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// syscallCoverageAnnotation is set on a merged SeccompProfile (mergeStrategy=
+// Containers) recording, per syscall, how many collected partials contained it.
+// Informational only; never affects the enforced profile.
+const syscallCoverageAnnotation = "spo.x-k8s.io/syscall-coverage"
+
+// syscallCoverageSchemaVersion is the annotation value's schema version.
+const syscallCoverageSchemaVersion = "v1"
+
+// syscallCoverage is the JSON document stored in syscallCoverageAnnotation:
+// Total merged partials, and per syscall how many of them contained it.
+type syscallCoverage struct {
+	Version  string         `json:"version"`
+	Total    int            `json:"total"`
+	Syscalls map[string]int `json:"syscalls"`
+}
+
+func seccompCoverageAnnotation(partials []mergeableProfile) (string, error) {
+	if len(partials) == 0 {
+		return "", nil
+	}
+
+	counts := make(map[string]int)
+
+	for _, partial := range partials {
+		seccompPartial, ok := partial.(*mergeableSeccompProfile)
+		if !ok {
+			// mergeTypedProfiles groups profiles by type before calling this
+			// function, so mixed profile types are not expected. Coverage is
+			// informational and must not prevent profile merging, so safely omit
+			// the annotation rather than returning an error.
+			return "", nil
+		}
+
+		seen := make(map[string]struct{})
+
+		for i := range seccompPartial.Spec.Syscalls {
+			for _, name := range seccompPartial.Spec.Syscalls[i].Names {
+				seen[name] = struct{}{}
+			}
+		}
+
+		for name := range seen {
+			counts[name]++
+		}
+	}
+
+	coverage := syscallCoverage{
+		Version:  syscallCoverageSchemaVersion,
+		Total:    len(partials),
+		Syscalls: counts,
+	}
+
+	data, err := json.Marshal(coverage)
+	if err != nil {
+		return "", fmt.Errorf("marshal syscall coverage: %w", err)
+	}
+
+	return string(data), nil
+}
+
+func setSyscallCoverageAnnotation(obj metav1.Object, value string) {
+	if value == "" {
+		return
+	}
+
+	annotations := obj.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+
+	annotations[syscallCoverageAnnotation] = value
+	obj.SetAnnotations(annotations)
+}
--- a/internal/pkg/manager/recordingmerger/coverage_test.go
+++ b/internal/pkg/manager/recordingmerger/coverage_test.go
@@ -1,0 +1,613 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package recordingmerger
+
+import (
+	"context"
+	"encoding/json"
+	"sort"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	apparmorprofileapi "sigs.k8s.io/security-profiles-operator/api/apparmorprofile/v1"
+	profilebase "sigs.k8s.io/security-profiles-operator/api/profilebase/v1"
+	profilerecordingapi "sigs.k8s.io/security-profiles-operator/api/profilerecording/v1"
+	seccompprofile "sigs.k8s.io/security-profiles-operator/api/seccompprofile/v1"
+	selinuxprofileapi "sigs.k8s.io/security-profiles-operator/api/selinuxprofile/v1"
+)
+
+// seccompPartial builds a mergeable seccomp partial profile from a set of rules,
+// where each rule is a (action, names) pair.
+func seccompPartial(name string, rules ...seccompprofile.Syscall) mergeableProfile {
+	return &mergeableSeccompProfile{
+		SeccompProfile: seccompprofile.SeccompProfile{
+			ObjectMeta: metav1.ObjectMeta{Name: name},
+			Spec: seccompprofile.SeccompProfileSpec{
+				DefaultAction: seccompprofile.ActErrno,
+				Syscalls:      rules,
+			},
+		},
+	}
+}
+
+func allow(names ...string) seccompprofile.Syscall {
+	return seccompprofile.Syscall{Action: seccompprofile.ActAllow, Names: names}
+}
+
+// parseCoverage unmarshals the annotation value into the coverage struct.
+func parseCoverage(t *testing.T, value string) syscallCoverage {
+	t.Helper()
+
+	var cov syscallCoverage
+	require.NoError(t, json.Unmarshal([]byte(value), &cov))
+
+	return cov
+}
+
+func TestSeccompCoverageAnnotation(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name         string
+		partials     []mergeableProfile
+		expectEmpty  bool
+		expectTotal  int
+		expectCounts map[string]int
+	}{
+		{
+			name: "heterogeneous",
+			partials: []mergeableProfile{
+				seccompPartial("a", allow("read", "write", "openat")),
+				seccompPartial("b", allow("read", "write", "socket")),
+				seccompPartial("c", allow("read", "write")),
+			},
+			expectTotal:  3,
+			expectCounts: map[string]int{"read": 3, "write": 3, "openat": 1, "socket": 1},
+		},
+		{
+			// All counts equal the total; no values are suppressed.
+			name: "homogeneous",
+			partials: []mergeableProfile{
+				seccompPartial("a", allow("read", "write")),
+				seccompPartial("b", allow("read", "write")),
+				seccompPartial("c", allow("read", "write")),
+			},
+			expectTotal:  3,
+			expectCounts: map[string]int{"read": 3, "write": 3},
+		},
+		{
+			// A single partial still produces coverage with total=1.
+			name: "single partial",
+			partials: []mergeableProfile{
+				seccompPartial("a", allow("read", "write")),
+			},
+			expectTotal:  1,
+			expectCounts: map[string]int{"read": 1, "write": 1},
+		},
+		{
+			// Duplicate names within one rule count once for that partial.
+			name: "duplicate names in one rule",
+			partials: []mergeableProfile{
+				seccompPartial("a", allow("read", "read", "write")),
+				seccompPartial("b", allow("read")),
+			},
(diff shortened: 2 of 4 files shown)


saschagrunert:
@idriss-eliguene thank you for the PR! Please sign the CLA to continue here.

saschagrunert:
/ok-to-test

CI is not happy, though.

idriss-eliguene:
checking for ci failed

idriss-eliguene:
@saschagrunert   
The previous CI issues have been addressed and all required Prow jobs are now passing.

The only remaining failure is the GitHub Actions test/e2e-apparmor-profile workflow.

From the logs, the failure occurs in the AppArmor recording path before the recording merger is reached, while this PR only changes the merged SeccompProfile path.

idriss-eliguene:
Looked into the Codecov flags:

**Patch coverage (10 lines missing):** all 10 lines are inside `mergeTypedProfiles`, `createUpdateSeccompProfile`, `createUpdateSelinuxProfile`, and `createUpdateApparmorProfile` in `recordingmerger.go`. Checked against `main` directly: these functions were already at 0% unit coverage before this PR (they're exercised by the e2e suite, not `go test`, in this codebase). This PR only threads a new `coverageAnnotation` parameter through them — it doesn't add new untested logic. The actual new code, `coverage.go`, is at 94-100%, and `createUpdateProfile` (previously 0%) is now covered at 86.7% by the new integration test in `coverage_test.go`. Package coverage for `recordingmerger` goes from 36.1% to 53.7% with this PR.

**Project coverage (-19.92%, "1393 commits behind main"):** the diff base Codecov used here is stale (79 files vs. head's 132), so it's comparing against a commit from well before this branch point rather than current `main` — not a real regression from this change.

internal/pkg/manager/recordingmerger/coverage.go
```diff
@@ -0,0 +1,94 @@
...
+	if len(partials) == 0 {
+		return "", nil
+	}
+
+	counts := make(map[string]int)
+
+	for _, partial := range partials {
+		seccompPartial, ok := partial.(*mergeableSeccompProfile)
+		if !ok {
+			return "", nil
```
saschagrunert, line 56: Nit: if a non-seccomp partial ever ended up in this slice (e.g. after a future caller change), the function silently returns `""` after partially populating `counts`. That can't happen today since `mergeTypedProfiles` groups by kind, but the silent discard could mask a bug. A brief comment on why `return "", nil` (rather than an error) is intentional here would help.

internal/pkg/manager/recordingmerger/recordingmerger.go
```diff
@@ -179,7 +179,14 @@ func (r *PolicyMergeReconciler) mergeTypedProfiles(
 
 		mergedRecordingName := mergedProfileName(profileRecording.Name, cntPartialProfiles[0])
 
-		res, err := createUpdateMergedProfile(ctx, r.client, profileRecording, mergedRecordingName, mergedProfile)
+		// Informational only; empty for non-seccomp kinds.
+		coverageAnnotation, err := seccompCoverageAnnotation(cntPartialProfiles)
```
saschagrunert: Consider logging the computed coverage at debug level. The reconciler logs the create/update result at line 192 but not the coverage value, which would help troubleshoot cases where the annotation is unexpectedly empty or missing.

```suggestion
		coverageAnnotation, err := seccompCoverageAnnotation(cntPartialProfiles)
		if err != nil {
			return fmt.Errorf("cannot compute syscall coverage: %w", err)
		}

		r.log.V(1).Info("Computed syscall coverage", "container", cntName, "coverage", coverageAnnotation)
```

internal/pkg/manager/recordingmerger/coverage_test.go
```diff
@@ -0,0 +1,521 @@
...
+					allow("read"),
+					seccompprofile.Syscall{Action: seccompprofile.ActLog, Names: []string{"write"}},
+				),
+				seccompPartial("b", allow("read")),
+			},
+			expectTotal:  2,
+			expectCounts: map[string]int{"read": 2, "write": 1},
+		},
+		{
+			// Case 15: empty syscall rules; partial still contributes to total.
```
saschagrunert: Nit: case numbering jumps from 6 to 15 here, and the standalone tests below use non-sequential numbers (14, 11+18, 13, 7, 8-10, 12). Either renumber sequentially or drop the case numbers, since the `name` field already identifies each test.

internal/pkg/manager/recordingmerger/coverage.go
idriss-eliguene: Good point, thanks. The current caller guarantees homogeneous profile types because `mergeTypedProfiles` groups partial profiles by kind before invoking this function, so this path is not expected today.

I added a comment to document that invariant and to clarify that returning an empty annotation is intentional because syscall coverage is informational and should not prevent profile merging.

internal/pkg/manager/recordingmerger/coverage_test.go
idriss-eliguene: Good catch, thanks. I removed the case numbering and now rely on the descriptive test names, which makes the table easier to maintain as cases are added or reordered.

idriss-eliguene (commented):
all changes need have been done

internal/pkg/manager/recordingmerger/recordingmerger.go
```diff
@@ -179,7 +179,16 @@ func (r *PolicyMergeReconciler) mergeTypedProfiles(
...
```
saschagrunert: `mergeMergeableProfiles` at line 168 mutates `cntPartialProfiles[0]` in-place (its `Spec.Syscalls` becomes the union of all partials). By the time this call iterates the same slice, the first element contains every syscall, inflating counts.

Reproduction: with partials `A:[read,write]` and `B:[read,socket]`, `socket` is reported as 2 instead of 1.

Move this block (lines 182-188) above `mergeMergeableProfiles` at line 168.

internal/pkg/manager/recordingmerger/coverage_test.go
```diff
@@ -0,0 +1,513 @@
...
+		t.Parallel()
+
+		obj := &seccompprofile.SeccompProfile{}
+		setSyscallCoverageAnnotation(obj, "")
+		require.Nil(t, obj.GetAnnotations())
+	})
+}
+
+// This guards against the coverage work accidentally altering merged spec output.
+func TestMergeProfiles_SeccompUnionUnchanged(t *testing.T) {
```
saschagrunert, line 364: The unit tests call `seccompCoverageAnnotation` on fresh partials, but the reconciler calls it after `mergeMergeableProfiles` mutates the same slice. A test that calls merge then coverage on the same `[]mergeableProfile` would catch this:

```go
func TestCoverageAfterMerge(t *testing.T) {
    t.Parallel()
    partials := []mergeableProfile{
        seccompPartial("a", allow("read", "write")),
        seccompPartial("b", allow("read", "socket")),
    }
    _, err := mergeMergeableProfiles(partials)
    require.NoError(t, err)
    value, err := seccompCoverageAnnotation(partials)
    require.NoError(t, err)
    cov := parseCoverage(t, value)
    require.Equal(t, 1, cov.Syscalls["socket"])
}
```

saschagrunert (changes requested):
Previous review comments were all addressed, thanks. One blocking issue remains: the coverage computation runs after the merge mutates the first partial in-place, inflating the counts.

internal/pkg/manager/recordingmerger/recordingmerger.go
```diff
@@ -165,6 +165,12 @@ func (r *PolicyMergeReconciler) mergeTypedProfiles(
 	for cntName, cntPartialProfiles := range partialProfiles {
 		r.log.Info("Merging profiles for container", "container", cntName)
 
+		// Informational only; empty for non-seccomp kinds.
+		coverageAnnotation, err := seccompCoverageAnnotation(cntPartialProfiles)
+		if err != nil {
+			return fmt.Errorf("cannot compute syscall coverage: %w", err)
```
saschagrunert: The annotation is described as informational-only and explicitly not affecting the recording lifecycle. Returning an error here aborts the entire merge for this container group.

In practice `json.Marshal` on `map[string]int` cannot fail, so this is theoretical. But for consistency with the stated contract, consider logging the error and continuing with an empty annotation:

```suggestion
			r.log.Error(err, "Cannot compute syscall coverage, skipping annotation", "container", cntName)
```

internal/pkg/manager/recordingmerger/coverage_test.go
```diff
@@ -0,0 +1,592 @@
...
+		[]string{"openat", "read", "socket", "write"},
+		names,
+	)
+
+	// The merge itself never sets the coverage annotation; that is done in the
+	// controller path (createUpdateProfile), not in MergeProfiles.
+	require.NotContains(t, merged.GetAnnotations(), syscallCoverageAnnotation)
+}
+
+func TestMergeTypedProfiles_ComputesCoverageBeforeMerge(t *testing.T) {
```
saschagrunert, line 437: The existing assertion at line 470 (`socket: 1`) already catches the ordering regression (merge-then-coverage would inflate `socket` to 2). A dedicated unit test that calls `mergeMergeableProfiles` first and then `seccompCoverageAnnotation` on the same slice would make the ordering constraint more self-documenting for future contributors:

```go
func TestCoverageMustRunBeforeMerge(t *testing.T) {
	t.Parallel()
	partials := []mergeableProfile{
		seccompPartial("a", allow("read", "write")),
		seccompPartial("b", allow("read", "socket")),
	}
	// Correct ordering: coverage sees the original partials.
	value, err := seccompCoverageAnnotation(partials)
	require.NoError(t, err)
	before := parseCoverage(t, value)
	require.Equal(t, 1, before.Syscalls["socket"])

	// merge mutates partials[0] in-place (union of all syscalls).
	_, err = mergeMergeableProfiles(partials)
	require.NoError(t, err)

	// After merge, partials[0] contains {read, write, socket},
	// so coverage now over-counts socket.
	value, err = seccompCoverageAnnotation(partials)
	require.NoError(t, err)
	after := parseCoverage(t, value)
	require.Equal(t, 2, after.Syscalls["socket"],
		"inflated count proves coverage must run before merge")
}
```

Optional, since the integration test already covers the regression.

idriss-eliguene: Good point, thanks. The existing assertion already detects the regression indirectly, but I agree that a dedicated test makes the ordering dependency much more explicit for future readers.

I'll add a focused regression test covering that specific mutation scenario.

internal/pkg/manager/recordingmerger/recordingmerger.go
idriss-eliguene: That's a good point. The coverage annotation is intended to be informational only, so aborting the merge is stronger than the intended contract.

Although `json.Marshal` on the current data structure is effectively non-failing today, I agree it's better to keep the implementation aligned with the documented semantics. I'll switch this to log the error and continue without the annotation.

idriss-eliguene:
The latest test workflow failure appears to be infrastructure-related: the `e2e-fedora` and `e2e-apparmor-profile` hosted runners lost communication with GitHub. The relevant unit, analysis, coverage, and lint checks are passing. I’m retriggering the failed jobs.

saschagrunert (approved):
Thanks!

PR #3357, opened 2026-08-06
feat: add support for custom selinux templates

### LINKED ISSUE #3356: Allow mounting extra volumes into SPOD pod containers

`spec.extraVolumes` and `spec.extraVolumeMounts` fields on the `SecurityProfilesOperatorDaemon` CRD, allowing injection of arbitrary volumes and mounts into the SPOD pods.

#### Why is this needed:

On refpolicy-based distributions such as Flatcar Linux, the selinuxd image bundles udica templates from `container-selinux` that reference Fedora-specific SELinux types that are incompatible with refpolicy. This causes the `selinux-shared-policies-copier` init container to fail during `semodule -i /usr/share/selinuxd/templates/*.cil`. Related issue: https://github.com/containers/selinuxd/issues/166.

The `extraVolumes` and `extraVolumeMounts` can provide a workaround to mount refpolicy-compatible CIL templates over `/usr/share/selinuxd/templates/` in the init container:

```yaml
spec:
  selinux:
    enable: true
  extraVolumes:
    - name: refpolicy-templates
      configMap:
        name: selinuxd-refpolicy-templates
  extraVolumeMounts:
    - name: refpolicy-templates
      mountPath: /usr/share/selinuxd/templates
      readOnly: true
      containers:
        - selinux-shared-policies-copier
```

mirza-src:
#### What type of PR is this?

/kind feature
/kind api-change

Adds a `customTemplatesConfigMap` field to the `SPODSelinuxConfig` API that allows users to supply a ConfigMap containing `.cil` files which replace the bundled selinuxd templates directory entirely. This is required to make SPO functional on distributions such as Flatcar Linux, whose SELinux policy base (refpolicy) is incompatible with the templates shipped with selinuxd.

When `customTemplatesConfigMap` is set, the referenced ConfigMap is mounted over `/usr/share/selinuxd/templates/` in the `selinux-shared-policies-copier` init container.

Fixes #3356

#### Does this PR have test?

N/A

N/A

```release-note
Added `selinux.customTemplatesConfigMap` field to the SPOD API and Helm chart, allowing users to replace the bundled selinuxd CIL templates with a custom ConfigMap. This enables SPO to run on distributions whose SELinux policy base is incompatible with the shipped templates, such as Flatcar Linux.
```

### DIFF
--- a/api/spod/v1/spod_types.go
+++ b/api/spod/v1/spod_types.go
@@ -225,7 +225,8 @@ type SPODSelinuxConfig struct {
 	// files that replace the bundled selinuxd templates entirely. The ConfigMap
 	// must exist in the same namespace as the SPOD daemonset. Use this on
 	// distributions (e.g. Flatcar Linux) whose SELinux policy base is incompatible
-	// with the templates shipped with selinuxd.
+	// with the templates shipped with selinuxd. Note: changes to the ConfigMap
+	// contents require restarting the DaemonSet pods to take effect.
 	// +optional
 	// +kubebuilder:validation:MaxLength=253
 	CustomTemplatesConfigMap string `json:"customTemplatesConfigMap,omitempty"`
--- a/api/spod/v1/spod_types.go
+++ b/api/spod/v1/spod_types.go
@@ -229,6 +229,7 @@ type SPODSelinuxConfig struct {
 	// contents require restarting the DaemonSet pods to take effect.
 	// +optional
 	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
 	CustomTemplatesConfigMap string `json:"customTemplatesConfigMap,omitempty"`
 }
 
--- a/api/spod/v1/spod_types.go
+++ b/api/spod/v1/spod_types.go
@@ -221,6 +221,14 @@ type SPODSelinuxConfig struct {
 	// +optional
 	// +default={}
 	Options SelinuxOptions `json:"options,omitzero,omitempty"`
+	// customTemplatesConfigMap if defined, names a ConfigMap containing .cil
+	// files that replace the bundled selinuxd templates entirely. The ConfigMap
+	// must exist in the same namespace as the SPOD daemonset. Use this on
+	// distributions (e.g. Flatcar Linux) whose SELinux policy base is incompatible
+	// with the templates shipped with selinuxd.
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
+	CustomTemplatesConfigMap string `json:"customTemplatesConfigMap,omitempty"`
 }
 
 // SPODEnricherConfig contains log enricher, JSON enricher, and BPF recorder configuration.
--- a/internal/pkg/manager/spod/spod_controller.go
+++ b/internal/pkg/manager/spod/spod_controller.go
@@ -591,6 +591,8 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 
 			return nil, fmt.Errorf("unable to mount custom SELinux templates: %w", err)
 		}
+	} else if cfg.Spec.Selinux.CustomTemplatesConfigMap != "" {
+		r.log.Info("customTemplatesConfigMap is set but SELinux is disabled, the field will be ignored")
 	}
 
 	// Custom host proc volume
--- a/internal/pkg/manager/spod/spod_controller.go
+++ b/internal/pkg/manager/spod/spod_controller.go
@@ -185,7 +185,11 @@ func (r *ReconcileSPOd) Reconcile(ctx context.Context, req reconcile.Request) (r
 		return reconcile.Result{}, fmt.Errorf("get ca inject type: %w", err)
 	}
 
-	configuredSPOd := r.getConfiguredSPOd(ctx, spod, image, pullPolicy, caInjectType)
+	configuredSPOd, err := r.getConfiguredSPOd(ctx, spod, image, pullPolicy, caInjectType)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("get configured SPOD: %w", err)
+	}
+
 	webhook := r.getConfiguredWebook(spod, image, pullPolicy, caInjectType)
 	metricsService := bindata.GetMetricsService(r.namespace, caInjectType)
 	serviceMonitor := bindata.ServiceMonitor(caInjectType,
@@ -525,7 +529,7 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 	image string,
 	pullPolicy corev1.PullPolicy,
 	caInjectType bindata.CAInjectType,
-) *appsv1.DaemonSet {
+) (*appsv1.DaemonSet, error) {
 	newSPOd := r.baseSPOd.DeepCopy()
 
 	newSPOd.SetName(cfg.GetName())
@@ -583,8 +587,9 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 			fmt.Sprintf("--with-raw-selinux=%t", enableRawSelinux))
 
 		if err := addSelinuxCustomTemplatesVolume(cfg, templateSpec); err != nil {
-			r.log.Info("Unable to mount custom SELinux templates", "error", err)
 			r.record.Event(cfg, util.EventTypeWarning, reasonCannotMountCustomTemplates, err.Error())
+
+			return nil, fmt.Errorf("unable to mount custom SELinux templates: %w", err)
 		}
 	}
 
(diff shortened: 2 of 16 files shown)


internal/pkg/manager/spod/spod_controller.go
```diff
@@ -580,6 +580,14 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 		templateSpec.Containers[bindata.ContainerIDDaemon].Args = append(
 			templateSpec.Containers[bindata.ContainerIDDaemon].Args,
 			fmt.Sprintf("--with-raw-selinux=%t", enableRawSelinux))
+
+		if cfg.Spec.Selinux.CustomTemplatesConfigMap != "" {
```
saschagrunert: Missing test coverage for this branch. Please add a test case to the existing controller reconciliation tests that verifies:
- volume + mount are added when `CustomTemplatesConfigMap` is non-empty
- nothing changes when it is empty

Same for `CustomTemplatesVolume` in `bindata/spod.go`.

api/spod/v1/spod_types.go
```diff
@@ -221,6 +221,12 @@ type SPODSelinuxConfig struct {
 	// +optional
 	// +default={}
 	Options SelinuxOptions `json:"options,omitzero,omitempty"`
+	// customTemplatesConfigMap if defined, names a ConfigMap containing .cil
+	// files that replace the bundled selinuxd templates entirely. Use this on
+	// distributions (e.g. Flatcar Linux) whose SELinux policy base is
+	// incompatible with the templates shipped with selinuxd.
+	// +optional
+	CustomTemplatesConfigMap string `json:"customTemplatesConfigMap,omitempty"`
```
saschagrunert, line 233: Consider adding `+kubebuilder:validation:MaxLength=253` since ConfigMap names are DNS subdomain names. This gives users a clear admission error instead of a pod scheduling failure.

api/spod/v1/spod_types.go
```diff
@@ -221,6 +221,12 @@ type SPODSelinuxConfig struct {
 	// +optional
 	// +default={}
 	Options SelinuxOptions `json:"options,omitzero,omitempty"`
+	// customTemplatesConfigMap if defined, names a ConfigMap containing .cil
```
saschagrunert, line 224: Nit: the description should mention that the ConfigMap must exist in the operator's namespace (the namespace where the SPOD DaemonSet runs).

api/spod/v1/spod_types.go
```diff
@@ -221,6 +221,14 @@ type SPODSelinuxConfig struct {
...
+	// files that replace the bundled selinuxd templates entirely. The ConfigMap
+	// must exist in the same namespace as the SPOD daemonset. Use this on
+	// distributions (e.g. Flatcar Linux) whose SELinux policy base is incompatible
+	// with the templates shipped with selinuxd.
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
+	CustomTemplatesConfigMap string `json:"customTemplatesConfigMap,omitempty"`
```
saschagrunert, line 233: Should this field also be added to the v1alpha1 `SPODSelinuxConfig` and wired through `ConvertTo`/`ConvertFrom` in `api/spod/v1alpha1/conversion.go`? A v1 -> v1alpha1 -> v1 round-trip will silently drop this value right now. If keeping it v1-only is intentional (same pattern as the deny/allow lists in `SelinuxOptions`), that's fine, just want to confirm.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -899,6 +901,24 @@ func (r *ReconcileSPOd) getConfiguredWebook(cfg *spodapi.SecurityProfilesOperato
...
 
+func addSelinuxCustomTemplatesVolume(
+	cfg *spodapi.SecurityProfilesOperatorDaemon,
+	templateSpec *corev1.PodSpec,
+) {
+	if cfg.Spec.Selinux.CustomTemplatesConfigMap == "" {
+		return
+	}
+
+	idx := bindata.InitContainerIDSelinuxSharedPoliciesCopier
```
saschagrunert: This relies on the init container's position in `templateSpec.InitContainers` matching `InitContainerIDSelinuxSharedPoliciesCopier` (1). That holds today because NonRootenabler is at 0 and the SELinux copier was just appended at 1, but if another init container is ever inserted between them, this silently mounts the volume on the wrong container. Consider finding the container by name instead.

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -143,6 +143,69 @@ func Test_getConfiguredJsonEnricherNilInterval(t *testing.T) {
...
+	}
+
+	templateSpec := &v1.PodSpec{
+		InitContainers: []v1.Container{{}, {Name: "test"}},
+	}
+
+	addSelinuxCustomTemplatesVolume(cfg, templateSpec)
+
+	require.Len(t, templateSpec.Volumes, 1)
+	require.Len(t, templateSpec.InitContainers[1].VolumeMounts, 1)
```
saschagrunert, line 215: These only check that one volume and one mount exist, but not what they contain. Worth asserting the volume references ConfigMap `"test-templates"` and the mount targets `/usr/share/selinuxd/templates`, to verify the API value flows through correctly.

api/spod/v1/spod_types.go
mirza-src: I see, no good reason to leave it out of v1alpha1, adding it there too.

saschagrunert, line 233: The API doc should mention that changes to the ConfigMap contents require restarting the DaemonSet pods to take effect, since the volume is mounted on an init container that only runs at pod startup.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -899,6 +901,26 @@ func (r *ReconcileSPOd) getConfiguredWebook(cfg *spodapi.SecurityProfilesOperato
...
+	templateSpec *corev1.PodSpec,
+) {
+	if cfg.Spec.Selinux.CustomTemplatesConfigMap == "" {
+		return
+	}
+
+	idx := slices.IndexFunc(templateSpec.InitContainers, func(c corev1.Container) bool {
+		return c.Name == bindata.SelinuxPoliciesCopierContainerName
+	})
+	if idx == -1 {
```
saschagrunert, line 926: This function is only called inside the `if enableSelinux {}` block, where the copier init container has already been appended to `templateSpec.InitContainers`. If `idx == -1` here while `CustomTemplatesConfigMap` is non-empty, it means the user's configuration is silently dropped. Consider logging a warning.

saschagrunert (commented):
Previous review items have been addressed. Two remaining suggestions.

internal/pkg/manager/spod/spod_controller.go
mirza-src: Added a log + Event (CannotMountCustomTemplates) for visibility via `kubectl describe`. Let me know if you'd rather drop the event

saschagrunert (commented):
Minor note: if a user sets `customTemplatesConfigMap` but SELinux is disabled, the field is silently ignored since the call site is inside the `if enableSelinux` block. Consider logging a warning for that case.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -580,6 +581,11 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 		templateSpec.Containers[bindata.ContainerIDDaemon].Args = append(
 			templateSpec.Containers[bindata.ContainerIDDaemon].Args,
 			fmt.Sprintf("--with-raw-selinux=%t", enableRawSelinux))
+
+		if err := addSelinuxCustomTemplatesVolume(cfg, templateSpec); err != nil {
+			r.log.Info("Unable to mount custom SELinux templates", "error", err)
```
saschagrunert: The error is swallowed here and logged at `Info` level. If the user explicitly set `customTemplatesConfigMap`, deploying the DaemonSet with bundled templates silently contradicts their intent.

In practice the error is currently unreachable (the init container is always appended on lines 568-570 just above), but the contract is still wrong: if `addSelinuxCustomTemplatesVolume` fails, the function should return the error rather than continue with a misconfigured DaemonSet.

At minimum, `r.log.Info` should be `r.log.Error`.

api/spod/v1/spod_types.go
```diff
@@ -221,6 +221,15 @@ type SPODSelinuxConfig struct {
...
 	// +default={}
 	Options SelinuxOptions `json:"options,omitzero,omitempty"`
+	// customTemplatesConfigMap if defined, names a ConfigMap containing .cil
+	// files that replace the bundled selinuxd templates entirely. The ConfigMap
+	// must exist in the same namespace as the SPOD daemonset. Use this on
+	// distributions (e.g. Flatcar Linux) whose SELinux policy base is incompatible
+	// with the templates shipped with selinuxd. Note: changes to the ConfigMap
+	// contents require restarting the DaemonSet pods to take effect.
+	// +optional
+	// +kubebuilder:validation:MaxLength=253
```
saschagrunert, line 231: nit: A `Pattern` constraint for DNS subdomain names (e.g. `+kubebuilder:validation:Pattern="^[a-z0-9]([a-z0-9.-]*[a-z0-9])?$"`) would catch invalid ConfigMap names at admission time rather than failing at pod scheduling. Not a blocker since many K8s operators rely on kubelet to report this, but it would be a stronger API contract.

saschagrunert (approved):
Thanks!

PR #3376, opened 2026-08-21
Include profile kind in node status naming and validate SELinux profile names

saschagrunert:
#### What type of PR is this?

/kind feature

Two enhancements to improve profile management reliability:

1. **Include profile kind in SecurityProfileNodeStatus naming**: Changes status object names from `<profileName>-<nodeName>` to `<kind>-<profileName>-<nodeName>` (e.g., `seccompprofile-myprofile-worker-1`). This prevents collisions when a SelinuxProfile and SeccompProfile share the same name, which previously caused status objects to overwrite each other. Includes migration logic to clean up old-format objects on upgrade, with a label ownership check to prevent cross-profile deletion in edge cases.

2. **Validate SELinux profile names against system modules**: At reconcile time on each node, the daemon checks whether a SELinux profile name conflicts with an existing module in the host's module store (`/var/lib/selinux`). If a conflict is detected on first installation, the profile enters an error state with a descriptive event. The check is skipped on pod restarts and upgrades to avoid false positives.

None

#### Does this PR have test?

Yes. Unit tests for kind-based naming (`TestPerNodeStatusNameIncludesKind`), legacy status migration (`TestRemoveLegacyNodeStatus`), and module conflict detection (`TestIsSystemSELinuxModule`).

The system module check only runs on first install (`firstInstall = !wasMigrated`). On upgrade, the legacy node status migration sets `wasMigrated = true`, suppressing the check. On pod restart, the node status CRD persists in etcd, so `Exists()` returns true and the check is skipped entirely.

```release-note
SecurityProfileNodeStatus objects now include the profile kind in their name to prevent collisions between different profile types. SELinux profile names that conflict with system SELinux policy modules are rejected at reconcile time with a descriptive error event.
```

### DIFF
--- a/internal/pkg/nodestatus/nodestatus.go
+++ b/internal/pkg/nodestatus/nodestatus.go
@@ -21,12 +21,14 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 
 	profilebase "sigs.k8s.io/security-profiles-operator/api/profilebase/v1"
 	profilerecordingapi "sigs.k8s.io/security-profiles-operator/api/profilerecording/v1"
@@ -34,6 +36,8 @@ import (
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/config"
 	"sigs.k8s.io/security-profiles-operator/internal/pkg/util"
 )
+
+var log = logf.Log.WithName("nodestatus")
 
 const (
 	partialProfileFinalizer = "spo.x-k8s.io/partial-profile-finalizer"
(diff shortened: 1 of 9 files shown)


saschagrunert:

internal/pkg/webhooks/validation/validation.go
```diff
@@ -30,12 +31,131 @@ import (
...
+// +kubebuilder:rbac:groups=security-profiles-operator.x-k8s.io,resources=selinuxprofiles,verbs=get;list;watch
+
+// systemSELinuxModules contains module names shipped in the base SELinux policy
+// on RHEL, CentOS Stream, and Fedora. SPO installs modules at priority 350,
+// overriding system modules at priority 100. If a user-created profile
+// collides with one of these names, the system module is replaced, breaking
+// types that other modules depend on (e.g., krb5_keytab_t from the kerberos
+// module).
+var systemSELinuxModules = map[string]bool{
+	"abrt":           true,
```
haircommander: is there a way to autodiscover instead of hardcode? this seems brittle and wouldn't work for other distros

saschagrunert: Good point, replaced the hardcoded list with node-side autodiscovery. The daemon container now mounts `/var/lib/selinux` read-only and walks the host's module store (`<policy_type>/active/modules/<priority>/<name>/`) to detect conflicts before writing policy files. This works for any distro and requires no maintenance.

The tradeoff is that the check happens at install time (surfaced via `SecurityProfileNodeStatus` error state and events) rather than at admission time, but that's consistent with how other node-side validation failures are already reported.

internal/pkg/nodestatus/nodestatus.go
```diff
@@ -61,28 +65,66 @@ func NewForProfile(pol profilebase.SecurityProfileBase, c client.Client) (*Statu
...
+
+	// Verify the object belongs to this profile. A profile named
+	// "<kind>-<other>" has a legacy name that collides with the
+	// new-format name of profile "<other>".
+	if old.Labels[secprofnodestatusapi.StatusToProfLabel] != util.KindBasedDNSLengthName(nsf.pol) {
+		return false
+	}
+
+	if err := nsf.client.Delete(ctx, old); err != nil && !kerrors.IsNotFound(err) {
+		log.Error(err, "failed to remove legacy node status", "name", legacyName)
```
rphillips, line 124: return false missing?

saschagrunert: Yes, added the missing return. 👍

PR #3377, opened 2026-08-22
fix(recording): avoid merged profile name collision

### LINKED ISSUE #3328: `mergeStrategy: Containers` doesn't merge when the recorded pods have a fixed name (no `generateName`)
#### What happened:

If you record a workload whose pods have a fixed `metadata.name` (no `generateName`) with `mergeStrategy: Containers`,
you never get a merged profile, the `ProfileRecording` gets stuck on deletion and the merge controller just loops on:

```
"Reconciler error" err="cannot recorded profiles: cannot merge profiles: cannot create or update merged profile:
action:  MutateFn cannot mutate object name and/or object namespace" controller="policymerger"
```

There seems to be two things going on, and both are in `v1.0.0` and `main`:

1. The recorder doesn't give fixed-name pods a unique suffix. `collectProfile`
   (`internal/pkg/daemon/profilerecorder/profilerecorder.go`) only sets `replicaSuffix` when the pod has a
   `generateName`. Without one, the partial ends up named `<recording>-<container>` - which is the exact name the
   merger wants for the merged profile (`mergedProfileName`). So the single partial is already sitting on the target
   name.
2. The merged profile is cluster-scoped but gets built with a namespace. `mergedObjectMeta`
   (`internal/pkg/manager/recordingmerger/merge_utils.go`) sets `Namespace: namespace`. `SeccompProfile` /
   `SelinuxProfile` / `AppArmorProfile` are cluster-scoped, so the live object has no namespace - and once (1) pushes
   `CreateOrUpdate` onto the update path, controller-runtime refuses the namespace change from `""` to the recording
   namespace. That's the `MutateFn` error.

Where this causes issues is anything that names its own pods. GitLab Runner's Kubernetes executor names build pods
with `generateNameForK8sResources(...)` and sets that as a fixed `metadata.name` (no `generateName`) on both v19.0.1 and
`main`, so every recorded runner build pod hits it.

Deleting the recording merges the per-pod partials into `<recording>-<container>`, it reaches `Installed`, and the
partials get cleaned up - the same as the documented Deployment flow does today.

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: rec
  labels:
    spo.x-k8s.io/enable-recording: "true"
---
apiVersion: security-profiles-operator.x-k8s.io/v1
kind: ProfileRecording
metadata:
  name: fixed
  namespace: rec
spec:
  kind: SeccompProfile
  recorder: Bpf
  mergeStrategy: Containers
  podSelector:
    matchLabels:
      app: fixed
---
apiVersion: v1
kind: Pod
metadata:
  name: fixed-pod            # fixed name, no generateName
  namespace: rec
  labels:
    app: fixed
    spo.x-k8s.io/enable-recording: "true"
spec:
  restartPolicy: Never
  containers:
    - name: probe
      image: ubuntu
      command: ["/bin/sh", "-ec", "id; sleep 10"]
```

Enable the BPF recorder (`spec.enricher.enableBpfRecorder: true` on the `spod`), apply the above, let the pod finish,
then `kubectl delete profilerecording fixed -n rec`.

What you see: one partial `SeccompProfile/fixed-probe` gets created (that's the merged target's name), on deletion the
merge controller loops on `MutateFn cannot mutate ...`, no merged profile shows up, and the recording keeps the
`spo.x-k8s.io/has-unmerged-profiles` finalizer. A Deployment with the same recording (so, `generateName` pods) merges
fine, which is what `test/tc_profilemerging_test.go` covers.

I've made a fix that works, but you probably know a better spot for it than I picked, so this just a suggestion:

- recorder: when it's a `Containers` recording and the pod has no `generateName`, fall back to something unique (the
  pod name works) so the partial can't land on the merged target's name. Leave `None` alone so a re-record still
  updates the same profile in place.
- merger: don't set `Namespace` on the cluster-scoped merged object in `mergedObjectMeta` - the recording namespace is
  already kept in the `spo.x-k8s.io/recording-namespace` label - so the update path stops erroring.

I built both against `v1.0.0` and ran it on CRI-O: the fixed-name case now merges through the create path, the
two-replica Deployment still merges, `mergeStrategy: None` single-pod is unchanged, and I don't get `MutateFn` errors
anymore. That same namespace guard also makes a second same-named pod fail under `mergeStrategy: None`. I can open a PR with the two changes plus a fixed-name case in `tc_profilemerging_test.go` if you're okay with the direction.

#### Environment:

- Cloud provider or hardware configuration: bare-metal K3s (`v1.34.7+k3s1`), CRI-O `1.34.7`, BPF recorder
- OS (e.g: `cat /etc/os-release`): `Ubuntu 24.04.4 LTS (Noble Numbat)`
- Kernel (e.g. `uname -a`): `Linux 6.8.0-134-generic #134-Ubuntu SMP PREEMPT_DYNAMIC x86_64`
- Others: Security Profiles Operator `v1.0.0`; also in `main` by inspection

idriss-eliguene:
What type of PR is this?

/kind bug

What this PR does / why we need it:

Fixes #3328.

With `mergeStrategy: Containers`, a fixed-name Pod can produce a partial
profile named:

    <recording>-<container>

which is also the name used for the final merged profile.

This causes the recording merger to find the existing partial when creating
the merged profile.

There is a second object-identity issue on that path: merged profile resources
are cluster-scoped, but the merged object metadata carries the recording
namespace in `metadata.namespace`.

When `CreateOrUpdate` retrieves the existing cluster-scoped partial, the
returned object has an empty namespace. The resulting object-key mismatch
causes the merge to fail.

This change keeps the existing naming behavior for generated Pods and for
`mergeStrategy: None`, while giving fixed-name Pods under
`mergeStrategy: Containers` a distinct partial-profile name:

    <recording>-<container>-<pod-name>

The final merged profile remains:

    <recording>-<container>

Merged profile objects also no longer set `metadata.namespace`. The recording
namespace remains available through the existing
`spo.x-k8s.io/recording-namespace` label.

Partial-profile discovery and cleanup are unchanged because both are
label-driven rather than dependent on the partial profile object name.

The syscall coverage behavior for merged seccomp profiles is unchanged.

Which issue(s) this PR fixes:

Fixes #3328

Special notes for your reviewer:

The regression test exercises the existing `collectBpfProfiles` path.

Against the parent implementation, the test compiles and fails behaviorally
because the fixed-name Pod produces:

    recording-container

With this change, the same path produces:

    recording-container-fixed-pod

while the final merged profile remains:

    recording-container

The cluster-scope regression test also verifies that merged profile metadata
does not carry a Kubernetes namespace while retaining the recording namespace
label.

A fixed-name Pod recreated with the same name can still reuse the same partial
profile identity. This change intentionally follows SPO's existing Pod-name
based naming semantics rather than introducing Pod UID identity as part of
this bug fix.

Does this PR introduce a user-facing change?

```release-note
Fix profile merging for fixed-name Pods using `mergeStrategy: Containers` by
preventing partial profiles from colliding with the final merged profile.
```

### DIFF
--- a/internal/pkg/daemon/profilerecorder/profilerecorder_test.go
+++ b/internal/pkg/daemon/profilerecorder/profilerecorder_test.go
@@ -35,6 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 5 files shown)


internal/pkg/daemon/profilerecorder/profilerecorder_test.go
```diff
@@ -62,6 +63,108 @@ func TestName(t *testing.T) {
...
+				key types.NamespacedName,
+				obj client.Object,
+			) error {
+				recordingResult, ok := obj.(*recordingapi.ProfileRecording)
+				require.True(t, ok)
+				require.Equal(t, client.ObjectKeyFromObject(recording), key)
+
+				recording.DeepCopyInto(recordingResult)
+
+				return nil
```
saschagrunert: Nit: the type assertion (`require.True(t, ok)`) means any future `ClientGet` call in the production flow that passes a different object type will panic here with a confusing message. Consider routing by type instead:

```go
mock.ClientGetCalls(func(_ context.Context, _ client.Client, key types.NamespacedName, obj client.Object) error {
    switch o := obj.(type) {
    case *recordingapi.ProfileRecording:
        require.Equal(t, client.ObjectKeyFromObject(recording), key)
        recording.DeepCopyInto(o)
        return nil
    default:
        t.Fatalf("unexpected ClientGet for %T", obj)
        return nil
    }
})
```

saschagrunert (commented):
Pre-existing: `profilePartial` (line 1306) returns `true, err` when the recording is NotFound. The caller propagates the error so the `true` is never acted on, but `return false, err` would be clearer. Fine to address separately.

internal/pkg/daemon/profilerecorder/profilerecorder_test.go
idriss-eliguene: Good point, thanks! Updated the mock to route explicitly by object type so unexpected ClientGet calls now fail with a clear message.

saschagrunert (approved):

PR #3384, opened 2026-08-25
fix(manager): avoid redundant profile status updates

### LINKED ISSUE #3383: Repeated no-op profile status updates cause resourceVersion churn and conflicts
/kind bug

#### What happened:

The workload annotator can repeatedly update a `SeccompProfile` status even when the set of active workloads has not changed. At the same time, the node status controller updates the same status subresource without retrying optimistic concurrency conflicts.

In a cluster where a profile is referenced by multiple Pods, regular Pod update events can cause the profile `resourceVersion` to increase continuously while the meaningful object content remains unchanged. The node status controller may then repeatedly fail with errors similar to:

```text
updating policy status: Operation cannot be fulfilled on seccompprofiles.security-profiles-operator.x-k8s.io "example-profile": the object has been modified; please apply your changes to the latest version and try again
```

For example, two snapshots can contain the same status data while only `resourceVersion` changes:

```text
before: resourceVersion=1000, activeWorkloads=["example/pod-a", "example/pod-b"]
after:  resourceVersion=1125, activeWorkloads=["example/pod-a", "example/pod-b"]
```

The relevant behavior is:

- `updatePodReferencesForSeccomp` always calls `Status().Update()` after listing linked Pods, even when `status.activeWorkloads` represents the same set.
- Pod list ordering is not guaranteed, so equivalent sets can also be written in different orders.
- The node status controller performs a full status update without `RetryOnConflict` and without checking whether the calculated status is already present.

- `status.activeWorkloads` should only be updated when the set of referenced Pods changes.
- Equivalent workload sets should have a stable order.
- The node status controller should skip no-op updates.
- Status update conflicts should be retried after reading the latest profile and its current `resourceVersion`.

1. Create a `SeccompProfile`.
2. Reference the profile from multiple Pods.
3. Generate regular Pod update events, for example through readiness changes or workload rollouts.
4. Observe the profile `resourceVersion` over time:

   ```bash
   kubectl get seccompprofile example-profile \
     -o jsonpath='{.metadata.resourceVersion}{"\n"}'
   ```

5. Compare two JSON snapshots after removing `metadata.resourceVersion` and `metadata.managedFields` and sorting `status.activeWorkloads`.
6. Observe the operator logs for optimistic concurrency conflicts from the `nodestatus` controller.

The repeated updates create unnecessary API server and etcd writes and can starve legitimate status updates under sustained Pod event activity.

A proposed fix should:

- sort and compare `activeWorkloads` before writing;
- skip unchanged workload and aggregate profile status updates;
- use `RetryOnConflict` with a fresh `Get` for every node status retry attempt.

#### Environment:

- Security Profiles Operator: v1.0.0; the affected update paths are also present on the current `main` branch
- Kubernetes: reproducible independently of a specific cloud provider
- Cloud provider or hardware configuration: not specific
- OS: not specific
- Kernel: not specific
- Others: profile referenced by multiple Pods with recurring Pod update events

frayme:
#### What type of PR is this?

/kind bug

This PR prevents redundant profile status writes and makes node status aggregation resilient to optimistic concurrency conflicts.

It:

- sorts `SeccompProfile.status.activeWorkloads` and `SelinuxProfile.status.activeWorkloads` before storing them;
- compares active workload sets and skips unchanged status updates for both profile types;
- rechecks each workload set after reading a fresh profile during conflict retries;
- exits workload status and finalizer reconciliation cleanly when a profile has been deleted;
- skips node status updates when the calculated status fields are unchanged;
- retries node status conflicts with `RetryOnConflict` and performs an uncached API read on every attempt;
- adds tests for order-independent workload comparison, no-op status detection, deleted profiles, and conflict retries with fresh reads.

Fixes #3383

#### Does this PR have test?

Yes:

```text
go test ./internal/pkg/manager/... -count=1
```

The workload annotator continues to maintain the active workload finalizer. The change only suppresses status writes when the workload set is semantically unchanged.

Freshness-critical profile reads use `Manager.GetAPIReader()`. This adds one direct API server `GET` per relevant reconciliation, plus additional reads on conflict retries, and avoids stale cached no-op decisions and stale `resourceVersion` reuse.

```release-note
Avoid redundant security profile status writes and retry node status update conflicts using the latest resource version.
```

### DIFF
--- a/internal/pkg/manager/nodestatus/nodestatus.go
+++ b/internal/pkg/manager/nodestatus/nodestatus.go
@@ -420,7 +420,26 @@ func (r *StatusReconciler) updateProfileStatus(
 }
 
 func profileStatusChanged(current, desired profilebaseapi.StatusBaseUser) bool {
-	return !reflect.DeepEqual(current, desired)
+	switch typedCurrent := current.(type) {
+	case *seccompprofileapi.SeccompProfile:
+		typedDesired, ok := desired.(*seccompprofileapi.SeccompProfile)
+
+		return !ok || !reflect.DeepEqual(typedCurrent.Status, typedDesired.Status)
+	case *selinuxprofileapi.SelinuxProfile:
+		typedDesired, ok := desired.(*selinuxprofileapi.SelinuxProfile)
+
+		return !ok || !reflect.DeepEqual(typedCurrent.Status, typedDesired.Status)
+	case *selinuxprofileapi.RawSelinuxProfile:
+		typedDesired, ok := desired.(*selinuxprofileapi.RawSelinuxProfile)
+
+		return !ok || !reflect.DeepEqual(typedCurrent.Status, typedDesired.Status)
+	case *apparmorapi.AppArmorProfile:
+		typedDesired, ok := desired.(*apparmorapi.AppArmorProfile)
+
+		return !ok || !reflect.DeepEqual(typedCurrent.Status, typedDesired.Status)
+	default:
+		return true
+	}
 }
 
 func daemonSetIsReady(ds *appsv1.DaemonSet) bool {
--- a/internal/pkg/manager/nodestatus/nodestatus.go
+++ b/internal/pkg/manager/nodestatus/nodestatus.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"reflect"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -32,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -348,6 +350,24 @@ func (r *StatusReconciler) reconcileStatus(
 	state secprofnodestatusapi.ProfileState,
 	l logr.Logger,
 ) error {
+	key := client.ObjectKeyFromObject(prof)
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		current := prof.DeepCopyToStatusBaseIf()
+		if err := r.client.Get(ctx, key, current); err != nil {
+			return client.IgnoreNotFound(err)
+		}
+
+		return r.updateProfileStatus(ctx, current, state, l)
+	})
+}
+
+func (r *StatusReconciler) updateProfileStatus(
+	ctx context.Context,
+	prof profilebaseapi.StatusBaseUser,
+	state secprofnodestatusapi.ProfileState,
+	l logr.Logger,
+) error {
 	pCopy := prof.DeepCopyToStatusBaseIf()
 
 	// We always set this status
@@ -385,6 +405,10 @@ func (r *StatusReconciler) reconcileStatus(
 		))
 	}
 
+	if !profileStatusChanged(prof, pCopy) {
+		return nil
+	}
+
 	l.V(config.VerboseLevel).Info("Updating status")
 
 	if updateErr := r.client.Status().Update(ctx, pCopy); updateErr != nil {
@@ -392,6 +416,10 @@ func (r *StatusReconciler) reconcileStatus(
 	}
 
 	return nil
+}
+
+func profileStatusChanged(current, desired profilebaseapi.StatusBaseUser) bool {
+	return !reflect.DeepEqual(current, desired)
 }
 
 func daemonSetIsReady(ds *appsv1.DaemonSet) bool {
--- a/internal/pkg/manager/nodestatus/nodestatus.go
+++ b/internal/pkg/manager/nodestatus/nodestatus.go
@@ -66,6 +66,7 @@ func NewController() controller.Controller {
 // A StatusReconciler monitors node changes and updates the profile status.
 type StatusReconciler struct {
 	client client.Client
+	reader client.Reader
 	log    logr.Logger
 	record record.EventRecorder
 }
@@ -354,7 +355,7 @@ func (r *StatusReconciler) reconcileStatus(
 
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		current := prof.DeepCopyToStatusBaseIf()
-		if err := r.client.Get(ctx, key, current); err != nil {
+		if err := r.reader.Get(ctx, key, current); err != nil {
 			return client.IgnoreNotFound(err)
 		}
 
--- a/internal/pkg/manager/nodestatus/nodestatus_test.go
+++ b/internal/pkg/manager/nodestatus/nodestatus_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2026 The Kubernetes Authors.
+Copyright The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/internal/pkg/manager/nodestatus/nodestatus_test.go
+++ b/internal/pkg/manager/nodestatus/nodestatus_test.go
@@ -1,0 +1,114 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodestatus
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/go-logr/logr"
+	"github.com/stretchr/testify/require"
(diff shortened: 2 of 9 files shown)


internal/pkg/manager/workloadannotator/workloadannotator.go
```diff
@@ -220,6 +226,19 @@ func (r *PodReconciler) updatePodReferencesForSeccomp(ctx context.Context, sp *s
...
+	desiredCopy := slices.Clone(desired)
+	slices.Sort(currentCopy)
+	slices.Sort(desiredCopy)
+
+	return slices.Equal(currentCopy, desiredCopy)
+}
+
 // updatePodReferencesForSelinux updates a SelinuxProfile with the identifiers of pods using it and ensures
 // it has a finalizer indicating it is in use to prevent it from being deleted.
 func (r *PodReconciler) updatePodReferencesForSelinux(ctx context.Context, se *selinuxprofileapi.SelinuxProfile) error {
```
saschagrunert: `updatePodReferencesForSelinux` has the same write-churn pattern: no sorting, no `sameActiveWorkloads` check, cached reads on retry, update-then-read-on-failure. Should get the same treatment for consistency, either in this PR or as a follow-up.

internal/pkg/manager/workloadannotator/workloadannotator.go
```diff
@@ -185,17 +187,21 @@ func (r *PodReconciler) updatePodReferencesForSeccomp(ctx context.Context, sp *s
 		pod := linkedPods.Items[i]
 		podList[i] = pod.Namespace + "/" + pod.Name
 	}
+	slices.Sort(podList)
 
 	if err := util.Retry(func() error {
-		sp.Status.ActiveWorkloads = podList
+		if err := r.reader.Get(ctx, util.NamespacedName(sp.GetName(), sp.GetNamespace()), sp); err != nil {
+			return fmt.Errorf("retrieving profile: %w", err)
```
saschagrunert, line 255: If the profile gets deleted, this wraps the NotFound error, `util.IsNotFoundOrConflict` matches, and all 5 retry attempts are wasted on a profile that won't come back. In `nodestatus.go` (line 325), the same scenario uses `client.IgnoreNotFound` to exit cleanly. Consider the same approach here.

internal/pkg/manager/workloadannotator/workloadannotator.go
```diff
@@ -220,6 +226,19 @@ func (r *PodReconciler) updatePodReferencesForSeccomp(ctx context.Context, sp *s
...
 
+func sameActiveWorkloads(current, desired []string) bool {
+	if len(current) != len(desired) {
+		return false
+	}
+
+	currentCopy := slices.Clone(current)
+	desiredCopy := slices.Clone(desired)
+	slices.Sort(currentCopy)
+	slices.Sort(desiredCopy)
```
saschagrunert, line 307: nit: The caller already sorts `podList` (line 190) before passing it as `desired`. This sort and its clone are redundant for the current call site. Fine to keep for defensive safety, but worth noting.

internal/pkg/manager/nodestatus/nodestatus.go
```diff
@@ -354,6 +379,10 @@ func (r *StatusReconciler) reconcileStatus(
 	return nil
 }
 
+func profileStatusChanged(current, desired profilebaseapi.StatusBaseUser) bool {
+	return !reflect.DeepEqual(current, desired)
```
saschagrunert: This compares the full object (metadata, spec, status), not just status. It works because `pCopy` is derived from `prof` and only status fields are modified, but the function name suggests a status-only comparison. If a future change touches non-status fields between the deep copy and this check, it would silently defeat the no-op optimization. Minor concern since the code path is self-contained today.

saschagrunert (commented):
The core idea is good: skip no-op writes and retry conflicts with fresh reads. Main concern is the SELinux gap (see inline). The rest is minor.

internal/pkg/manager/workloadannotator/workloadannotator.go
frayme: Addressed in ce8b0b78. The SELinux path now sorts activeWorkloads, performs the no-op check after an uncached read, and retries status updates with a fresh resourceVersion. Added regression coverage for stale reads and order-independent no-op updates.

internal/pkg/manager/workloadannotator/workloadannotator.go
frayme: Addressed in ce8b0b78. A NotFound response now ends workload status reconciliation cleanly before finalizer processing, avoiding retries for a profile that has been deleted. Added regression coverage for this path.

internal/pkg/manager/workloadannotator/workloadannotator.go
frayme: Kept intentionally for defensive safety: sameActiveWorkloads remains order-independent regardless of caller normalization. The caller-side sort still provides deterministic stored order.

internal/pkg/manager/nodestatus/nodestatus.go
frayme: Addressed in ce8b0b78. profileStatusChanged now compares the concrete status structs for SeccompProfile, SelinuxProfile, RawSelinuxProfile, and AppArmorProfile, so metadata/spec-only differences do not trigger status writes. The test covers all four supported owner types.

frayme:
Addressed the review feedback in ce8b0b78:

- applied the no-op/fresh-read/retry behavior to the SELinux workload path;
- stopped retrying deleted profiles and skipped subsequent finalizer work;
- limited profile change detection to concrete status structs for all supported profile types;
- fixed the wsl_v5 failures.

Validation:

```text
go test ./internal/pkg/manager/... -count=1
go test -race ./internal/pkg/manager/nodestatus ./internal/pkg/manager/workloadannotator -count=1
go vet ./internal/pkg/manager/nodestatus ./internal/pkg/manager/workloadannotator
golangci-lint (changed packages): 0 issues
```

saschagrunert (approved):

frayme:
Follow-up in 2c48f70f closes the remaining deletion race between status reconciliation and finalizer processing. `NotFound` is now treated as a successful exit for Seccomp and SELinux finalizer add/remove retries; conflicts and unrelated errors still propagate normally.

Added regression coverage for deletion before both finalizer addition and removal across both profile types. The tests were verified RED before the change and GREEN afterward. Manager tests, race tests, vet, and changed-package golangci-lint all pass.

internal/pkg/manager/nodestatus/nodestatus_test.go
```diff
@@ -0,0 +1,167 @@
+/*
+Copyright 2026 The Kubernetes Authors.
```
saschagrunert: I think we need to remove this:
```suggestion
Copyright The Kubernetes Authors.
```

vendor/k8s.io/client-go/util/retry/util.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2016 The Kubernetes Authors.
```
saschagrunert, line 2:
```suggestion
Copyright The Kubernetes Authors.
```

frayme:
Ty, i accepted this suggestion

saschagrunert:
https://github.com/kubernetes-sigs/security-profiles-operator/pull/3389 will merge sooner than later, which should make the CI green again

saschagrunert:
Please rebase to fix the CI.

frayme:
Rebased onto current `main` (`4f5ed479`) as requested. Conflicts in the workload annotator path were resolved while preserving the no-op, uncached-read, SELinux parity, and deletion/finalizer behavior.

Post-rebase validation:

```text
go test ./internal/pkg/manager/... -count=1
go test -race ./internal/pkg/manager/nodestatus ./internal/pkg/manager/workloadannotator -count=1
go vet ./internal/pkg/manager/nodestatus ./internal/pkg/manager/workloadannotator
make verify-boilerplate
golangci-lint (changed packages): 0 issues
```

The full local lint target remains blocked only by missing native `bpf/bpf.h` headers; the changed packages pass with the repository’s updated linter.

vendor/k8s.io/client-go/util/retry/util.go
frayme: The accepted yearless suggestion conflicts with the repository vendor verification: `go mod vendor` restores the upstream client-go header (`Copyright 2016 The Kubernetes Authors.`), and `verify-go-mod` fails when the vendored file differs. Restored the generated upstream header in b8fad12b; `verify-boilerplate` excludes vendor and still passes.

saschagrunert:
@frayme we switched to go 1.27 for prow testing means that gofmt behaves a bit different now. Please reformat using 1.27.

frayme:
`main` advanced again to `f2ae638e`, which contains the fixes for the unrelated `gci` failures seen in the previous verify run. Rebased onto that commit and force-pushed the branch.

Post-rebase checks pass: manager tests, targeted race tests, vet, `verify-go-mod`, `verify-boilerplate`, changed-package golangci-lint, range-diff, and diff checks. New CI is running on `8d84c83d`.

frayme:
The `e2e-flatcar` failure is an infrastructure flake before any tests run: the `Boot Virtual Machine` step failed and all test steps were skipped. The same step also failed on current `main` run `32989867192`. I will rerun the failed job after the workflow finishes.

frayme:
All required Prow checks and all other GitHub Actions jobs have passed. The only remaining failure is the `e2e-flatcar` infrastructure failure in `Boot Virtual Machine`; no E2E tests ran.

I attempted to rerun the failed job, but GitHub requires repository admin permissions (`run 32992294355 cannot be rerun; Must have admin rights to Repository`). Could a maintainer please rerun that failed job?

saschagrunert:
CI failure is unrelated and need a separate fix. Thank you!

PR #3400, opened 2026-09-01
helm: Add restrictToNamespace value

### LINKED ISSUE #3395: Add ability to restrict to namespace with helm deployment

Within the docs it [explains how you can restrict the namespace searched for seccompProfile's in OLM and manifest deployment](https://github.com/kubernetes-sigs/security-profiles-operator/blob/main/installation-usage.md#restricting-to-a-single-namespace).

But after looking at the helm values and templates, there is no way to add this ability to the helm (without altering the resources after deployment)

#### Why is this needed:

As explained in the docs this feature is beneficial:

> This is advantageous because it allows for tightening the RBAC permissions required by the operator's ServiceAccount

By allowing restricting it to a specific namespace, privileged pod creations can be tightly controlled and monitored, allowing for the distribution of privileges of other namespaces without the ability to create seccomp profiles.

#### User story covered

Security team should be able to limit access to creation of privileged pods.

saschagrunert:
/kind feature

Adds a `restrictToNamespace` Helm value that sets the `RESTRICT_TO_NAMESPACE` environment variable on the manager deployment. This allows Helm users to restrict the operator to watch profiles in a single namespace, which was previously only possible via the upstream manifests or OLM.

Fixes #3395

#### Does this PR have test?
N/A (Helm template change, verified with `helm template`)

The operator already supports `RESTRICT_TO_NAMESPACE` and propagates it from the manager deployment to the SPOd DaemonSet automatically (`setup.go:231-236`), so only the Helm chart needs to be updated.

```release-note
Add `restrictToNamespace` Helm chart value to restrict the operator to a single namespace.
```

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -32,6 +32,7 @@ please refer to [Installation Guide](https://github.com/kubernetes-sigs/security
 | nodeSelector                                | object | `{}` | `specify on which node to deploy the workload` |
 | podSecurityContext                          | object | `` | `pod security contexts`                        |
 | replicaCount                                | int | `3` | `the number of replicas of the pods`           |
+| restrictToNamespace                         | string | `""` | `restrict the operator to a single namespace`  |
 | resources.limits.memory                     | string | `"128Mi"` | `memory limits for the pod`                    |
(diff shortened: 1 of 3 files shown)


saschagrunert:

ccojocar (approved):

PR #3401, opened 2026-09-01
Add configurable webhook tolerations

### LINKED ISSUE #3181: Configurable webhook tolerations
Webhook Deployment should not automatically inherit tolerations from SPOD daemon configuration.

Currently, configuring:

```
daemon:
  tolerations:
  - operator: Exists
```

also applies the same tolerations to the generated security-profiles-operator-webhook Deployment.

It would be useful to have separate webhook toleration configuration from SPOD daemon.

related code is [here](https://github.com/kubernetes-sigs/security-profiles-operator/blob/22c9486361e1f4affaf1fc4c8adb1057b3191422/internal/pkg/manager/spod/spod_controller.go#L879)

#### Why is this needed:
SPOD is node-local functionality and is often intentionally scheduled onto all nodes using broad tolerations.

The webhook Deployment is not node-local and does not need the same scheduling behavior. Inheriting daemon tolerations causes the webhook to tolerate all taints and be scheduled onto dedicated/specialized nodes unintentionally.

#### User story covered

Operators should be able to configure SPOD DaemonSet scheduling independently from webhook scheduling to support tainted node pools and workload isolation.

saschagrunert:
/kind feature

Adds a `tolerations` field to `SPODWebhookConfig` so that webhook scheduling can be configured independently from the SPOD daemon. Previously, the webhook Deployment always inherited the daemon's tolerations, which caused it to be scheduled onto tainted/dedicated nodes unintentionally when broad daemon tolerations (e.g. `operator: Exists`) were configured. When webhook tolerations are not set, the webhook falls back to the daemon's tolerations, preserving backward compatibility.

Fixes #3181

#### Does this PR have test?
N/A

The full deployment manifests (`deploy/operator.yaml`, etc.) need a `make deployments` regeneration to pick up the CRD changes. The base CRD and Helm CRD are already updated.

```release-note
Add webhook.tolerations field to SPOD config for independent webhook scheduling.
```


saschagrunert:

ccojocar (approved):

PR #3421, opened 2026-09-09
Publish runtime-spec seccomp profiles as KEP-6061 artifacts

saschagrunert:
#### What type of PR is this?

/kind feature

KEP-6061 (kubernetes/enhancements#6062) defines how container runtimes consume seccomp profiles from OCI artifacts: exactly one layer holding the profile as OCI runtime-spec JSON, identified by the media type `application/vnd.cncf.seccomp-profile.config.v1+json`. `spoc push` currently packs everything with the generic `application/vnd.unknown.config.v1+json`, which runtimes will reject.

`spoc push` now detects a raw runtime-spec seccomp JSON input (for example the output of `spoc convert`) and produces the runtime format: one raw JSON layer, not platform qualified because such artifacts hold exactly one platform independent profile, with the media type on the manifest config as well as on `artifactType`. The config descriptor has to be set explicitly, ORAS otherwise defaults to the empty OCI config and leaves the media type in `artifactType` only.

Input which is neither a profile CRD nor a runtime-spec profile is rejected, so that no artifact gets published which no consumer understands, and content container runtimes reject is warned about: listener fields, `SCMP_ACT_NOTIFY`, and the 1 MiB size and 128 syscall entry limits they apply by default.

Profile CRD inputs keep the existing format used for `oci://` base profiles. `spoc pull` and `oci://` references accept both. Runtime format artifacts are recognized by their media type and read from the single layer the format requires, so layer names and annotations do not matter, which KEP-6061 mandates nothing about. Other artifacts keep using the layer names push produces, with a fallback to the single layer of the artifact if it is not bound to another platform. A runtime format artifact is exposed as a `SeccompProfile` whose spec drops fields the CRD cannot express.

`spoc convert` now writes a portable raw profile. `state` and `baseProfileName` are gone, they are CRD only fields which made its output fail the strict runtime-spec decode, and so are the listener fields, which tie the profile to the seccomp agent of this operator and are rejected in KEP-6061 artifacts. Dropping a base profile, whose syscalls are not part of the raw profile, and profiles which need a notifier are logged.

Relates to kubernetes/enhancements#6061

#### Does this PR have test?

Unit tests for push format detection and rejection, the packed manifest, layer names and media types, the pull resolution by media type and its fallbacks, the conversion and the error cases.

None

```release-note
`spoc push` publishes raw OCI runtime-spec seccomp profiles as KEP-6061 artifacts (single raw JSON layer, media type `application/vnd.cncf.seccomp-profile.config.v1+json` on the manifest config and `artifactType`), and `spoc pull` as well as `oci://` base profiles accept that format. `spoc convert` no longer writes the CRD only fields `state` and `baseProfileName` nor the operator specific listener fields into the raw profile.
```

https://claude.ai/code/session_016cmWPv2zETBDBwk3hUv9zv

### DIFF
--- a/cli.md
+++ b/cli.md
@@ -6,6 +6,7 @@
   - [Run commands with seccomp profiles](#run-commands-with-seccomp-profiles)
   - [Pull security profiles from OCI registries](#pull-security-profiles-from-oci-registries)
   - [Push security profiles to OCI registries](#push-security-profiles-to-oci-registries)
+  - [Pushing profiles for container runtimes](#pushing-profiles-for-container-runtimes)
   - [Using multiple platforms](#using-multiple-platforms)
 <!-- /toc -->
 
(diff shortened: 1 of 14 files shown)


saschagrunert:

Issue #3317, opened 2026-07-13
SPO uses broader privileges and hostPath mounts than necessary


TomWis97:
#### What happened:
There are various ways in which the Security Profiles Operator seems to have more privileges than strictly necessary.

##### securityContext for security-profiles-operator and security-profiles-operator-webhook
For both the `security-profiles-operator` and the `security-profiles-operator-webhook` pods, more restructive securityContext should be configured. Currently, only `"allowPrivilegeEscalation": false` and `"readOnlyRootFilesystem": true` are set.

##### DaemonSet is mounting more than required
In this example, SELinux support was disabled and AppArmor was enabled. I've found that the `spod` pod mounts the following hostPaths:
```
$ kubectl get pods -n security-profiles-operator spod-nggb5 -ojson | jq '.spec.volumes[] | select(has("hostPath")) | .hostPath.path'
"/var/lib"
"/var/lib/security-profiles-operator"
"/sys/fs/selinux"
"/etc/selinux"
"/var/lib/selinux"
"/tmp/security-profiles-operator-recordings"
"/var/log/audit"
"/var/log"
"/sys/kernel/debug"
"/sys/kernel/security"
"/sys/kernel/tracing"
"/etc/os-release"
"/"
```

Especially the SELinux I cannot understand, as SELinux support was disabled. Additionally, it's unclear to me why a lot of the other directories get mounted, as they're not mentioned in the `securityModel.md` file.

##### securityContext for security-profiles-operator and security-profiles-operator-webhook
Presumably the following parameters should also be set for at least both the `security-profiles-operator` and `security-profiles-operator-webhook` pods and containers:
```
spec:
  securityContext:
    seccompProfile:
      type: RuntimeDefault
  containers:
    - securityContext:
        allowPrivilegeEscalation: false
        privileged: false
        readOnlyRootFilesystem: true
        runAsNonRoot: true
        runAsUser: 1000
        runAsGroup: 1000
        fsGroup: 1000
        capabilities:
          drop:
            - ALL
```

##### DaemonSet is mounting more than required
This issue is multifaceted: limiting the amount of hostPath mounts, more clearly describing which mounts are required and (I'd expect) mounting AppArmor related mounts when AppAmor is enabled in the Helm Chart.

At least when SELinux support is disabled, I'd not expect SELinux-related directories to be mounted by the `spod` pod. Also, I have some doubts with some mounts, for example the host's `/`.

[...]

PR #3422, opened 2026-09-09
Harden operator and webhook securityContext: runAsNonRoot, drop ALL

KR-Ravindra:
#### What type of PR is this?

/kind bug

## Problem

The `security-profiles-operator` manager container and the admission `webhook` container run with a minimal `securityContext` that only sets `readOnlyRootFilesystem: true` and `allowPrivilegeEscalation: false`. They do not declare `runAsNonRoot: true` or drop Linux capabilities, even though neither component needs any capability and both images already switch to `USER 65535:65535`. Pod Security admission in the `restricted` profile and common policy engines flag this, and it leaves the containers with the default capability set for no reason.

## Root cause

The hardening fields were simply never added at the places where these two Deployments are defined:

- `deploy/kustomize-deployment/manager_deployment.yaml:27-29` (manager container, source for `deploy/operator.yaml`, `deploy/namespace-operator.yaml`, `deploy/openshift-dev.yaml`, `deploy/openshift-downstream.yaml`, `deploy/webhook-operator.yaml`)
- `deploy/helm/templates/deployment.yaml:47-49` (manager container in the Helm chart)
- `internal/pkg/manager/spod/bindata/webhook.go:681-684` (webhook container in the Deployment the spod controller creates)
- `deploy/overlays/webhook/webhook_deployment.yaml:31-33` (webhook container in the standalone webhook overlay)
- `bundle/manifests/security-profiles-operator.clusterserviceversion.yaml:731-733` (manager container in the OLM bundle)

## Fix

Add `runAsNonRoot: true` and `capabilities.drop: ["ALL"]` to the container `securityContext` in all of the above, and regenerate the derived manifests with `kustomize build` (same output as `make deployments`). The Dockerfile and Dockerfile.ubi both end with `USER 65535:65535`, so `runAsNonRoot: true` is satisfied without changing the image. The pod-level `seccompProfile: RuntimeDefault` was already in place in the kustomize, OLM and webhook manifests and is unchanged; the Helm chart leaves the pod-level profile to `.Values.podSecurityContext`, which is also unchanged here.

The remaining items from the issue (making the SELinux/BPF `hostPath` volumes on the spod DaemonSet conditional and narrowing the host-root volume) touch the privileged daemon containers in `internal/pkg/manager/spod/bindata/spod.go` and are intentionally left for a follow-up.

Part of https://github.com/kubernetes-sigs/security-profiles-operator/issues/3317 (the hostPath-volume items remain open, so this does not close it).

## How tested

New unit test `TestWebhook_DeploymentSecurityContext` in `internal/pkg/manager/spod/bindata/webhook_test.go` builds the webhook Deployment via `GetWebhook` and asserts `RunAsNonRoot == true`, `Capabilities.Drop == ["ALL"]`, `AllowPrivilegeEscalation == false`, `ReadOnlyRootFilesystem == true`, and pod `SeccompProfile == RuntimeDefault`.

Before the fix:

```
$ go test -count=1 -run TestWebhook_DeploymentSecurityContext ./internal/pkg/manager/spod/bindata/
--- FAIL: TestWebhook_DeploymentSecurityContext (0.00s)
    webhook_test.go:263:
        	Error Trace:	.../internal/pkg/manager/spod/bindata/webhook_test.go:263
        	Error:      	Expected value not to be nil.
        	Test:       	TestWebhook_DeploymentSecurityContext
FAIL
FAIL	sigs.k8s.io/security-profiles-operator/internal/pkg/manager/spod/bindata	0.024s
```

After the fix:

```
$ go test -count=1 ./internal/pkg/manager/spod/...
ok  	sigs.k8s.io/security-profiles-operator/internal/pkg/manager/spod	0.022s
ok  	sigs.k8s.io/security-profiles-operator/internal/pkg/manager/spod/bindata	0.016s
```

`gofmt -l` and `go vet ./internal/pkg/manager/spod/...` are clean; `go build ./internal/pkg/manager/spod/...` succeeds. The generated `deploy/*.yaml` files were produced with kustomize v5.8.1 (the pinned `KUSTOMIZE_VERSION`) and differ from their sources only by the four added lines per container.

The OLM bundle CSV was edited by hand to mirror the kustomize output because `make bundle` needs operator-sdk and a cgo toolchain that were not available in my environment; `make verify-bundle` in CI will confirm it matches.

## Links

- Issue: https://github.com/kubernetes-sigs/security-profiles-operator/issues/3317
- Pod Security Standards (restricted profile): https://kubernetes.io/docs/concepts/security/pod-security-standards/

```release-note
The operator and webhook containers now set `runAsNonRoot: true` and drop all Linux capabilities in their `securityContext`.
```

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

### DIFF
--- a/bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
@@ -730,7 +730,11 @@ spec:
                     memory: 50Mi
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
                   readOnlyRootFilesystem: true
+                  runAsNonRoot: true
               securityContext:
                 seccompProfile:
                   type: RuntimeDefault
--- a/internal/pkg/manager/spod/bindata/webhook_test.go
+++ b/internal/pkg/manager/spod/bindata/webhook_test.go
@@ -251,7 +251,11 @@ func TestWebhook_DeploymentSecurityContext(t *testing.T) {
 	podSpec := sut.deployment.Spec.Template.Spec
 	require.NotNil(t, podSpec.SecurityContext)
 	require.NotNil(t, podSpec.SecurityContext.SeccompProfile)
-	assert.Equal(t, corev1.SeccompProfileTypeRuntimeDefault, podSpec.SecurityContext.SeccompProfile.Type)
+	assert.Equal(
+		t,
+		corev1.SeccompProfileTypeRuntimeDefault,
+		podSpec.SecurityContext.SeccompProfile.Type,
+	)
 
 	require.Len(t, podSpec.Containers, 1)
 	sc := podSpec.Containers[0].SecurityContext
--- a/internal/pkg/manager/spod/bindata/webhook_test.go
+++ b/internal/pkg/manager/spod/bindata/webhook_test.go
@@ -268,4 +268,5 @@ func TestWebhook_DeploymentSecurityContext(t *testing.T) {
 	assert.True(t, *sc.RunAsNonRoot)
 	require.NotNil(t, sc.Capabilities)
 	assert.Equal(t, []corev1.Capability{"ALL"}, sc.Capabilities.Drop)
+	assert.Empty(t, sc.Capabilities.Add)
 }
--- a/internal/pkg/manager/spod/bindata/webhook_test.go
+++ b/internal/pkg/manager/spod/bindata/webhook_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	admissionregv1 "k8s.io/api/admissionregistration/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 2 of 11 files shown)


KR-Ravindra:
The `e2e-flatcar` and `e2e-seccomp-profile` failures on 0ac629f both died in the "Boot Virtual Machine" step with "The hosted runner lost communication with the server"; no test ran. `macos` (unit + go lint) is green after the golines fix. A rerun of the two e2e jobs should be enough.

saschagrunert:
> The `e2e-flatcar` and `e2e-seccomp-profile` failures on [0ac629f](https://github.com/kubernetes-sigs/security-profiles-operator/commit/0ac629ffc32f2a71e019061675e5f9eba4beb411) both died in the "Boot Virtual Machine" step with "The hosted runner lost communication with the server"; no test ran. `macos` (unit + go lint) is green after the golines fix. A rerun of the two e2e jobs should be enough.

Thank you for the PR! That's an infra issue which I found no solution for yet.

internal/pkg/manager/spod/bindata/webhook_test.go
```diff
@@ -238,3 +239,33 @@ func TestWebhook_getWebhookConfig(t *testing.T) {
...
+	sc := podSpec.Containers[0].SecurityContext
+	require.NotNil(t, sc)
+	require.NotNil(t, sc.AllowPrivilegeEscalation)
+	assert.False(t, *sc.AllowPrivilegeEscalation)
+	require.NotNil(t, sc.ReadOnlyRootFilesystem)
+	assert.True(t, *sc.ReadOnlyRootFilesystem)
+	require.NotNil(t, sc.RunAsNonRoot)
+	assert.True(t, *sc.RunAsNonRoot)
+	require.NotNil(t, sc.Capabilities)
+	assert.Equal(t, []corev1.Capability{"ALL"}, sc.Capabilities.Drop)
```
saschagrunert, line 270: Nit: consider also asserting that no capabilities are added back:

```suggestion
	assert.Equal(t, []corev1.Capability{"ALL"}, sc.Capabilities.Drop)
	assert.Empty(t, sc.Capabilities.Add)
```

Without this, a future change that accidentally sets `Add: ["NET_BIND_SERVICE"]` (or similar) would still pass the test.

bundle/manifests/security-profiles-operator.clusterserviceversion.yaml
```diff
@@ -730,7 +730,11 @@ spec:
                     memory: 50Mi
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
                   readOnlyRootFilesystem: true
+                  runAsNonRoot: true
```
saschagrunert, line 737: The PR description mentions this file was hand-edited because `make bundle` was not available locally. Confirmed that CI will catch any drift: `.github/workflows/olm_tests.yaml` runs `make verify-bundle` on every PR (line 41), so this will be validated.

KR-Ravindra:
@saschagrunert done in 9d131e2: added `assert.Empty(t, sc.Capabilities.Add)` right after the `Drop` assertion in `webhook_test.go`; `go test ./internal/pkg/manager/spod/bindata/` passes. Thanks for confirming that `make verify-bundle` covers the hand-edited CSV.

Round 1 self-review. All checks were green on 0ac629f (prow build/verify/unit/test-e2e, the GitHub e2e matrix, and the OLM `tests` job); the new commit only adds that one test assertion. Re-checked the diff: the same three fields (`runAsNonRoot`, `capabilities.drop: [ALL]`, existing `readOnlyRootFilesystem`/`allowPrivilegeEscalation`) land in the five sources (kustomize manager deployment, helm deployment, `webhook.go`, webhook overlay, CSV) and the regenerated kustomize outputs, and both Dockerfiles end with `USER 65535:65535`, so `runAsNonRoot` matches the images. Marking ready for review.

saschagrunert (approved):
Thanks!

PR #3436, opened 2026-09-13
Only render SPOd hostPath volumes of enabled features

KR-Ravindra:
#### What type of PR is this?

/kind bug

#### Problem

The base SPOd DaemonSet in `internal/pkg/manager/spod/bindata/spod.go` declares the hostPath volumes of every optional feature unconditionally: `host-fsselinux-volume`, `host-etcselinux-volume`, `host-varlibselinux-volume` (SELinux), `host-auditlog-volume`, `host-syslog-volume` (log/JSON enricher) and `sys-kernel-debug-volume`, `sys-kernel-security-volume`, `sys-kernel-tracing-volume`, `host-etc-osrelease-volume` (bpf recorder / JSON enricher).

`getConfiguredSPOd` in `internal/pkg/manager/spod/spod_controller.go` deep-copies that template and only conditionally adds the feature *containers*; `templateSpec.Volumes` is only ever appended to. So a SPOD with SELinux, the enrichers and the bpf recorder disabled still mounts `/sys/fs/selinux`, `/etc/selinux`, `/var/lib/selinux`, `/var/log/audit`, `/var/log`, `/sys/kernel/debug`, `/sys/kernel/security`, `/sys/kernel/tracing` and `/etc/os-release` from every node, without any container using them. This is one of the remaining items from the issue after #3422 (securityContext hardening).

#### Fix

After all containers are rendered, `getConfiguredSPOd` now calls `pruneUnmountedVolumes`, which drops every volume of the pod template that no init container or container mounts. Because the feature containers are already conditional, this makes the SELinux, enricher and bpf recorder volumes follow their features exactly, including the mixed case of the JSON enricher (audit logs plus `/sys/kernel/debug` and `/sys/kernel/tracing`, but not `/sys/kernel/security` or `/etc/os-release`), and it does not need a hand maintained volume-to-feature list that can drift when volumes are added. Volumes mounted by always-present containers (non-root-enabler, daemon) are unaffected. Re-enabling a feature brings its volumes back with its container.

`spodNeedsUpdate` also compares the volume count of the configured and live pod templates next to the existing container counts. `apiequality.Semantic.DeepDerivative` accepts a shorter configured slice as a prefix of the live one, so dropping the trailing volume (the JSON enricher log volume, which `updateJsonEnricherSpec` appends last) would otherwise not roll out to existing DaemonSets.

Part of #3317 (does not close it)

#### Does this PR have test?

Yes, unit tests in `internal/pkg/manager/spod/spod_controller_test.go`:

- `Test_getConfiguredSPOdVolumesFollowFeatures`: with AppArmor on and SELinux/enrichers/bpf off, none of the SELinux, audit/syslog or `/sys/kernel/*` volumes are rendered; enabling SELinux, the log enricher or the bpf recorder brings back exactly its volumes; disabling drops them again; every rendered mount and volume device is still backed by a volume.
- `Test_getConfiguredSPOdJsonEnricherVolumes`: the JSON enricher keeps the audit log, debug and tracing volumes but not the security or os-release ones, and the operator ConfigMap's log volume is rendered only while the enricher is enabled.
- `Test_spodNeedsUpdateVolumeCount`: removing the trailing volume, removing a middle one or adding one requires an update; without the volume count check the trailing case fails.

#### How tested

Before the fix (`go test -run 'Test_getConfiguredSPOd(VolumesFollowFeatures|JsonEnricherVolumes)' ./internal/pkg/manager/spod/`):

```
--- FAIL: Test_getConfiguredSPOdJsonEnricherVolumes (0.07s)
    Messages:   	volume sys-kernel-security-volume   expected: false  actual: true
--- FAIL: Test_getConfiguredSPOdVolumesFollowFeatures (0.00s)
    Messages:   	volume host-fsselinux-volume        expected: false  actual: true
FAIL	sigs.k8s.io/security-profiles-operator/internal/pkg/manager/spod	0.090s
```

After the fix:

```
ok  	sigs.k8s.io/security-profiles-operator/internal/pkg/manager/spod	0.097s
ok  	sigs.k8s.io/security-profiles-operator/internal/pkg/manager/spod/bindata	0.013s
```

`gofmt` and `go vet` on `./internal/pkg/manager/spod/...` report no issues.

This change was prepared with an AI agent operated by KR-Ravindra, who reviewed and tested it.

```release-note
The spod DaemonSet no longer mounts the SELinux (`/sys/fs/selinux`, `/etc/selinux`, `/var/lib/selinux`), audit log (`/var/log/audit`, `/var/log`) and `/sys/kernel/{debug,security,tracing}` / `/etc/os-release` host paths when the feature that uses them (SELinux, log enricher, JSON enricher, bpf recorder) is disabled.
```

Links: https://github.com/kubernetes-sigs/security-profiles-operator/issues/3317


KR-Ravindra:
Round 1 self-review.

1. e2e-ubuntu failed in \`namespaced: JSON Enricher\` (and e2e-fedora is still running). This change alters exactly that container's volume set, so before undrafting: re-run, and if it reproduces, diff the rendered DaemonSet volumes with and without the prune to rule out an interaction with \`json-enricher-log-output-volume\` added in \`setup.go\`.
2. Checked against main: the daemon's \`host-varlibselinux-volume\` mount is only added under \`enableSelinux\`, so no mount can reference a pruned volume; the new test asserts this for every rendered container.
3. \`deploy/\` and the Helm chart do not carry the DaemonSet, so no generated manifests need regenerating.
4. \`security-model.md\` still lists the old mounts; leaving that for a follow-up under #3317.

Staying draft until 1 is resolved.

KR-Ravindra:
Round 2: marking this ready. The `e2e-ubuntu` failure I flagged in round 1 was in `namespaced: JSON Enricher`, which is the container whose volume set this change alters, so I held the PR until it was re-run. It now passes, as do e2e-fedora, e2e-flatcar, e2e-apparmor-profile, e2e-seccomp-profile and e2e-spoc. No code changed since round 1.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -888,9 +888,42 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
...
+// features, like /sys/fs/selinux, /var/log/audit or /sys/kernel/debug, stay
+// part of the DaemonSet on every node.
+func pruneUnmountedVolumes(templateSpec *corev1.PodSpec) {
+	mounted := map[string]bool{}
+
+	for _, containers := range [][]corev1.Container{
+		templateSpec.InitContainers, templateSpec.Containers,
+	} {
+		for i := range containers {
+			for _, mount := range containers[i].VolumeMounts {
```
saschagrunert, line 911: Kubelet counts both `VolumeMounts` and `VolumeDevices` as uses of a volume (see `GetPodVolumeNames`). Nothing here uses block devices today, but could we also collect `containers[i].VolumeDevices` into `mounted`? Then a future device volume wouldn't get dropped silently.

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -224,6 +226,128 @@ func Test_getConfiguredSPOdEnricherArgsAreRevertible(t *testing.T) {
...
+// keeps exactly the host paths it mounts: the audit logs and the tracing
+// filesystems, but neither /sys/kernel/security nor /etc/os-release, which
+// only the bpf recorder needs.
+func Test_getConfiguredSPOdJsonEnricherVolumes(t *testing.T) {
+	// The JSON enricher volume lookup requires the operator namespace, so
+	// this test cannot run in parallel.
+	t.Setenv(config.OperatorNamespaceEnvKey, "security-profiles-operator")
+
+	r := newTestReconciler()
+	r.clientReader = fake.NewClientBuilder().Build()
```
saschagrunert: The fake client is empty, so `getJsonEnricherVolume` always fails and the code that replaces or appends `json-enricher-log-output-volume` never runs. Could you seed the client with the `util.OperatorConfigMap` ConfigMap (with the `util.JsonEnricherLogVolumeSourceJson` and `util.JsonEnricherLogVolumeMountPath` keys) and build `r.baseSPOd` with `getEffectiveSPOd`? Then assert that the log volume is kept with the JSON enricher on and dropped with it off. This PR fixes that second case too: before, the volume stayed in the DaemonSet even with the enricher disabled.

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -224,6 +226,128 @@ func Test_getConfiguredSPOdEnricherArgsAreRevertible(t *testing.T) {
...
+	// AppArmor only: none of the SELinux, enricher or bpf host paths are needed.
+	off := featureVolumeNames(t, r, &spodapi.SPODSpec{
+		EnableAppArmor: new(true),
+		Selinux:        spodapi.SPODSelinuxConfig{Enable: new(false)},
+	})
+	requireVolumes(t, off, selinuxHostVolumes, false)
+	requireVolumes(t, off, enricherHostVolumes, false)
+	requireVolumes(t, off, bpfHostVolumes, false)
+
+	selinux := featureVolumeNames(t, r, &spodapi.SPODSpec{
```
saschagrunert: Could we add a case for the OpenShift default, where `Selinux.Enable` is unset and the CA inject type is `bindata.CAInjectTypeOpenShift`? `getConfiguredSPOd` turns SELinux on in that case. It's the production path for most SELinux users, so it's worth asserting that the SELinux volumes stay.

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -224,6 +226,128 @@ func Test_getConfiguredSPOdEnricherArgsAreRevertible(t *testing.T) {
...
+	requireVolumes(t, logEnricher, bpfHostVolumes, false)
+
+	bpfRecorder := featureVolumeNames(t, r, &spodapi.SPODSpec{
+		Enricher: spodapi.SPODEnricherConfig{EnableBpfRecorder: new(true)},
+	})
+	requireVolumes(t, bpfRecorder, bpfHostVolumes, true)
+	requireVolumes(t, bpfRecorder, selinuxHostVolumes, false)
+	requireVolumes(t, bpfRecorder, enricherHostVolumes, false)
+
+	// Disabling a feature again must drop its volumes again.
```
saschagrunert: Each render starts from a fresh `DeepCopy` of the base, so this line doesn't test disabling a feature. It only shows that earlier renders didn't change `baseSPOd`. Could you reword the comment to say that, or drop the assertion?

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -224,6 +226,128 @@ func Test_getConfiguredSPOdEnricherArgsAreRevertible(t *testing.T) {
 	}
 }
 
+// featureVolumeNames renders the SPOd for the given spec and returns the set
+// of volume names of the pod template. It also asserts that every volume mount
+// of the rendered containers is still backed by a volume.
+func featureVolumeNames(
```
saschagrunert: Nit: this returns the names of every rendered volume, not only the feature ones, so something like `renderedVolumeNames` would fit better. Also, the `off` case at L294 sets `EnableAppArmor: true`, but AppArmor doesn't change which volumes are rendered. Could you drop it, or add a comment saying why it's there?

KR-Ravindra:
@saschagrunert done in 143ddbb1, all five points:

1. `pruneUnmountedVolumes` now also collects `VolumeDevices` of every init container and container, matching kubelet's `GetPodVolumeNames`.
2. `Test_getConfiguredSPOdJsonEnricherVolumes` seeds the fake client with the `security-profiles-operator-profile` ConfigMap (both log volume keys) and builds `baseSPOd` with `getEffectiveSPOd`. It asserts the log output volume is rendered with the ConfigMap's hostPath when the enricher is on and dropped when it is off.
3. Added the OpenShift default case (`Selinux.Enable` unset, `CAInjectTypeOpenShift`): the SELinux volumes stay; the same unset spec under cert-manager keeps them off.
4. Replaced the "disabling again" comment and assertion with a check that pruning the rendered copy did not touch `baseSPOd` (a later SELinux-on render still finds its volumes).
5. Renamed the helper to `renderedVolumeNames` (on top of a `renderedPodSpec` helper) and dropped the unrelated `EnableAppArmor` from the all-off case.

Without the prune call both tests fail (`host-fsselinux-volume`, `sys-kernel-security-volume`); `go test ./internal/pkg/manager/spod/...` passes with it. Could you take another look?

Drafted with AI assistance and checked against the code.

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -224,6 +229,215 @@ func Test_getConfiguredSPOdEnricherArgsAreRevertible(t *testing.T) {
...
+
+	// OpenShift default: Selinux.Enable is unset and getConfiguredSPOd turns
+	// SELinux on because of the CA injection type, so the SELinux volumes stay.
+	openshift := renderedVolumeNames(t, r, &spodapi.SPODSpec{}, bindata.CAInjectTypeOpenShift)
+	requireVolumes(t, openshift, selinuxHostVolumes, true)
+	requireVolumes(t, openshift, enricherHostVolumes, false)
+	requireVolumes(t, openshift, bpfHostVolumes, false)
+
+	// The same unset Selinux.Enable outside OpenShift keeps SELinux off.
+	requireVolumes(t, renderedVolumeNames(t, r, &spodapi.SPODSpec{}, certManager), selinuxHostVolumes, false)
```
saschagrunert: `golines` rejects this at 106 characters, and it is the only finding in both `macos` and `pull-security-profiles-operator-verify`, so fixing it turns both green. Could you bind the render to a variable first, the way the other cases do?

```go
noSelinux := renderedVolumeNames(t, r, &spodapi.SPODSpec{}, certManager)
requireVolumes(t, noSelinux, selinuxHostVolumes, false)
```

internal/pkg/manager/spod/spod_controller_test.go
```diff
@@ -224,6 +229,215 @@ func Test_getConfiguredSPOdEnricherArgsAreRevertible(t *testing.T) {
...
+	volumes := map[string]bool{}
+	for i := range podSpec.Volumes {
+		volumes[podSpec.Volumes[i].Name] = true
+	}
+
+	for _, containers := range [][]v1.Container{podSpec.InitContainers, podSpec.Containers} {
+		for i := range containers {
+			for _, mount := range containers[i].VolumeMounts {
+				require.True(t, volumes[mount.Name],
+					"container %s mounts missing volume %s", containers[i].Name, mount.Name)
```
saschagrunert, line 258: Nit: now that `pruneUnmountedVolumes` counts `VolumeDevices` as a use as well, this invariant still only checks `VolumeMounts`, so a volume kept alive solely by a device would not be covered. Nothing declares block devices today, so this is only for symmetry with the production code, and fine to leave as is.

saschagrunert (commented):
The `ubi-image` failure is not from this PR: the UBI base image picked up eight `libevent` CVEs yesterday and it fails on `main` too. #3474 bumps the digest and fixes it, so please ignore that check here.

Thanks for working through the earlier round, all five points are addressed. One blocking nit left and one optional one.

internal/pkg/manager/spod/spod_controller.go
```diff
@@ -888,9 +888,47 @@ func (r *ReconcileSPOd) getConfiguredSPOd(
 	templateSpec.ImagePullSecrets = cfg.Spec.ImagePullSecrets
 	templateSpec.PriorityClassName = cfg.Spec.Scheduling.PriorityClassName
 
+	pruneUnmountedVolumes(templateSpec)
```
saschagrunert, line 891: `spodNeedsUpdate` can miss this pruning when the removed volume is the last one. It falls through to `apiequality.Semantic.DeepDerivative`, whose slice rule only reports a difference when the configured slice is *longer* than the live one; a shorter one is compared index-wise as a prefix and accepted. Dropping a trailing volume gives `true`, a middle one `false`.

That affects the json-enricher log volume, which `updateJsonEnricherSpec` appends last (`setup.go:271`): with the JSON enricher disabled but the operator ConfigMap configuring `json-enricher-log-volume-source`, nothing else in the template differs and the container counts are unchanged, so no update is issued and the stale hostPath stays on every node. The volumes in the commit message are all mid-list, so those do roll out correctly.

Could we add a volume count check next to the existing container count ones?

```go
len(configured.Spec.Template.Spec.Volumes) != len(found.Spec.Template.Spec.Volumes) ||
```

KR-Ravindra:
@saschagrunert done in 43748ec8, single commit amended:

1. `spodNeedsUpdate` now also compares `len(Spec.Template.Spec.Volumes)` next to the container counts. I checked the forked `deepValueDerive` slice rule you pointed at: a shorter configured slice is walked index-wise as a prefix, so dropping the last volume returned `false`. New `Test_spodNeedsUpdateVolumeCount` covers trailing, middle and added volumes; the trailing case fails without the count check.
2. The `renderedVolumeNames` call at the old line 336 is bound to `noSelinux` first, as suggested; no added line is over 100 characters now.
3. Took the optional nit as well: `renderedPodSpec` asserts `VolumeDevices` are backed by a volume too, for symmetry with `pruneUnmountedVolumes`.

Understood on `ubi-image`, thanks. `gofmt`, `go vet` and the spod/bindata unit tests pass locally. Could you take another look?

This comment was drafted with AI assistance and checked against the code.

saschagrunert (approved):
Thanks!

Issue #3442, opened 2026-09-15
Sign and attest staging images and profiles


saschagrunert:
#### What would you like to be added:

The staging Cloud Build job (`cloudbuild.yaml`, `hack/image-cross.sh`) signs everything it pushes and attaches signed attestations as OCI referrers (Sigstore bundles, not cosign v2 `.sig` tags). All signatures and attestations are signed keylessly by the build account, `SIGN=false` skips them.

| Artifact | Attestations |
|---|---|
| per-arch operator images (amd64, arm64, ppc64le) | SLSA provenance, SPDX SBOM, vulnerability scan, OpenVEX, build environment |
| manifest list | signature only, the per-arch images carry the attestations |
| bundle, catalog, Helm chart, newly pushed profiles (`base/runc`, `base/crun`, `seccomp-test-profiles`) | SLSA provenance, these contain no software packages |

**Status**

| Step | State |
|---|---|
| Staging in `us-central1-docker.pkg.dev/k8s-staging-images/sp-operator`, built as `sp-operator-sa@k8s-staging-images` | Done: kubernetes/k8s.io#9951, #3452, kubernetes/test-infra#37878; promotion source switched in kubernetes/k8s.io#9952 |
| Per-arch image configs report the right architecture | Done: #3445 |
| `spoc push` signs as bundles attached as referrers | Done: #3447 |
| Verification counts only signature bundles | Done: #3449 |
| Released catalog references the promoted bundle (staging is deleted after 90 days) | Done: #3451 |
| Unused `-bundle` and `-catalog` promotion prefixes dropped | Done: kubernetes/k8s.io#9950 |
| Signing in the staging build (manifest list, per-arch images, bundle, catalog, chart, new profiles) | Done: #3448, back on with #3458; first signed build verified with `cosign verify` |
| ID token for keyless signing in Cloud Build | Done: every build account holds `roles/iam.serviceAccountOpenIdTokenCreator` on itself (kubernetes/k8s.io#9953) |
| Signatures for the profiles published before signing worked (`seccomp-test-profiles`, `base/runc:v1.5.1`, `base/crun:v1.29.1`) | Done: #3459 signs published versions without a build account signature, all six verified |
| Promotion of signed staging images | Checked: `kpromo cip` dry run with the signed digests plans all edges and passes validation |
[...]

PR #3448, opened 2026-09-17
Sign staging images and profiles

saschagrunert:
#### What type of PR is this?

/kind feature

The staging build pushes all images and profiles unsigned. With this, it signs everything it pushes keylessly as Sigstore bundles attached as OCI referrers:

- the manifest list, the per-arch images and the bundle and catalog images, once by digest, with cosign v3.1.3 (`hack/sign-images.sh`, checksum verified)
- new base profiles and test profiles through `spoc push`, which signs with bundles since #3447

cosign gets the identity token for `sp-operator-sa@k8s-staging-images`, the account the build runs as since #3452, from the metadata server, so no impersonation is needed. The certificate identity is `sp-operator-sa@k8s-staging-images.iam.gserviceaccount.com` with issuer `https://accounts.google.com`. Bundles as referrers are what the image promoter can carry to production, while `.sig` tags from project identities are rejected there today.

Like `push-base-profiles.sh` already did, the scripts sign by default, and `SIGN=false` skips signing for manual runs such as `make image-cross` or `make push-test-artifacts`. The last staging build took 16.5 of the 20 minutes, so the timeout is raised to 30 minutes to make room for signing.

Refs #3442

#### Does this PR have test?

No. The scripts were run with stubbed `docker`, `cosign` and `spoc`, with the default and with `SIGN=false`; the signing itself can only run in Cloud Build.

This builds on #3452 (staging registry move to `k8s-staging-images`, merged). Signing with `sp-operator-sa` from the metadata server was verified outside Cloud Build (`cosign sign-blob` and `verify-blob` with that identity).

The new staging registry starts empty, so the first build pushes and signs all profiles.

The existing cosign v2.6.5 pin in `dependencies.yaml` is for the Vagrant CI and stays as is.

/hold

```release-note
NONE
```


saschagrunert:
/hold
Signing is enabled by default now, so this has to wait for kubernetes/k8s.io#9945 to be applied and kubernetes/test-infra#37876 to be merged.

saschagrunert:
/hold cancel
The first staging build in `k8s-staging-images` after #3452 passed and pushed all images and profiles, so this can go in. cosign gets the `sp-operator-sa` identity token from the metadata server.

PR #3452, opened 2026-09-17
Move the staging registry to k8s-staging-images

saschagrunert:
#### What type of PR is this?

/kind cleanup

The staging registry moves from `gcr.io/k8s-staging-sp-operator` to `us-central1-docker.pkg.dev/k8s-staging-images/sp-operator`, and the staging build runs as the dedicated `sp-operator-sa@k8s-staging-images` service account (kubernetes/k8s.io#9951, kubernetes/k8s.io#9486).

- `cloudbuild.yaml` sets `serviceAccount`, logs in to `us-central1-docker.pkg.dev` and pushes to the new repository, like `kubernetes/registry.k8s.io` does for `infra-tools`
- the push scripts, deploy manifests, Helm values, OLM example, bundle CSV, e2e tests and docs use the new staging location
- `hack/release.sh` and `hack/back-to-dev.sh` map it to `registry.k8s.io` and back

`k8s-staging-images` deletes images older than 90 days. The staging `latest` tags are pushed on every build, and the base profile and test artifact scripts push missing versions again, so the next main build restores cleaned up artifacts. #3451 makes the released catalog reference the promoted bundle, so nothing in production depends on staging.

Refs #3442

#### Does this PR have test?

No. The scripts pass shellcheck and the YAML files parse; the build itself can only run in Cloud Build.

This waits for kubernetes/k8s.io#9951 and kubernetes/test-infra#37878, which moves `post-security-profiles-operator-push-image` to the `k8s-staging-images` project. The base profile e2e test reads versions from staging; between a cleanup and the next main build it can miss them.

/hold

```release-note
Development images, profiles and manifests now use the staging registry us-central1-docker.pkg.dev/k8s-staging-images/sp-operator instead of gcr.io/k8s-staging-sp-operator.
```


saschagrunert:
/hold cancel
kubernetes/test-infra#37878 is merged, so the push job now builds in `k8s-staging-images`.

